PR #79, opened 2017-03-28
Fixes #72 adding obs data to empty comparison

### LINKED ISSUE #72: Adding an observed data to a new individual comparison chart crashes
To reproduce:
1. Create a new Individual Comparison Chart
1. Drad and drop some observed Data

msevestre:
Also with some minor changes not directly related but that were missing from our port to 7.1


src/PKSim.Core/CoreConstants.cs
```diff
@@ -63,8 +63,6 @@ public static class CoreConstants
       public static readonly string ExpressionPassword = "rhcp!!06";
       public static readonly string DefaultSkin = "Office 2013 Light Gray";
       public static readonly string AccessUserName = "Admin";
-      public static readonly string SQLiteUserName = "admin";
-      public static readonly string TemplatePassword = "";
```
msevestre: not required anymore

src/PKSim.Infrastructure/ORM/Core/TemplateDatabase.cs
```diff
@@ -8,8 +6,5 @@ public interface ITemplateDatabase : IDatabase
 
    public class TemplateDatabase : AccessDatabase, ITemplateDatabase
    {
-      public TemplateDatabase(): base(CoreConstants.TemplatePassword)
```
msevestre: null and empty is not the same. This was fix in 7.0 but I guess we did not migrate that

src/PKSim.Presentation/Presenters/Charts/IndividualSimulationComparisonPresenter.cs
```diff
@@ -195,6 +195,7 @@ private void editChart(IndividualSimulationComparison chart)
       {
          InitializeAnalysis(chart);
          UpdateTemplatesBasedOn(_chartPresenterContext.ProjectRetriever.CurrentProject);
+         BindChartToEditors();
```
msevestre: This is the actual fix. Make sure we init the editors even if the comparison has NO simulations

src/PKSim.UI/PKSim.UI.csproj
```diff
@@ -427,7 +427,6 @@
     <Compile Include="Views\Individuals\IndividualMoleculePropertiesView.Designer.cs">
       <DependentUpon>IndividualMoleculePropertiesView.cs</DependentUpon>
     </Compile>
-    <Compile Include="Views\MRUArrayList.cs" />
```
msevestre: This file is not used anymore (move to core)

PR #93, opened 2017-04-07
Fixes #69

### LINKED ISSUE #69: NHANES population needs update
The implemented NHANES population causes inconsistencies when generating individuals younger than 5 years of age. This is due to a methodological deviation from other databases. The physiology needs to be defined for the entire body in the same age steps. Once we update the physiology, we should also reevaluate the distribution for anthropometric parameters for the entire NHANES population

Yuri05:


tests/PKSim.Tests/IntegrationTests/DatabaseUpdateSpecs.cs
```diff
@@ -241,6 +242,25 @@ public override void GlobalContext()
...
+      public void should_update_nhanes_values()
+      {
+         var nhanesParams = _parameterDistributionRepository.All().Where(pd => pd.Population.EndsWith("NHANES_1997")).ToList();
+
+         var bwParams = nhanesParams.Where(p => p.ParameterName.Equals(CoreConstants.Parameter.MEAN_WEIGHT)).ToList();
+         var heightParams = nhanesParams.Where(p => p.ParameterName.Equals(CoreConstants.Parameter.MEAN_HEIGHT)).ToList();
+         var volumeParams = nhanesParams.Where(p => p.ParameterName.Equals(CoreConstants.Parameter.VOLUME)).ToList();
+
+         //check number of new bw/height/volume parameters
+         (bwParams.Count()+ heightParams.Count()+ volumeParams.Count()).ShouldBeEqualTo(1420);
```
msevestre: you could remove all the () after Count since you are dealing with lists. 
Suggestion is to run the clean method from resharper

msevestre:
@Yuri05 In future, I'd suggest to also write text of issue not only number so that we can see right away, what is being fixed

PR #101, opened 2017-04-11
Fixes #95 Simulation creation: handling of selecting the same inhibit…

### LINKED ISSUE #95: Simulation creation: handling of selecting the same inhibition/induction process more than once
In a project with inhibition/induction processes, when creating a simulation:

Creation Simulation -> Processes -> Inhibition/Induction,

when adding an interaction and selecting a process already present in the list (see screenshot), a warning dialog appears:
"PK-Sim®

'Inhibitor CYP3A4-01_Comp' cannot be selected more than once.
For more information or questions, please visit the Open Systems Pharmacology Forum (forum.open-systems-pharmacology.org)."

However, after selecting "OK", the process is selected. When clicking on "Next" and then "Previous", the process next to the initially selected is chosen (secreenshot 2). The behavior is misleading. It should not be possible to proceed until the duplicated process is deselected by user.

inhibitioninduction
inhibitioninduction2

msevestre:
…ion/induction process more than once

### DIFF
--- a/src/PKSim.Presentation/Presenters/Simulations/SimulationCompoundInteractionSelectionPresenter.cs
+++ b/src/PKSim.Presentation/Presenters/Simulations/SimulationCompoundInteractionSelectionPresenter.cs
@@ -85,10 +85,18 @@ namespace PKSim.Presentation.Presenters.Simulations
       public void CompoundProcessSelectionChanged(SimulationInteractionProcessSelectionDTO dto, PartialProcess interactionProcess)
       {
          if (processAlreadySelected(interactionProcess))
+         {
+            refreshView();
             throw new CannotSelectThePartialProcessMoreThanOnceException(interactionProcess);
+         }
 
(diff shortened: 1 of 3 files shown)


src/PKSim.Presentation/Presenters/Simulations/SimulationCompoundInteractionSelectionPresenter.cs
```diff
@@ -85,12 +85,20 @@ private SimulationInteractionProcessSelectionDTO mapFrom(SimulationPartialProces
...
             throw new CannotSelectThePartialProcessMoreThanOnceException(interactionProcess);
+         }
 
          dto.SimulationPartialProcess.CompoundProcess = interactionProcess;
          updateWaring();
       }
 
+      private void refreshView()
+      {
+         _view.Repaint();
```
msevestre: Probably required after DevExpress update. We need to refresh the view after the exception is thrown

PR #108, opened 2017-04-13
Fixes #107 batch simulation cannot be run anymore

### LINKED ISSUE #107: Cannot run batch simulations because of wrong filter

msevestre:


src/PKSim.BatchTool/Services/JsonSimulationRunner.cs
```diff
@@ -46,39 +46,39 @@ public Task RunBatch(dynamic parameters)
...
 
-            _logger.AddInSeparator("Starting batch run: {0}".FormatWith(DateTime.Now.ToIsoFormat()));
+            _logger.AddInSeparator($"Starting batch run: {DateTime.Now.ToIsoFormat()}");
 
             var inputDirectory = new DirectoryInfo(inputFolder);
             if (!inputDirectory.Exists)
                throw new ArgumentException($"Input folder '{inputFolder}' does not exist");
 
-            var allSimulationFiles = inputDirectory.GetFiles(Constants.Filter.JSON_FILE_FILTER);
+            var allSimulationFiles = inputDirectory.GetFiles(CoreConstants.Filter.JSON_FILTER);
```
msevestre: This was the issue. Not using the right filter to get the files after refactoring.
Rest of changes is just about using new string interpolation instead of method `FormatWith` that should probably become obsolete

PR #114, opened 2017-04-13
Fixes #110 comparison does not work

### LINKED ISSUE #110: Working Journal and simulation comparison
If simulation was loaded from working journal, comparison with other simulations does not work properly.
E.g. in the project attached, simulation _**Paracetamol Solution**_ was loaded from working journal (also attached).
Comparing it with the simulation _**NEW_rabbit paracetamol solution**_ shows no differences (both in Compare Formula and Compare Value mode). But e.g. body weight and all organ volumes differ.
[Comparison_WJ.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/920197/Comparison_WJ.zip)

msevestre:


msevestre:
DiffBuilder in PKSim where not registered anymore. Adding the DiffBuilder convention based is enough. A test also to make sure that we are laoding the right diffbuilder

PR #130, opened 2017-04-24
Fixes #129 add batch tool to PKSim MSI

### LINKED ISSUE #129: Add BatchTool to PKSim msi

msevestre:


msevestre:
This PR ensures that the two files required for the Batch tool to run are added to the setup. Also it was required to patch assembly binding for SX here as well

PR #132, opened 2017-04-24
Fixes #131 add cli to batch tool

### LINKED ISSUE #131: Add CLI to batch tool
--output => Path where batch output will be created
--input => Path where json files are located
--log => Full path of communication log file if a path is given or log file name to be saved under output

msevestre:


src/PKSim.BatchTool/Presenters/BatchMainPresenter.cs
```diff
@@ -44,16 +53,18 @@ public void GenerateTrainingMaterial()
...
+         View.Hide();
          presenter.InitializeWith(_startOptions);
       }
 
       public void GenerateProjectOverview()
       {
          start<IGenerateProjectOverviewPresenter>();
       }
+
+      public IView ActiveView => _startOptions.IsValid() ? _batchSilentView : BaseView;
```
msevestre: We need to start some kind of view in a Program.cs. I do not want to see the actual view popup so for now, I have this silent view that is minimized. Do you have a better idea?

src/PKSim.BatchTool/Program.cs
```diff
@@ -1,34 +1,44 @@
...
          Application.SetCompatibleTextRenderingDefault(false);
          try
-         { 
+         {
             BatchStarter.Start();
             var mainPresenter = IoC.Resolve<IBatchMainPresenter>();
-            mainPresenter.Initialize(BatchStartOptions.From(args));
-            Application.Run(mainPresenter.View.DowncastTo<Form>());
+            var options = BatchStartOptions.From(args);
+            mainPresenter.Initialize(options);
```
msevestre: This starts the actual run asynchronously

src/PKSim.BatchTool/Program.cs
```diff
@@ -1,34 +1,44 @@
...
+            Application.Run(mainPresenter.ActiveView.DowncastTo<Form>());
```
msevestre: This starts the app. The app will stay here until the exit is called programatically

tests/PKSim.Tests/BatchTool/BatchStartOptionsSpecs.cs
```diff
@@ -0,0 +1,123 @@
...
+         sut.LogFileFullPath.ShouldBeEqualTo(_logFileFullPath);
+      }
+   }
+
+   public class When_parsing_valid_arguments_to_the_batch_start_options_with_full_keys : concern_for_BatchStartOptions
+   {
+      protected override void Context()
+      {
+         base.Context();
+         _args.Add("--input");
```
msevestre: example of command line.
The logfile is not used yet but the cli is available

src/PKSim.BatchTool/Presenters/BatchMainPresenter.cs
rwmcintosh: If you need to start a message loop with Application.Run then I don't have an alternative. 

I didn't investigate how this all works under the hood but inside main() if you don't call Application.Run then you don't need this view.

src/PKSim.BatchTool/Program.cs
rwmcintosh: If you called mainPresenter.StartBatchRun instead of Application.Run then wouldn't you have a 'headless' mode? Again, not sure how the batch runner works.

All depends if you can get away without a message loop.

msevestre:
I think we need the message loop here. Can't tell what's going to go south
if we don't

PR #144, opened 2017-04-28
Fixes #129 Add BatchTool to PKSim msi - Simplify setup to use harvesting

### LINKED ISSUE #129: Add BatchTool to PKSim msi

msevestre:


msevestre:
Now that we copy the batch tool to PKSim bin debug folder, there is no need to add those files by hand. They will be harvested automatically by the heat task

PR #167, opened 2017-05-18
Fixes #166 Hide Relative Expression parameters in sim when Tissue=Int…

### LINKED ISSUE #166: Hide Relative Expression parameters in sim when Tissue=Intra & Loc. = Insterstitial"
`Localization in Tissue=Intracellular` & `Localization in vasc. Endo=Interstitial`

msevestre:
…ra & Loc. = Insterstitial

### DIFF
--- a/src/PKSim.Core/Services/SimulationModelCreator.cs
+++ b/src/PKSim.Core/Services/SimulationModelCreator.cs
@@ -1,10 +1,10 @@
 using System.Collections.Generic;
 using System.Linq;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Services;
 using OSPSuite.Utility.Extensions;
 using PKSim.Core.Model;
 using PKSim.Core.Model.Extensions;
-using OSPSuite.Core.Domain;
-using OSPSuite.Core.Domain.Services;
 
 namespace PKSim.Core.Services
 {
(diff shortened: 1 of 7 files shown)


tests/PKSim.Tests/IntegrationTests/SimulationModelCreatorSpecs.cs
```diff
@@ -81,5 +89,25 @@ public void the_start_amount_of_all_protein_should_point_the_the_start_amount_pa
          path.ShouldNotBeNull();
          path.Last().ShouldBeEqualTo(CoreConstants.Parameter.START_AMOUNT);
       }
+
+      [Observation]
+      public void should_hide_the_relative_expression_parameters_defined_in_interstitial_for_protein_defined_with_a_tissue_localization_in_cell_and_a_vascular_endothelium_localization_in_interstitial()
```
msevestre: and tests for new implementation

src/PKSim.Core/Services/SimulationModelCreator.cs
```diff
@@ -101,62 +101,108 @@ private void hideAllUndefinedParameterForMolecules(Simulation simulation)
...
+         var protein = molecule as IndividualProtein;
+
+         if (protein?.TissueLocation != TissueLocation.Intracellular || protein.IntracellularVascularEndoLocation != IntracellularVascularEndoLocation.Interstitial)
+            return;
+
+         var vascularEndothelium = protein.ExpressionContainer(CoreConstants.Compartment.VascularEndothelium);
+         if (vascularEndothelium == null)
+            return;
+
+         var allInterstitialContainer = _expressionContainersRetriever.AllContainersFor(individual.Organism, allOrganismContainers, protein, vascularEndothelium);
```
Yuri05: We could also directly get all interstitial containers of the individual and replace lines 139-143 
But anyway, the above code is also correct :-)

msevestre: This is correct. That being said, if we ever changed which container is affected by the molecule  properties, we won't have to change that code hopefully (the knowledge of what container are depending on vascular endothelium is only in one place)

PR #187, opened 2017-05-24
Fixes #124 show individual id in box plot tooltip

### LINKED ISSUE #124: Show Individual Numbers within Population Box Whisker Analysis
To connect the values in those plots, e.g. specific quantiles of PK parameters, with individuals, the mouse-over text above the box whisker analysis plots should additionally show the individual numbers associated with the corresponding value. This would allow follow-up analyses on those individuals, e.g. separate sensitivity analyses for the 5,50 and 95 percentile individuals.

(split from #122 that would allow a quick extraction of those individuals)

msevestre:


appveyor-nightly.yml
```diff
@@ -39,6 +39,8 @@ before_build:
   - nuget sources add -name simmodel -source https://ci.appveyor.com/nuget/ospsuite-simmodel
   - nuget sources add -name cvodes -source https://ci.appveyor.com/nuget/ospsuite-simmodel-solver-cvodes-282
   - nuget sources add -name core -source https://ci.appveyor.com/nuget/ospsuite-core
+  #TODO DELETE
+  - nuget sources add -name core -source https://ci.appveyor.com/nuget/ospsuite-core-version7-2-0
```
msevestre: This will have to go once we merge develop into master

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -945,15 +944,33 @@ public static string XAsTooltip(string x)
...
+                  sb.AppendLine(valueWithIndividualId(v,outlierIndividualIds[i]));
+               });
+            }
+
+            return sb.ToString();
          }
 
+         private static string percentilWithIndividualId(string percentil, string value, int individualId) => valueWithIndividualId($"{percentil}% = {value}", individualId);
+
+         private static string valueWithIndividualId(string value, int individualId) => $"{value}, IndividualId = {individualId}";
```
msevestre: @Yuri05  This line may have to be changed depending on how we want to display the tooltip

src/PKSim.Core/PKSim.Core.csproj
```diff
@@ -107,6 +107,9 @@
       <RequiredTargetFramework>3.5</RequiredTargetFramework>
     </Reference>
     <Reference Include="System.Drawing" />
+    <Reference Include="System.ValueTuple, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
```
msevestre: This is a neat package until we move to .NET 4.7 .We can use tuple as return value 
sthg like 
```
(string a, string b) function test(){
}
```

I did not use it after all, but we kept it anyways for later use

src/PKSim.Core/Services/BoxWhiskerChartsDataCreator.cs
```diff
@@ -64,37 +67,70 @@ private void setSeriesValues(CurveData<BoxWhiskerXValue, BoxWhiskerYValue> serie
...
+      private void updateIndexes(BoxWhiskerYValue boxWhiskerYValue, float[] fieldValues, List<float> sortedValues)
+      {
+         boxWhiskerYValue.AllValues.Each(v =>
          {
-            int result = _columnFields[i].Compare(arg1[i], arg2[i]);
-            if (result != 0)
-               return result;
-         }
+            int index = sortedValues.BinarySearch(v);
+            v.IndividualId = Array.IndexOf(fieldValues, index >= 0 ? v : sortedValues[~index]);
```
msevestre: That's the actual retrieval of the IndividualId. I cannot do it in the aggregate function because we do not have access to the full dataset

src/PKSim.Core/Model/AggregationFunctions.cs
```diff
@@ -28,40 +28,43 @@ public static class AggregationFunctions
 
       private static Aggregate<BoxWhiskerYValue> boxWhiskerAggregation(double percentile)
       {
+         const float outlierRange = 1.5F;
+
          return new Aggregate<double, BoxWhiskerYValue>
          {
             Aggregation = doubles =>
```
msevestre: This was where I tried to make it happen first. But the (doubles) passed as parameter only contains a subset of all values (already pivoted so to speak)

src/PKSim.Core/Model/AggregationFunctions.cs
```diff
@@ -28,40 +28,43 @@ public static class AggregationFunctions
```
Yuri05: why is `BoxWhisker`**_90_**`Aggregation = boxWhiskerAggregation(`**_5_**`)`?
same for `BoxWhisker95Aggregation `(also this one is not used anywhere)

msevestre: I think this is because of the range that is being created (from 5% to 95% => 90) as opposed to (2.5% => 97.5%). I did not choose that nomenclature, it was part of the specs for 5.3...

Yuri05: probably that's it.

PR #190, opened 2017-05-24
Fixes #164 - observed data that are used in a comparison

### LINKED ISSUE #164: It should not be possible to remove observed data that are used in a comparison
Should it be possible to delete observed data that are used in comparison? We are blocking that from happening when observed data are used in simulation and/or Parameter Identification.

Implementation right now is to allow the observed data to be deleted

rwmcintosh:


src/PKSim.Infrastructure/Services/SimulationComparisonContentLoader.cs
```diff
@@ -1,3 +1,6 @@
+using System.Linq;
+using FluentNHibernate.Utils;
```
msevestre: Remove usage here of FluentNhibernate

rwmcintosh: done

PR #193, opened 2017-05-26
Fixes #170 - Export units for population parameters

### LINKED ISSUE #170: Export units for population parameters
When exporting a population, it would be helpful to also export units. I guess everything is based on PK-Sim standard units, so in principle ok, but sometimes inconvenient and confusing. For example, for some composite parameters, values/units might not correspond to expectation and to what is displayed within PK-Sim, e.g. a BMI values of 0.25 not 25. I suggest putting units in brackets behind the parameter name or adding an additional line.

rwmcintosh:


rwmcintosh:
Exports with unit if available and also removes the unit from the column name when importing.

src/PKSim.Infrastructure/Services/IndividualPropertiesCacheImporter.cs
```diff
@@ -72,6 +72,15 @@ private IndividualPropertiesCache individualPropertiesCacheFrom(string fileFullP
          return individualPropertiesCache;
       }
 
+      private string removeUnitsFromHeader(string header)
```
msevestre: This is not the right place to remove units from header. It is possible to have parameter with unit in name!
For example `Acidic phospholipids [mg/g] - RR`

This one does not end with [] but you see my point .Parameter can be named by the user etc.. so this will potentially fail.
Instead I would suggest to import as is and when trying to match parameter by name, have a fall back option to load a parameter removing potential unit if the parameter was not found

src/PKSim.Infrastructure/Services/PopulationExportTask.cs
```diff
@@ -246,12 +243,17 @@ private void addParameterToTable(IVectorialParametersContainer parameterContaine
          //some path have changed and the parameter is not found anymore
          if (parameter == null) return;
 
-         addColumnForParameterToTable(parameterContainer, dataTable, _entityPathResolver.PathFor(parameter));
+         addColumnForParameterToTable(parameterContainer, dataTable, _entityPathResolver.PathFor(parameter), parameter.Dimension.BaseUnit.ToString());
```
msevestre: use unit.Name

src/PKSim.Infrastructure/Services/PopulationExportTask.cs
```diff
@@ -246,12 +243,17 @@ private void addParameterToTable(IVectorialParametersContainer parameterContaine
...
+      }
+
+      private void addColumnForParameterToTable(IVectorialParametersContainer parameterContainer, DataTable dataTable, string parameterPath, string displayUnit)
+      {
+         addColumnValues(parameterContainer, dataTable, columnName(parameterPath, displayUnit), parameterContainer.AllValuesFor(parameterPath));
       }
 
-      private void addColumnForParameterToTable(IVectorialParametersContainer parameterContainer, DataTable dataTable, string parameterPath)
+      private static string columnName(string parameterPath, string displayUnit)
```
msevestre: this is NOT display unit. This is baseUnit

src/PKSim.Infrastructure/Services/PopulationExportTask.cs
```diff
@@ -140,10 +139,8 @@ private void addColumnValues<T>(IVectorialParametersContainer parameterContainer
 
       private bool parameterShouldBeExported(IParameter parameter, IEnumerable<IParameter> advancedParameters)
       {
-         //BMI and BodyWeight should always be exported (Mantis 2907)
```
msevestre: leave the comment please but remove the reference to Mantis

src/PKSim.Infrastructure/Services/PopulationExportTask.cs
```diff
@@ -140,10 +139,8 @@ private void addColumnValues<T>(IVectorialParametersContainer parameterContainer
...
          if (parameter.NameIsOneOf(CoreConstants.Parameter.BMI, CoreConstants.Parameter.WEIGHT)) return true;
 
-         //BMI MeanHeight MeanWeight should never be exported (Mantis 2907)
```
msevestre: leave the comment please but remove the reference to Mantis

src/PKSim.Infrastructure/Services/PopulationExportTask.cs
```diff
@@ -246,12 +243,17 @@ private void addParameterToTable(IVectorialParametersContainer parameterContaine
...
+      private void addColumnForParameterToTable(IVectorialParametersContainer parameterContainer, DataTable dataTable, string parameterPath, string baseUnit)
+      {
+         addColumnValues(parameterContainer, dataTable, columnName(parameterPath, baseUnit), parameterContainer.AllValuesFor(parameterPath));
       }
 
-      private void addColumnForParameterToTable(IVectorialParametersContainer parameterContainer, DataTable dataTable, string parameterPath)
+      private static string columnName(string parameterPath, string baseUnit)
       {
-         addColumnValues(parameterContainer, dataTable, parameterPath, parameterContainer.AllValuesFor(parameterPath));
+         return string.IsNullOrEmpty(baseUnit) ? parameterPath : $"{parameterPath} [{baseUnit}]";
```
msevestre: Use the method already available in Constants `NameWithUnitFor`

PR #201, opened 2017-05-31
Fixes #170 export units for population

### LINKED ISSUE #170: Export units for population parameters

rwmcintosh:


src/PKSim.Infrastructure/Services/PopulationExportTask.cs
```diff
@@ -55,12 +55,12 @@ public class PopulationExportTask : IPopulationExportTask
 
       public void ExportToCSV(Population population)
       {
-         exportVectorialParametersContainerToCSV(population, CreatePopulationDataFor);
+         exportVectorialParametersContainerToCSV(population, x => CreatePopulationDataFor(x, true));
```
msevestre: use named parameter here. Makes the code so much more readable

src/PKSim.Infrastructure/Services/PopulationExportTask.cs
```diff
@@ -55,12 +55,12 @@ public class PopulationExportTask : IPopulationExportTask
...
       }
 
       public void ExportToCSV(PopulationSimulation populationSimulation)
       {
-         exportVectorialParametersContainerToCSV(populationSimulation, CreatePopulationDataFor);
+         exportVectorialParametersContainerToCSV(populationSimulation, x => CreatePopulationDataFor(x, true));
```
msevestre: Same here

src/PKSim.Infrastructure/Services/PopulationExportTask.cs
```diff
@@ -238,12 +238,14 @@ private bool settingsRequired(Simulation simulation)
...
       }
 
-      private void addParameterToTable(IVectorialParametersContainer parameterContainer, DataTable dataTable, IParameter parameter)
+      private void addParameterToTable(IVectorialParametersContainer parameterContainer, DataTable dataTable, IParameter parameter, bool includeUnitsInHeader)
       {
          //some path have changed and the parameter is not found anymore
          if (parameter == null) return;
 
-         addColumnForParameterToTable(parameterContainer, dataTable, _entityPathResolver.PathFor(parameter), parameter.Dimension.BaseUnit.Name);
+         var unitName = includeUnitsInHeader ? parameter.Dimension.BaseUnit.Name : string.Empty;
```
msevestre: Suggestion is to create a columnName here instead of passing an empty unitName if not requried. 
Then rename the addColumnForParameterToTable parameter to columnName

msevestre: Suggestion is to create a columnName here instead of passing an empty unitName if not requried. 
Then rename the parameter in `addColumnForParameterToTable` to columnName

src/PKSim.Infrastructure/Services/PopulationExportTask.cs
```diff
@@ -55,12 +55,12 @@ public class PopulationExportTask : IPopulationExportTask
 
       public void ExportToCSV(Population population)
       {
-         exportVectorialParametersContainerToCSV(population, x => CreatePopulationDataFor(x, true));
+         exportVectorialParametersContainerToCSV(population, populationToCreateDataFor => CreatePopulationDataFor(populationToCreateDataFor, true));
```
msevestre: LOL. I meant the bool parameter. The x if fine. But name the optional parameter

PR #206, opened 2017-05-31
Fixes #175 wrong values for AUC ratio

### LINKED ISSUE #175: Wrong values for AUC Ratio and Cmax ratio in global PK analysis
The values for AUCR and CmaxR show in the global PK analysis after single dose po do not correspond to AUC inf or Cmax with inhibitor divided by AUCinf without inhibitor taken from PK analysis. The difference for both is about 2 % in a Midazolam example (attached in PK-sim file)

For multiple dose po applications (AUC_tDlast-1_tDlast is taken here) and slow clearance cpds this divergence gets bigger for AUC: e.g. 97 and 156 % (!) after 5 or 10 days respectively.
But it gets smaller for Cmax R: 0,3 and 0,26 % after 5 or 10 days respectively.

See attached excel files for details

[DDI Demo Github.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/1018815/DDI.Demo.Github.zip)

msevestre:


msevestre:
@Yuri05 I have simply updated the tooltip to use the right parameter. Test written to ensure proper database update. Other calculations are fine in my opinion (see discussion in #175)

PR #211, opened 2017-06-02
203 body surface area dosing

msevestre:


src/PKSim.Infrastructure/ORM/Repositories/DimensionRepository.cs
```diff
@@ -102,18 +102,39 @@ public IDimension MergedDimensionFor(object objectThatMightHaveDimension)
...
+         inputDoseDimension.BaseUnit.Visible = false;
+
+         //factor to convert mg to kg
+         addInputDoseUnit(inputDoseDimension, CoreConstants.Units.mg, 1e-6);
+
+         //factor to convert mg/kg to kg/kg
+         inputDoseDimension.DefaultUnit = addInputDoseUnit(inputDoseDimension, CoreConstants.Units.MgPerKg, 1e-6);
+
+         //factor to convert mg/m2 to kg/dm2
+         addInputDoseUnit(inputDoseDimension, CoreConstants.Units.MgPerM2, 1e-8);
```
msevestre: This is the conversion factor to convert mg/m2 (input unit) to kg/dm2 (core unit)

src/PKSim.Infrastructure/ProjectConverter/v7_2/Converter710To720.cs
```diff
@@ -0,0 +1,125 @@
...
+      }
+
+      public void Visit(Population population)
+      {
+         convertIndividual(population.FirstIndividual);
+         addBSAParameterValues(population);
+      }
+
+      private void convertIndividual(Individual individual)
+      {
```
msevestre: I am only converting the Individual and Population. (building block and in the simulation). I am not touching the simulations

src/PKSim.Infrastructure/ProjectConverter/v7_2/Converter710To720.cs
```diff
@@ -0,0 +1,125 @@
...
+
+         var parameterValues = new ParameterValues(_entityPathResolver.PathFor(bsaParameter));
+         for (int i = 0; i < allWeights.Count; i++)
+         {
+            weightParameter.Value = allWeights[i];
+            heightParameter.Value = allHeights[i];
+            parameterValues.Add(bsaParameter.Value);
+         }
+
+         population.IndividualPropertiesCache.Add(parameterValues);
```
msevestre: For a population, I also calculate the new BSA values. That way, we can use the parameters right away for any other simuations

msevestre:
All db changes are tested defacto in code since I am creating new individuals and checking for their properties. No need to test that explicitely

@Yuri05 Project conversion: I remember that for some special case, you had to convert the individual IN all converters instead of only in the last one in the stack. I forgot why. Can you let me know if you think I need to add the BSA parameter earlier? Hopefully running all integrationstest will tell us more :)

msevestre:
I have only added the 2 methods from the Issue. Very easy to add more if we want to

src/PKSim.Infrastructure/ORM/FlatObjects/FlatPopulation.cs
```diff
@@ -6,6 +6,7 @@ public class FlatPopulation : FlatObject
         public string Species { get; set; }
         public bool IsAgeDependent { get; set; }
         public bool IsHeightDependent { get; set; }
+        public bool IsBodySurfaceAreaDependent { get; set; }
```
Yuri05: I'm not sure, if we really need `IsBodySurfaceAreaDependent` here.
If a species is height dependent, it is automatically BSA dependent.

We could keep the property `IsBodySurfaceAreaDependent` in `SpeciesPopulation` (for the clarity of code), but it would just return the value of `IsHeightDependent`

What do u think?

src/PKSim.Core/CoreConstants.cs
```diff
@@ -252,6 +259,7 @@ public static class CalculationMethod
          public static readonly string LinksCommon = "LinksCommon";
          public static readonly string DistributionInVitroPKSim = "DistributionInVitro_PKSim";
          public static readonly string ApplicationParameter = "ApplicationParameter";
+         public static readonly string ApplicationParameter_Human = "ApplicationParameter_Human";
```
Yuri05: You mean "new CM", right?

msevestre:
That's how it was implemented before but i changed it to because the code
was not symetrical anymore
We could argue that a species is automatically Height dependent if it is
IsAgeDepedent etc..

src/PKSim.Infrastructure/ORM/Repositories/DimensionRepository.cs
Yuri05: can u define those conversion factors as const please.

src/PKSim.Core/CoreConstants.cs
msevestre: no Category (and CM). You can only have one active CM per category. ApplicaitonParameter is still used for Human. The `ApplicationParameter_human` is used as an extra category

tests/PKSim.Tests/IntegrationTests/DatabaseUpdateSpecs.cs
```diff
@@ -2,224 +2,42 @@
...
-         var flatMoleculeMapper = IoC.Resolve<IFlatMoleculeToMoleculeBuilderMapper>();
-
-         var protein = flatMoleculeMapper.MapFrom(flatMoleculesRepo.FindBy(QuantityType.Protein), new FormulaCache());
-
-         protein.Parameter(CoreConstants.Parameter.ONTOGENY_FACTOR_GI).CanBeVaried.ShouldBeTrue();
-         protein.Parameter(CoreConstants.Parameter.ONTOGENY_FACTOR).CanBeVaried.ShouldBeTrue();
-      }
-   }
-
-   public class When_checking_the_changes_in_the_database_for_version_6_4 : concern_for_DatabaseUpdate
```
Yuri05: why did u remove those old conversion tests?

msevestre: There are not removed. I just flipped the order so that the latest tests are first. Otherwise one always have to scroll to see the latest changes. Or do you mean that I removed a class accidentally?

Yuri05:
Well, I can imagine having e.g. age-dependent but not height dependent species, but not height dependent and not bsa dependent.
But the current implementation is correct for sure.

Then please add one unit test checking that the values of BSA-dependent and height-dependent properties are the same for all populations
In case we get a different situation in the future (which I cannot imagine), the test will be broken.

Von: Michael Sevestre [mailto:notifications@github.com]
Gesendet: Dienstag, 6. Juni 2017 16:54
An: Open-Systems-Pharmacology/PK-Sim
Cc: Juri Solodenko; Mention
Betreff: Re: [Open-Systems-Pharmacology/PK-Sim] 203 body surface area dosing (#211)

That's how it was implemented before but i changed it to because the code
was not symetrical anymore
We could argue that a species is automatically Height dependent if it is
IsAgeDepedent etc..

src/PKSim.Infrastructure/ORM/Repositories/DimensionRepository.cs
msevestre: done

src/PKSim.Core/Model/SpeciesPopulation.cs
```diff
@@ -10,14 +10,12 @@ public class SpeciesPopulation : ObjectBase
       public virtual string Species { get; set; }
       public virtual bool IsAgeDependent { get; set; }
       public virtual bool IsHeightDependent { get; set; }
+      public virtual bool IsBodySurfaceAreaDependent { get; set; }
```
msevestre: @Yuri05 Removed as requested. This is now only a property get

src/PKSim.Infrastructure/ORM/FlatObjects/FlatPopulation.cs
msevestre: done

tests/PKSim.Tests/Core/SimulationConfigurationValidatorSpecs.cs
```diff
@@ -44,7 +44,7 @@ protected override void Context()
          _protocolToProtocolSchemaItemMapper = A.Fake<IProtocolToSchemaItemsMapper>();
          sut = new SimulationConfigurationValidator(_protocolToProtocolSchemaItemMapper);
 
-         _speciesPopulation.IsBodySurfaceAreaDependent = false;
+         _speciesPopulation.IsHeightDependent = false;
```
msevestre: @Yuri05 This is why I did not like it. The tests are not as expressive. But whatever

tests/PKSim.Tests/IntegrationTests/DatabaseUpdateSpecs.cs
Yuri05: ok, haven't seen them (too many changes in the file :-). Everything OK then

PR #235, opened 2017-06-26
Fixes #161 PK-Sim portable

### LINKED ISSUE #161: PK-Sim Portable
See discussion in forum 
https://github.com/Open-Systems-Pharmacology/Forum/issues/9

The goal would be to ensure that we can create a distribution package for PK-Sim that does not write in registry nor require admin rights. 

A few remarks

- PK-Sim relies on registry entries however to find where MoBi is installed. If those registry entries are not found, we could have the user specify in the application settings were MoBi is installed. Also we could try to find it by convention using the same location as PK-Sim and X.Y version

- For example if PK-Sim is installed under `C:\MyApp\Open Systems Pharmacology\PK-Sim\7.1`, we could try to find MoBi under   `C:\MyApp\Open Systems Pharmacology\MoBi\7.1`

- Some files are install globally (pk-parameters, PKSim DB, dimensions etc..). Those file would have to be installed in the PK-Sim folder instead and PK-Sim would need to handle those differences gracefully

msevestre:


appveyor-nightly.yml
```diff
@@ -45,6 +45,7 @@ before_build:
 after_build:
   - rake "create_setup[%APPVEYOR_BUILD_VERSION%, %CONFIGURATION%, C:/smartxls.zip, 2.6.2.9]"
   - 7z a setup.zip %APPVEYOR_BUILD_FOLDER%\setup\deploy\*.msi
+  - rake "create_portable_setup[%APPVEYOR_BUILD_VERSION%, %CONFIGURATION%]"
```
msevestre: new task called on nightly

appveyor-nightly.yml
```diff
@@ -57,6 +58,7 @@ notifications:
 
 on_finish: 
   - ps: Get-ChildItem setup.zip | % { Push-AppveyorArtifact $_.FullName -FileName $_.Name }
+  - ps: Get-ChildItem .\setup\portable-setup.zip | % { Push-AppveyorArtifact $_.FullName -FileName $_.Name }
```
msevestre: and new artifact

setup/setup.wxs
```diff
@@ -59,9 +59,6 @@
             </File>
           </Component>
 
-          <Component Id='log4net.config.xml' Guid='A6BF052C-E9BA-4959-82B1-5B61011A06AB'>
```
msevestre: This file should be installed under AppData

src/PKSim.Core/CoreConstants.cs
```diff
@@ -36,33 +36,34 @@ public static class CoreConstants
...
+      public static readonly string TEMPLATE_SYSTEM_DATABASE = "PKSimTemplateDBSystem.mdb";
+      public static readonly string TEMPLATE_USER_DATABASE = "PKSimTemplateDBUser.mdb";
+      public static readonly string TEMPLATE_USER_DATABASE_TEMPLATE = "PKSimTemplateDBUser.template";
+      public static readonly string PK_SIM_DB_FILE = "PKSimDB.mdb";
+      public static readonly string SIM_MODEL_SCHEMA_FILE = "OSPSuite.SimModel.xsd";
+      public static readonly string DIMENSION_FILE = "OSPSuite.Dimensions.xml";
+      public static readonly string PK_PARAMETERS_FILE = "OSPSuite.PKParameters.xml";
+      public static readonly string LOG_4_NET_CONFIG_FILE = "log4net.config.xml";
+      public static readonly string PRODUCT_NAME = "PK-Sim";
+      public static readonly string PRODUCT_NAME_WITH_TRADEMARK = "PK-Sim�";
```
msevestre: Use the opportunity to rename some constants

src/PKSim.Infrastructure/PKSimConfiguration.cs
```diff
@@ -1,67 +1,99 @@
...
-         DimensionFilePath = createApplicationDataPathForFile(CoreConstants.DimensionFile);
-         PKParametersFilePath = createApplicationDataPathForFile(CoreConstants.PKParametersFile);
-         TemplateSystemDatabasePath = createApplicationDataPathForFile(CoreConstants.TemplateSystemDatabase);
-      }
+
+         ApplicationSettingsFilePath = createAppDataPath("ApplicationSettings.xml");
+         ChartLayoutTemplateFolderPath = createApplicationDataOrLocalPathForFolder(CoreConstants.APP_DATA_CHART_LAYOUT_FOLDER_NAME, CoreConstants.LOCAL_CHART_LAYOUT_FOLDER_NAME);
+         TEXTemplateFolderPath = createApplicationDataOrLocalPathForFolder(CoreConstants.APP_DATA_TEX_TEMPLATE_FOLDER_NAME, CoreConstants.LOCAL_TEX_TEMPLATE_FOLDER_NAME);
+
+         PKSimDbPath = createApplicationDataOrLocalPathForFile(CoreConstants.PK_SIM_DB_FILE);
```
msevestre: Loading those guys either from AppData (first) and then from local

src/PKSim.Infrastructure/Serialization/Xml/Serializers/ApplicationSettingsXmlSerializer.cs
```diff
@@ -14,6 +14,7 @@ public ApplicationSettingsXmlSerializer()
 
       public override void PerformMapping()
       {
+         Map(x => x.MoBiPath);
```
msevestre: User can now specify the Path of the MoBi executable in the app settings (will be used as fall back option if not found in registry)

src/PKSim.Matlab/ApplicationStartup.cs
```diff
@@ -14,36 +14,23 @@ internal class ApplicationStartup
...
          if (IoC.Container != null)
             return;
 
          InfrastructureRegister.Initialize();
          var container = IoC.Container;
-         var pksimConfiguration = container.Resolve<IPKSimConfiguration>();
-
-         //path was specified, update the default path
-         if (!string.IsNullOrEmpty(dimensionFilePath))
-            pksimConfiguration.DimensionFilePath = dimensionFilePath;
```
msevestre: We do not need to set those guys in tests now. They can rely on the portable logic so to speak

appveyor-nightly.yml
```diff
@@ -45,6 +45,7 @@ before_build:
 after_build:
   - rake "create_setup[%APPVEYOR_BUILD_VERSION%, %CONFIGURATION%, C:/smartxls.zip, 2.6.2.9]"
   - 7z a setup.zip %APPVEYOR_BUILD_FOLDER%\setup\deploy\*.msi
+  - rake "create_portable_setup[%APPVEYOR_BUILD_VERSION%, %CONFIGURATION%, pksim-portable-setup.zip]"
```
rwmcintosh: there's no actual 'setup' though right?

msevestre: what do you mean?

msevestre: if you mean artifact, then yes it is pushed on appveyor 
https://ci.appveyor.com/project/open-systems-pharmacology-ci/pk-sim/build/artifacts

msevestre: create_setup creates a msi.
create_portable creaes a zip
It's symmetrical in that case

src/PKSim.Infrastructure/Serialization/Xml/Serializers/ApplicationSettingsXmlSerializer.cs
rwmcintosh: User is required to set this path before using MoBi?
Does the suite come with both apps zipped together as well?

msevestre: user is not required to set the path. If using the portable setups
The suite does not have portable options. For now, portable only available for PKSim and MoBi

appveyor-nightly.yml
rwmcintosh: you could call the rake task create_portable instead of create_portable_setup that's all. No setup is really created right?

src/PKSim.Infrastructure/Serialization/Xml/Serializers/ApplicationSettingsXmlSerializer.cs
rwmcintosh: Ok, I thought if they might be zipped together, PKSim could also check a sort of 'parallel' folder for MoBi.

If they aren't distributed that way, then it's not possible.

PR #237, opened 2017-06-28
Fixes #204 - Exception when optimizing relative expression and transferring to simulation

### LINKED ISSUE #204: Exception when optimizing relative expression and transferring to simulation
Optimization of relative expression currently only works for "relative expression (normalized)". This is all fine. However, a bug occurs if you click on" transfer to simulation": this is not possible and an exception is thrown.

rwmcintosh:


src/PKSim.Core/Commands/SetRelativeExpressionFromNormalizedCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+      }
+
+      private double getRelativeExpressionValue(IParameter normalizedParameter, double newNormalizedValue, IParameter relativeExpressionParameter)
+      {
+         return relativeExpressionParameter.Value * newNormalizedValue / normalizedParameter.Value;
+      }
+
+      private IParameter getRelativeExpressionFrom(IParameter normalizedParameter)
+      {
+         return normalizedParameter.ParentContainer.GetAllChildren<IParameter>().FirstOrDefault(isNonNormalizedRelativeExpression);
```
msevestre: This is a little bit confusing here
The non norm parameter is the parameter in the SAME Container having the same name without the NormSuffix
so sthg like
```
name = normalizedParameter.Name.Replace(NormSuffix, "");
return normalizedParameter.ParentContainer.Parameter(name)
```

src/PKSim.Core/Commands/SetRelativeExpressionFromNormalizedCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+
+         Add(new SetRelativeExpressionCommand(relativeExpressionParameter, relativeExpressionValue));
+         Add(new SetParameterValueCommand(_normalizedParameter, _newNormalizedValue));
+
+         base.Execute(context);
+      }
+
+      private double getRelativeExpressionValue(IParameter normalizedParameter, double newNormalizedValue, IParameter relativeExpressionParameter)
+      {
+         return relativeExpressionParameter.Value * newNormalizedValue / normalizedParameter.Value;
```
msevestre: This won't work if the previous value was zero (optimized from zero to non zero)

src/PKSim.Core/Commands/SetRelativeExpressionFromNormalizedCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace PKSim.Core.Commands
+{
+   public class SetRelativeExpressionFromNormalizedCommand : PKSimMacroCommand
+   {
+      private readonly double _newNormalizedValue;
+      private readonly IParameter _normalizedParameter;
+
+      public SetRelativeExpressionFromNormalizedCommand(IParameter normalizedParameter, double normalizedValue)
+      {
+         _normalizedParameter = normalizedParameter;
```
msevestre: Did you check how it looks in the history? I think Description, types etc are all missing.
You may be able to bypass that by setting one of the comand in the macro commadn to visible = false but this needs to be tested

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -127,5 +128,17 @@ public static bool CanBeDefinedAsAdvanced(this IParameter parameter)
 
          return parameter.CanBeVariedInPopulation && !parameter.IsChangedByCreateIndividual;
       }
+
+      public static IEnumerable<IParameter> AllRelatedRelativeExpressions(this IParameter relativeExpressionParameter)
+      {
+         var proteinContainer = relativeExpressionParameter.ParentContainer;
```
msevestre: rename to moleculeContainer

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -127,5 +128,17 @@ public static bool CanBeDefinedAsAdvanced(this IParameter parameter)
...
+         var rootContainer = relativeExpressionParameter.RootContainer;
+         var allParameters = rootContainer.GetAllChildren<IParameter>(x => string.Equals(x.GroupName, CoreConstants.Groups.RELATIVE_EXPRESSION))
```
msevestre: return immediatly

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -127,5 +128,17 @@ public static bool CanBeDefinedAsAdvanced(this IParameter parameter)
...
+            .Where(x => x.BuildingBlockType == PKSimBuildingBlockType.Individual)
+            .Where(x => string.Equals(proteinContainer.Name, x.ParentContainer.Name))
+            .Where(x => x.Name.StartsWith(CoreConstants.Parameter.RelExp))
```
msevestre: we can use x.IsExpression()

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -127,5 +128,17 @@ public static bool CanBeDefinedAsAdvanced(this IParameter parameter)
...
          return parameter.CanBeVariedInPopulation && !parameter.IsChangedByCreateIndividual;
       }
+
+      public static IEnumerable<IParameter> AllRelatedRelativeExpressions(this IParameter relativeExpressionParameter)
+      {
+         var proteinContainer = relativeExpressionParameter.ParentContainer;
+         var rootContainer = relativeExpressionParameter.RootContainer;
+         var allParameters = rootContainer.GetAllChildren<IParameter>(x => string.Equals(x.GroupName, CoreConstants.Groups.RELATIVE_EXPRESSION))
+            .Where(x => x.BuildingBlockType == PKSimBuildingBlockType.Individual)
+            .Where(x => string.Equals(proteinContainer.Name, x.ParentContainer.Name))
```
msevestre: we can use isNamed()

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -127,5 +128,17 @@ public static bool CanBeDefinedAsAdvanced(this IParameter parameter)
 
          return parameter.CanBeVariedInPopulation && !parameter.IsChangedByCreateIndividual;
       }
+
+      public static IEnumerable<IParameter> AllRelatedRelativeExpressions(this IParameter relativeExpressionParameter)
```
msevestre: Rename to AllRelatedRelativeExpressionParameters

msevestre: Should we return an IReadOnlyList? (e.g. TOList at the end). This will always be used with a ToList probably on the caller side otherwise

Issue #246, opened 2017-07-26
Flexible organsim structure to support user defined organs


msevestre:
There are cases where the default PK-Sim Model should be extended or a full species should be added that is not supported out of the box by the Suite. 
Those customization are at the moment not possible because the code knows exactly, and only in a few spots, which organs are to be expected in a PK-Sim organism

The database and the code should be modified to ensure that this list of organ is dynamic instead of static

PR #254, opened 2017-08-15
WIP #246 Flexible organ structure

msevestre:
Update organ type to be loaded from database. That way, we can suppor…t new organs unknown so far

the IsHuman flag is now dynamic as opposed to based on species id


msevestre:
@Yuri05 This is for now just to start a discussion. Here is what I have implemented so far and I'd like to know if you think that something else is missing

The goal is to be able to have a new species added to the database with a different organ structure. PK-Sim should be able to handle unknown organ gracefully. At the same time, the Create Individual Algorithm etc..  should continue working for a species that is not our default human

Here is what I did

1. Removed the usage of the OrganTypes enumeration that was listing organ explicitly. Instead the type of the organ (Tissue, GI etc.. is listed in the database in a new table)
1. Changed the organ construction and loading mechanism to use this flag into account
1. Added a flag IS_HUMAN in the species table that specifies that a species represents a human species. Right now this was strictly based on name but this should not be the case anymore


Now: Is there anything that I am missing and that should also be changed? I am going to try to create my own `Fish` species and see if that works but I'd like to hear your inputs before I carry on

src/PKSim.Core/Services/CreateIndividualAlgorithm.cs
```diff
@@ -401,12 +402,13 @@ private double[] createDefaultOrgansVolumesFrom(Individual individual)
          foreach (var organ in allOrgans)
          {
             _organDensity[iOrganIndex] = organ.Parameter(CoreConstants.Parameter.DENSITY).Value;
-            organVolumes[iOrganIndex] = organ.Parameter(CoreConstants.Parameter.VOLUME).Value;
+            organVolumes[iOrganIndex] = organ.Parameter(Constants.Parameters.VOLUME).Value;
+
 
-            if (organ.OrganType == OrganType.Skin)
```
msevestre: These were the only places where the organ specific type was used.

src/PKSim.Core/Mappers/IndividualToPopulationSettingsMapper.cs
```diff
@@ -37,7 +37,7 @@ public RandomPopulationSettings MapFrom(Individual individual)
             populationSettings.AddGenderRatio(new GenderRatio {Gender = gender, Ratio = 100 / genderCount});
          }
 
-         if (population.IsAgeDependent)
+         if (individual.IsAgeDependent)
```
Yuri05: what is the idea behind this change?

src/PKSim.Core/Services/CreateIndividualAlgorithm.cs
```diff
@@ -33,15 +33,12 @@ public interface ICreateIndividualAlgorithm
 
    public class CreateIndividualAlgorithm : ICreateIndividualAlgorithm
    {
+      private const double COVERGENCE_TOLERANCE = 1e-6;
```
Yuri05: CO**N**VERGENCE_TOLERANCE

src/PKSim.Core/Mappers/IndividualToPopulationSettingsMapper.cs
msevestre: This is strictly equivalent. The `IsAgeDependent` property is new in the individual so we do not need to go to the `SpeciesPopulation` which is here called population. That way the code is also symetrical with the `IsPreterm`

src/PKSim.Core/Services/CreateIndividualAlgorithm.cs
msevestre: done

src/PKSim.Infrastructure/ORM/Mappers/FlatSpeciesToCreatureMapper.cs
```diff
@@ -1,5 +1,7 @@
-﻿using System.Linq;
```
Yuri05: we should rename the file, lol

src/PKSim.Infrastructure/ProjectConverter/v7_2/Converter710To720.cs
```diff
@@ -105,7 +128,7 @@ private void addBSAParameterValues(Population population)
 
       private void addBSACalculationMethod(Individual individual)
       {
-         if (!individual.IsHuman)
+         if (!individual.Species.IsHuman)
```
Yuri05: should it be _individual.IsHeigthDependent_ instead of _individual.IsHuman_?

src/PKSim.Infrastructure/ORM/Mappers/FlatSpeciesToCreatureMapper.cs
msevestre: done

Yuri05:
>Here is what I did
>
>   Removed the usage of the OrganTypes enumeration that was listing organ explicitly. Instead the type of the organ (Tissue, GI etc.. is listed in the database in a new table)
    Changed the organ construction and loading mechanism to use this flag into account
    Added a flag IS_HUMAN in the species table that specifies that a species represents a human species. Right now this was strictly based on name but this should not be the case anymore

>Now: Is there anything that I am missing and that should also be changed? I am going to try to create my own Fish species and see if that works but I'd like to hear your inputs before I carry on

Seems complete to me

PR #290, opened 2017-08-29
Fixes #260 export of formulation snapshot

### LINKED ISSUE #260: Export of Formulation snapshot (json)

msevestre:


msevestre:
@Yuri05 This is the very beginning of the implementation. In order to get the formulation export going, I needed to also implement part of the infrastructure to export parameter and add context menu.

src/PKSim.Core/Snapshots/Mappers/FormulationMapper.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace PKSim.Core.Snapshots.Mappers
+{
+   public class FormulationMapper : ParameterContainerSnapshotMapperBase<ModelFormulation, SnapshotFormulation>
+   {
+      public FormulationMapper(ParameterMapper parameterMapper) : base(parameterMapper)
+      {
+      }
+
+      public override SnapshotFormulation MapToSnapshot(ModelFormulation modelFormulation)
```
msevestre: I think this should not be more complicated than that for basic container objects such as Events, Formulation, SimpleProtocol etc...

src/PKSim.Core/Snapshots/Mappers/ParameterMapper.cs
```diff
@@ -0,0 +1,37 @@
...
+      {
+         _tableFormulaMapper = tableFormulaMapper;
+      }
+
+      public override SnapshotParameter MapToSnapshot(IParameter modelParameter)
+      {
+         var snapshotParameter = new SnapshotParameter();
+         MapBaseProperties(modelParameter, snapshotParameter);
+         //override description that might have been set since parameter in PK-Sim all come from DB
+         snapshotParameter.Description = null;
```
msevestre: @Yuri05 We do not want to serialize the PK-Sim parameter description (or do we??) since they all come from the database. But we can decide otherwise easily

src/PKSim.Core/Snapshots/Mappers/ParameterMapper.cs
```diff
@@ -0,0 +1,37 @@
...
+         snapshotParameter.Value = modelParameter.ValueInDisplayUnit;
+         snapshotParameter.Unit = SnapshotValueFor(modelParameter.DisplayUnit.Name);
+         snapshotParameter.ValueDescription = SnapshotValueFor(modelParameter.ValueDescription);
+         snapshotParameter.TableFormula = mapFormula(modelParameter.Formula);
```
msevestre: The only type of formula that can be created in PKSim for a parameter is a table formula. I suggest to handle that explicitly for the moment. If this is a too simple approach that creates friction along the way, we can adjust it

src/PKSim.Core/Snapshots/Mappers/SnapshotMapper.cs
```diff
@@ -0,0 +1,55 @@
...
+      object MapToModel(object snapshot);
+   }
+
+   public class SnapshotMapper : ISnapshotMapper
+   {
+      private readonly List<ISnapshotMapperSpecification> _allMappers;
+
+      public SnapshotMapper(IRepository<ISnapshotMapperSpecification> snapshotMapperRepository)
+      {
+         _allMappers = snapshotMapperRepository.All().ToList();
```
msevestre: We need to implement one mapper for each <Model, Snapshot> combination

src/PKSim.Core/Snapshots/Mappers/SnapshotMapperBase.cs
```diff
@@ -0,0 +1,62 @@
...
+      where TModel: IObjectBase
+   {
+      public virtual object MapToSnapshot(object model)
+      {
+         return MapToSnapshot(model.DowncastTo<TModel>());
+      }
+
+      public virtual object MapToModel(object snapshot)
+      {
+         throw new NotImplementedException();
```
msevestre: This will be implemented next when doing the import of a formulation snapshot

msevestre:
@Yuri05 Once this is merged, I'll start with import of snapshot as building block. Once the loop is closed, we can implement all other building blocks

src/PKSim.Core/Snapshots/Mappers/ParameterMapper.cs
Yuri05: Agree. Description should not be saved

msevestre:
I did not want to close that!

msevestre:
@Yuri05 I am done with #263 but I need this one to be merged first. Let me know if you have more comments on that otherwise please merge

src/PKSim.Core/Snapshots/Mappers/SnapshotMapper.cs
```diff
@@ -0,0 +1,55 @@
...
+      /// <summary>
+      ///    Given a <paramref name="snapshot" /> object, returns the corresponding model.
+      /// </summary>
+      /// <exception cref="SnapshotNotFoundException">
+      ///    is thrown if a snapshot could not be found for the given
+      ///    <paramref name="snapshot" />
+      /// </exception>
+      object MapToModel(object snapshot);
+   }
+
```
Yuri05: Why `object MapTo`? Shouldn't we define it like
```
public interface ISnapshotMapper<TModel, TSnapshot>
{
    TSnapshot MapToSnapshot(TModel model);
    TModel MapToModel(TSnapshot snapshot);
}
```
?

msevestre: This would not help. I need a interface that is common to all mappers so that it can be called in a very simple way.  This can be seen in the SnapshotMapper implementation. It finds the matching mapper for a given model type

msevestre: I should add that the base implementation has the generic method that you described. The interface  not however. It would be too specific otherwise

msevestre:
@Yuri05 I need to merge to continue moving forward here. If you have any more comments, please update. (the next PR is coming shortly)

PR #296, opened 2017-08-31
Fixes #263 Import of formulation building block from snapshot (json)

### LINKED ISSUE #263: Import of formulation building block from  snapshot (json)

msevestre:


msevestre:
@Yuri05 Ok this closes the loop. Reimport from a formulation (I am only dealing with happy scenarios at the moment, where I am actually importing a file that matches the type)

The menu will probably be visible only when the -developer flag is available but this is convenient now to test

src/PKSim.Core/Snapshots/Mappers/FormulationMapper.cs
```diff
@@ -1,22 +1,49 @@
...
+      {
+         var template = _formulationRepository.FormulationBy(formulationSnapshot.FormulationType);
+         var formulation = _cloner.Clone(template);
+
+         MapSnapshotPropertiesIntoModel(formulationSnapshot, formulation);
+         foreach (var snapshotParameter in formulationSnapshot.Parameters)
+         {
+            var modelParameter = formulation.Parameter(snapshotParameter.Name);
+
+            if (modelParameter == null)
```
msevestre: I think that if we are trying to load a snapshot with a parameter that we do not know, something is clearly outdated and we should throw an exception.
I don't think we should support project conversion for snapshot too. We need to keep them uptodate

src/PKSim.Core/Snapshots/Mappers/FormulationMapper.cs
```diff
@@ -1,22 +1,49 @@
...
+               throw new SnapshotParameterNotFoundException(snapshotParameter.Name, formulationSnapshot.FormulationType);
+
+            _parameterMapper.UpdateParameterFromSnapshot(modelParameter, snapshotParameter);
```
msevestre: We do not create parameters. We always update an existing one

src/PKSim.Core/Snapshots/Mappers/ParameterMapper.cs
```diff
@@ -17,15 +19,36 @@ public ParameterMapper(TableFormulaMapper tableFormulaMapper)
...
+            Value = modelParameter.ValueInDisplayUnit,
+            Unit = SnapshotValueFor(modelParameter.DisplayUnit.Name),
+            ValueDescription = SnapshotValueFor(modelParameter.ValueDescription),
+            TableFormula = mapFormula(modelParameter.Formula)
+         };
+      }
+
+      public override IParameter MapToModel(SnapshotParameter snapshot)
+      {
+         throw new NotSupportedException("Parameter should not be created from snapshot. Instead use the update method");
```
msevestre: See above. No creation from scratch. Instead we always update an existing one

src/PKSim.Core/Snapshots/Mappers/SnapshotMapper.cs
```diff
@@ -38,18 +49,28 @@ public SnapshotMapper(IRepository<ISnapshotMapperSpecification> snapshotMapperRe
...
+
+      public Type SnapshotTypeFor<T>()
+      {
+         var modelType = typeof(T);
+         return mapperFor(modelType).SnapshotTypeFor<T>();
+      }
+
+      private ISnapshotMapper mapperFor(Type modelOrSnapshotType)
+      {
+         var mapper =  _allMappers.FirstOrDefault(x => x.IsSatisfiedBy(modelOrSnapshotType));
```
msevestre: See discussion from previous PR. This is where the non generic interface is required so that this object can old a reference on ALL snapshots mappers

src/PKSim.Core/Snapshots/Mappers/SnapshotMapperBase.cs
```diff
@@ -22,30 +22,45 @@ public virtual object MapToSnapshot(object model)
 
       public virtual object MapToModel(object snapshot)
       {
-         throw new NotImplementedException();
+         return MapToModel(snapshot.DowncastTo<TSnapshot>());
```
msevestre: Here we always cast to the type that was specified in the actual implementation

src/PKSim.Core/Snapshots/Mappers/TableFormulaMapper.cs
```diff
@@ -1,28 +1,62 @@
...
+         tableFormula.XDimension = _dimensionRepository.DimensionByName(snapshotTableFormula.XDimension);
+         tableFormula.XDisplayUnit = tableFormula.XDimension.Unit(UnitValueFor(snapshotTableFormula.XUnit));
+         tableFormula.XName = snapshotTableFormula.XName;
+
+         tableFormula.Dimension = _dimensionRepository.DimensionByName(snapshotTableFormula.YDimension);
+         tableFormula.YDisplayUnit = tableFormula.Dimension.Unit(UnitValueFor(snapshotTableFormula.YUnit));
+         tableFormula.YName = snapshotTableFormula.YName;
+
+         snapshotTableFormula.Points.Each(p => tableFormula.AddPoint(valuePointFrom(tableFormula, p)));
+         return tableFormula;
```
msevestre: This is going to be the typical work to do for all objects. Map one way and the other

msevestre: Probably I will have to extract some useful common method in the base class

src/PKSim.Core/Snapshots/Services/SnapshotSerializer.cs
```diff
@@ -1,12 +1,16 @@
...
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Schema;
 
 namespace PKSim.Core.Snapshots.Services
 {
    public interface ISnapshotSerializer
    {
       void Serialize(object snapshot, string fileName);
-      T Deserialize<T>(string fileName);
+      object[] DeserializeAsArray(string fileName, Type snapshotType);
```
msevestre: I am just implementing what I need now. I want to ensure that we can deserialize both a single item and an array of items.

src/PKSim.Core/Snapshots/Services/SnapshotSerializer.cs
```diff
@@ -24,9 +28,21 @@ public void Serialize(object snapshot, string fileName)
          File.WriteAllText(fileName, JsonConvert.SerializeObject(snapshot, Formatting.Indented, _settings));
       }
 
-      public T Deserialize<T>(string fileName)
+      public object[] DeserializeAsArray(string fileName, Type snapshotType)
       {
-         return JsonConvert.DeserializeObject<T>(File.ReadAllText(fileName), _settings);
+         var deserializedSnapshot = JsonConvert.DeserializeObject(File.ReadAllText(fileName), _settings);
+
+         //TODO. We should validate that the json actually matches our type. If not, we could throw an explicit error
```
msevestre: this is an internal Json object. In theory I could have selected a file with compound information while trying to load a formulation. This is where we will use json schema to ensure that the json structure matches our object. If not, we will throw an exception with hopefully some useful info

tests/PKSim.Tests/Core/FormulationMapperSpecs.cs
```diff
@@ -73,4 +81,58 @@ public void should_save_the_visible_formulation_parameters_only()
          _snapshot.Parameters.ExistsByName(_hiddenParameter.Name).ShouldBeFalse();
       }
    }
+
+   public class When_mapping_a_valid_formulation_snapshot_to_a_formulation : concern_for_FormulationMapper
```
msevestre: A lot of tests here making sure every combination I could think of is tested

src/PKSim.Core/Snapshots/Mappers/FormulationMapper.cs
Yuri05: Exception or should we just write a warning into logfile and continue? I would vote for the warning.
At the end the user still can decide if the project may be used or should be thrown away.

And yes, would I also think we don't need project conversion for that.

Yuri05: E.g. if starting in a batch mode.
If started in PK-Sim it might be not a log file but a summary (like in the old 4.2=>5.x project conversion)

msevestre: I think that loading something that does not match the template can only be wrong and lead to wrong result. How should a user know what to do?

msevestre: Besides those file should never be edited by hand in my opinion. Rather they should be generated from the tool

msevestre: I am going to create an entry for that to continue the discussion. We can always add this if we need

Yuri05: ok

PR #298, opened 2017-08-31
Fixes #261 Fixes #264 import export of event building block

### LINKED ISSUE #261: Export of Event snapshot (json)

### LINKED ISSUE #264: Import of event building block from  snapshot (json)

msevestre:


src/PKSim.Core/Snapshots/Mappers/EventMapper.cs
```diff
@@ -0,0 +1,37 @@
+﻿using OSPSuite.Core.Domain;
+using PKSim.Core.Model;
+using SnapshotEvent = PKSim.Core.Snapshots.Event;
+using ModelEvent = PKSim.Core.Model.PKSimEvent;
+
+
+namespace PKSim.Core.Snapshots.Mappers
+{
+   public class EventMapper : ParameterContainerSnapshotMapperBase<ModelEvent, SnapshotEvent>
```
msevestre: Those are going to be trivial now (Simple Container parameters)

src/PKSim.Core/Snapshots/Mappers/EventMapper.cs
```diff
@@ -0,0 +1,37 @@
...
+         MapModelPropertiesIntoSnapshot(modelEvent, snapshotEvent);
+         MapVisibleParameters(modelEvent, snapshotEvent);
+         snapshotEvent.Template = modelEvent.TemplateName;
+         return snapshotEvent;
+      }
+
+      public override ModelEvent MapToModel(SnapshotEvent snapshotEvent)
+      {
+         var modelEvent = _eventFactory.Create(snapshotEvent.Template);
+
```
msevestre: Some spacing here that I will remove in next PR

Yuri05:
Questions:
* The idea is to write only parameters which **can be changed by user**, into the snapshot file, correct?
So should we define and use _Map**Editable**Parameters_ instead of _Map**Visible**Parameters_ everywhere?
* Can we have any situation where a non-visible (or non-editable) parameter is overridden in BB or simulation? E.g. some import process. If so, mapping only visible/editable params might be not sufficient.

msevestre:
First, the code has changed already. Since I am using BDD approach, I am only implementing what I need at the moment of the implementation. 

>The idea is to write only parameters which can be changed by user
I don't think so. My understanding is that it depends probably on the building block and the type of parameters. But in general, it should probably be `Parameter overwritten by user"

* Events:
We can decided to write all editable parameters (indeed better than visible)
But we could also decide to write all parameters that differs from default value. For example, the number of calories per meal etc will hardly ever be changed by a user. Should we save it or not.

* For individual:

The concept of editable only does not make sense since we have so many parameters that the user never changes. Rather, this is the concept of a parameter whose value has changed from defaults

So in short, editable/visible is definitely not sufficient. The concept will be refined as we advanced in the implementation. 
My vote would be to only write parameters that have been changed by the user explicitely

Yuri05:
>My vote would be to only write parameters that have been changed by the user explicitely

Agree (for all types of BBs)

msevestre:
This concept is starting to take shape in the implementation of compound.

PR #300, opened 2017-09-01
Fixes #70 Population age limit #70

### LINKED ISSUE #70: Population age limit
The age limit of each population should be independent based on the underlying information in the database. Only individuals/populations up to the informed range end in the database should be generated by the user of the platform.

Yuri05:


Yuri05:
_Age_ parameter moved from CONTAINER_PARAMETER_VALUES to CONTAINER_PARAMETER_CURVES in order to enable setting min/max parameter bounds depending on the selected population

msevestre:
The build is failing....

msevestre:
Check appveyor log

msevestre:
Not sure how this is going to work. Age parameter is required to read in the curve table.... I think there is a dependency issue here

tests/PKSim.Tests/IntegrationTests/DatabaseUpdateSpecs.cs
```diff
@@ -43,8 +102,7 @@ public void should_set_the_body_surface_area_parameter_of_an_individual_to_can_b
       [Observation]
       public void only_the_species_human_should_have_the_flag_is_human()
       {
-         var flatSpeciesRepository = IoC.Resolve<IFlatSpeciesRepository>();
```
msevestre: Why did you move those guys global? They are only required locally correct?

Yuri05: just to have all IoC.Resolves in one place. not really required here, you're right

msevestre: Ok cool. I like it

msevestre:
@Yuri05 Are you aware that the build is broken?

Yuri05:
yes. will fix it as soon as I can

Yuri05:
>Why is this required. Age is set 2 lines after

The problem is that apparently _in the batch mode_ `originData.Age` is null and the next line crashes
`var meanAgeParameter = _individualModelTask.MeanAgeFor(originData);`

msevestre:
Why was it ok before then?

Yuri05:
Age was previously formula parameter and now is distributed parameter

Von: Michael Sevestre [mailto:notifications@github.com]
Gesendet: Mittwoch, 4. Oktober 2017 14:33
An: Open-Systems-Pharmacology/PK-Sim
Cc: Juri Solodenko; Mention
Betreff: Re: [Open-Systems-Pharmacology/PK-Sim] Fixes #70 Population age limit #70 (#300)

Why was it ok before then?

Issue #305, opened 2017-09-01
Export of compound snapshot (json)


msevestre:

msevestre:
This is the snapshot structure, encompassing alternatives, processes etc..

```
{
  "Name": "Test",
  "IsSmallMolecule": true,
  "PlasmaProteinBindingPartner": "Glycoprotein",
  "Lipophilicity": [
    {
      "Name": "Lipo",
      "IsDefault": true,
      "Parameters": [
        {
          "Name": "Lipophilicity",
          "Value": 1.0,
          "Unit": "Log Units"
        }
      ]
    },
    {
      "Name": "lipo2",
      "IsDefault": false,
      "Parameters": [
        {
          "Name": "Lipophilicity",
          "Value": 2.0,
          "Unit": "Log Units"
        }
      ]
    }
  ],
  "FractionUnbound": [
    {
      "Name": "Fu1",
      "IsDefault": false,
      "Species": "Monkey",
      "Parameters": [
        {
          "Name": "Fraction unbound (plasma, reference value)",
          "Value": 0.3
        }
      ]
    },
    {
      "Name": "Fu2",
      "IsDefault": true,
      "Species": "Dog",
      "Parameters": [
        {
          "Name": "Fraction unbound (plasma, reference value)",
          "Value": 0.4
        }
      ]
    }
  ],
  "Solubility": [
    {
      "Name": "Sol1",
      "IsDefault": true,
      "Parameters": [
        {
          "Name": "Solubility at reference pH",
          "Value": 8.0,
          "Unit": "mg/l"
        },
        {
          "Name": "Reference pH",
          "Value": 9.0
        },
        {
          "Name": "Solubility gain per charge",
          "Value": 10.0
        }
      ]
    }
  ],
  "IntestinalPermeability": [
    {
      "Name": "IntestinalPerm1",
      "IsDefault": false,
      "Parameters": [
        {
          "Name": "Specific intestinal permeability (transcellular)",
          "Value": 1.0,
          "Unit": "cm/s"
        }
      ]
    }
  ],
  "Permeability": [
    {
      "Name": "Spec Int 1",
      "IsDefault": false,
      "Parameters": [
        {
          "Name": "Permeability",
          "Value": 1.0,
          "Unit": "cm/s"
        }
      ]
    }
  ],
  "PkaTypes": [
    {
      "Type": "Base",
      "Pka": 6.0
    },
    {
      "Type": "Acid",
[...]

PR #315, opened 2017-09-07
268 305 import export compound

msevestre:


msevestre:
This was a serious chunk of work here. 
Exporting compound is by far the most complicated use case (as far as building block is concerned)

msevestre:
Example of json file (that can be used for or molecule catalogue) can be found here
https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/305

src/PKSim.Core/Snapshots/Mappers/ParameterContainerSnapshotMapperBase.cs
```diff
@@ -0,0 +1,78 @@
...
+
+      protected virtual void AddModelParametersToSnapshot(TModel model, TSnapshot snapshot)
+      {
+         AddParametersToSnapshot(model.AllParameters(ParameterHasChanged), snapshot);
+      }
+
+      protected virtual bool ParameterHasChanged(IParameter parameter)
+      {
+         //TODO Use ValueOrigin state when implemented. It should be != than PKSim default;
+         var canBeEdited = parameter.Visible && parameter.Editable && parameter.Formula.IsConstant();
```
msevestre: Following or discussion on event. There is this method to check if a parameter value was changed
Right now this cannot be estimated easily but this will be possible soon once we have the `ValueOriginMode` in Parameters

msevestre: This check to 0 is non sense. But is useful for now to remove some parameters

PR #325, opened 2017-09-11
Fixes #285 export project snapshot

### LINKED ISSUE #285: Export of full project snapshot (json)

msevestre:


msevestre:
I added Rob as reviewer as well because he will start working on the project this week 
A lot of changes in this PR because  I removed the IPKsimProject interface. We already have an IProject interface in core that describes the common interface between MoBi and PKSim. The PKSimProject is already specific enough. There is definitely no need for this abstraction here. (this is also consistent with the fact that we do not have any interfaces for all other domain types in PKSim)

src/PKSim.UI/Views/ObservedDataToCompoundMappingView.cs
```diff
@@ -1,68 +0,0 @@
...
-using DevExpress.XtraGrid.Views.Base;
-using PKSim.Assets;
-using PKSim.Presentation.DTO.ObservedData;
-using PKSim.Presentation.Presenters;
-using PKSim.Presentation.Views;
-using OSPSuite.UI.Views;
-
-namespace PKSim.UI.Views
-{
-   public partial class ObservedDataToCompoundMappingView : BaseModalView, IObservedDataToCompoundMappingView
```
msevestre: This was not used anymore (left over for 4.2 project conversion)

src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -1,19 +1,47 @@
...
+
+      public ProjectMapper(IExecutionContext executionContext)
+      {
+         _executionContext = executionContext;
+      }
+
       public override SnapshotProject MapToSnapshot(ModelProject project)
       {
-         throw new System.NotImplementedException();
+         return SnapshotFrom(project, snapshot =>
```
msevestre: This is the new code here. Only implemented for the building blocks that have been defined as snapshot for now. This will change over time obviously as we add more export to snapshots

PR #327, opened 2017-09-12
Fixes #271 export of population snapshot

### LINKED ISSUE #271: Export of population snapshot (json)

msevestre:
* Removed the usage of IAdvancedParameter
* Export of RandomPopulation only (we can recreate the values on the fly). Exporting other type of populations to snapshot (Imported or MoBi) does not make sense because we would have to save also all data points, which defeats the purpose of the snapshot
* added a "." at the end of all error messages. This was not consistent


src/PKSim.Core/Snapshots/Mappers/PopulationMapper.cs
```diff
@@ -0,0 +1,89 @@
...
+
+      private void mapPopulationProperties(SnapshotPopulation snapshot, ModelPopulation population)
+      {
+         switch (population)
+         {
+            case RandomPopulation randomPopulation:
+               mapRandomPopulationProperties(snapshot, randomPopulation);
+               break;
+            default:
+               throw new OSPSuiteException(PKSimConstants.Error.PopulationSnapshotOnlySupportedForRandomPopulation);
```
msevestre: Exception if we are not dealing with a random population. This is an OSPSuiteException which will result in NO stack trace

msevestre: @Yuri05 Do you agree?

src/PKSim.Core/Model/ParameterRange.cs
```diff
@@ -32,15 +19,20 @@ public ParameterRange()
...
+      public ParameterRange()
       {
-         get { return MinValue.HasValue && MaxValue.HasValue && MinValue.Value == MaxValue.Value; }
+         Rules = new BusinessRuleSet();
+         Rules.Add(AllRules.MinLessThanMax);
+         Rules.Add(AllRules.MaxGreaterThanMin);
+         Rules.Add(AllRules.MinGreaterThanDbMin);
+         Rules.Add(AllRules.MaxLessThanDbMax);
+         Rules.Add(AllRules.MinLessThanDbMax);
+         Rules.Add(AllRules.MaxGreaterThanDbMin);
```
Yuri05: I think `MinLessThanMax `& `MinGreaterThanDbMin ` & `MaxLessThanDbMax `implies both `MaxGreaterThanDbMin`
and `MinLessThanDbMax`

msevestre: ???

src/PKSim.Core/Model/AdvancedParameter.cs
```diff
@@ -153,7 +120,11 @@ public IDistributedParameter DistributedParameter
             Add(value);
          }
       }
-
+      
+      /// <summary>
+      ///    Reset the genrator with the value of the seed.
```
rwmcintosh: generator

src/PKSim.Core/Model/ParameterRange.cs
msevestre: This code has nothing to do with the PR BTW

msevestre: I did not see the implies. This might be correct but the rules should be symetrical so that it does not feel like something missing.

Yuri05: Well, it's marked as new code by GitHub.

I  mean e.g. if both `MinLessThanMax ` and `MinGreaterThanDbMin ` rules are satisfied, than `MaxGreaterThanDbMin` is automatically satisfied, so why would we need this additional rule?

Same for  `MinLessThanDbMax`

src/PKSim.Core/Snapshots/Mappers/AdvancedParameterMapper.cs
```diff
@@ -0,0 +1,33 @@
...
+   {
+      public AdvancedParameterMapper(ParameterMapper parameterMapper) : base(parameterMapper)
+      {
+      }
+
+      public override SnapshotAdvancedParameter MapToSnapshot(ModelAdvancedParameter advancedParameter)
+      {
+         return SnapshotFrom(advancedParameter, snapshot =>
+         {
+            //the parameter path is what identified the advanced parameter. The name is not used anywhere. We will use the path as key
```
rwmcintosh: What do you mean by 'path as key'?
Are you using the Name property here to make the json more easy to understand? As in, during deserialization, will it be used?

src/PKSim.Core/Snapshots/Parameter.cs
```diff
@@ -10,7 +10,7 @@ public class Parameter : SnapshotBase
 
    public class LocalizedParameter : Parameter
    {
-      //Full path of parameter in continer hierarchie
+      //Full path of parameter in container hierarchie
```
rwmcintosh: hierarchie => hierarchy

src/PKSim.Core/Model/ParameterRange.cs
rwmcintosh: I think it's marked that way because the code moved.

msevestre: @Yuri05  I don't think this is true 
it could be that there is a DBMin but no min is defined by the user. In that case we need to check that the Max is greater than the DBMin
But at any rate, this code is old and was not changed (Aside from formatting probably)

Yuri05: ok

msevestre:
OK I Am going to merge that and ensure that those comments are addressed in the next PR

Yuri05:
>Export of RandomPopulation only (we can recreate the values on the fly). Exporting other type of populations to snapshot (Imported or MoBi) does not make sense because we would have to save also all data points, which defeats the purpose of the snapshot

For release notes etc.: we should write down all the restrictions of snapshot creation and snapshot import

PR #330, opened 2017-09-13
Fixes #273 export observed data

### LINKED ISSUE #273: Export of observed data snapshot (json)
- Repository Extended properties
- Columns
- Column Extended Properties
- QuantityInfo
- DataInfo
- RelatedColumns

rwmcintosh:


src/PKSim.Core/Snapshots/Mappers/ExtendedPropertiesMapper.cs
```diff
@@ -0,0 +1,39 @@
...
+      private List<SnapshotExtendedProperty> mapExtendedPropertiesFrom(ModelExtendedProperties extendedProperties)
+      {
+         return extendedProperties.Any() ? extendedProperties.Select(property => _extendedPropertyMapper.MapToSnapshot(property)).ToList() : null;
+      }
+
+      public override ModelExtendedProperties MapToModel(SnapshotExtendedProperties snapshot)
+      {
+         var extendedProperties = new ModelExtendedProperties();
+
+         extendedProperties.AddRange(snapshot.ListOfExtendedProperties.Select(propertySnapshot => _extendedPropertyMapper.MapToModel(propertySnapshot)));
```
msevestre: Select(_extendedPropertyMapper.MapToModel)

src/PKSim.Core/Snapshots/Mappers/ExtendedPropertyMapper.cs
```diff
@@ -0,0 +1,72 @@
...
+      }
+
+      public override ModelExtendedProperty MapToModel(SnapshotExtendedProperty snapshot)
+      {
+
+         ModelExtendedProperty property;
+
+         if (snapshot.Type == typeof(string))
+         {
+            var stringProperty = new ExtendedProperty<string>();
```
msevestre: This is the same code 4 times. Can't you create  a method with a <T> parameter? Am I missing something?

msevestre: and a Func to convert object to T obviously

src/PKSim.Core/Snapshots/Mappers/QuantityInfoMapper.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace PKSim.Core.Snapshots.Mappers
+{
+   public class QuantityInfoMapper : SnapshotMapperBase<ModelQuantityInfo, SnapshotQuantityInfo>
+   {
+      public override SnapshotQuantityInfo MapToSnapshot(ModelQuantityInfo quantityInfo)
+      {
+         return SnapshotFrom(quantityInfo, snapshot =>
+         {
+            snapshot.OrderIndex = quantityInfo.OrderIndex;
+            snapshot.Path = quantityInfo.Path.ToList();
```
msevestre: Path as string? instead of string[]?

src/PKSim.Core/Snapshots/QuantityInfo.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+
+namespace PKSim.Core.Snapshots
+{
+   public class QuantityInfo : SnapshotBase
+   {
+      public List<string> Path { set; get; }
```
msevestre: string Path{get;set;} no?

src/PKSim.Core/Snapshots/ExtendedProperties.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace PKSim.Core.Snapshots
+{
+   public class ExtendedProperties : SnapshotBase
+   {
+      public List<ExtendedProperty> ListOfExtendedProperties { set; get; }
```
msevestre: Look at how the CalculationMethodCache is defined and serialized. You do not need to have a sub properties here. the json will just be complicated for nothing. Instead just defined your type ExtendedProperties as IEnuerable

Or remove that type alltogether. It really does not hold anything aside from the list.

src/PKSim.Core/Snapshots/Mappers/ObservedDataMapper.cs
```diff
@@ -0,0 +1,48 @@
...
+         {
+            snapshot.ExtendedProperties = mapExtendedProperties(dataRepository.ExtendedProperties);
+            snapshot.Columns = mapColumns(dataRepository.AllButBaseGrid().Where(column => !dataRepository.ColumnIsInRelatedColumns(column)));
+            snapshot.BaseGrid = _dataColumnMapper.MapToSnapshot(dataRepository.BaseGrid);
+         });
+      }
+
+      private List<DataColumn> mapColumns(IEnumerable<ModelDataColumn> dataRepositoryColumns)
+      {
+         return dataRepositoryColumns.Select(dataColumn => _dataColumnMapper.MapToSnapshot(dataColumn)).ToList();
```
msevestre: Select(_dataCOlumnMapper.MatoToSnapshot)

rwmcintosh:
The related columns are referenced twice in the data repository, once in the columns enumeration, and again from the data column they are related to.

In order to remove duplication, I wanted to only store the json in one place. Since I can store the column and the relation this way I decided to just keep that reference.

Keeping the related columns in the data column enumeration would mean I would have to separately store how the columns are related. This way, I don't have to store that detail.

msevestre:
I understand. This seems a little bit strange to me at first but if this makes life easier... why not

src/PKSim.Core/Snapshots/DataColumn.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace PKSim.Core.Snapshots
+{
+   public class DataColumn : SnapshotBase
+   {
+      public string Id { get; set; }
+      public List<DataColumn> RelatedColumns { get; set; }
+      public QuantityInfo QuantityInfo { get; set; }
+      public DataInfo DataInfo { get; set; }
+      public List<float> Values { get; set; }
+      public string DisplayUnit { get; set; }
```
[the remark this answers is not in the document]
rwmcintosh: They look like they are the same. I'll remove the DataInfo one and keep the unit on DataColumn

PR #333, opened 2017-09-14
267 import of population building block

msevestre:


src/PKSim.Core/Model/Individual.cs
```diff
@@ -33,14 +33,7 @@ public Individual() : base(PKSimBuildingBlockType.Individual)
       /// </summary>
       public virtual SpeciesPopulation Population => OriginData.SpeciesPopulation;
 
-      /// <summary>
-      ///    Create a new seed for the individual
-      /// </summary>
-      public virtual void GenerateSeed()
```
msevestre: This method is not used anymore so it was removed

src/PKSim.Core/Model/RandomPopulationFactory.cs
```diff
@@ -2,27 +2,27 @@
...
       /// <summary>
-      ///    Create a population using the specified <paramref name="populationSettings" />. The
-      ///    <paramref name="cancellationToken" />
-      ///    allows the caller to cancel the task
+      ///    Create a population using the specified <paramref name="populationSettings" />.
+      ///    The <paramref name="cancellationToken" /> allows the caller to cancel the task. If the <paramref name="seed" /> is
+      ///    provided, it will be set in the created population to generate random values
       /// </summary>
-      Task<RandomPopulation> CreateFor(RandomPopulationSettings populationSettings, CancellationToken cancellationToken);
+      Task<RandomPopulation> CreateFor(RandomPopulationSettings populationSettings, CancellationToken cancellationToken, int? seed = null);
```
msevestre: added the seed as parameter. It is used if not null

msevestre: This allow to reproduce results

src/PKSim.Core/Services/CreateIndividualAlgorithm.cs
```diff
@@ -28,7 +28,8 @@ public interface ICreateIndividualAlgorithm
       /// <param name="minWeight">minimal weight allowed for the randomized individual. Null means not constrained</param>
       /// <param name="maxWeight">maximal weight allowed for the randomized individual. Null means not constrained</param>
       /// <param name="allIndividualParameters">All individual parameters defined in the individual (optimization purpose only)</param>
-      void Randomize(Individual individual, IParameter weightParameter, double? minWeight, double? maxWeight, IEnumerable<IParameter> allIndividualParameters);
+      /// <param name="randomGenerator">Random generator used to create values</param>
+      void Randomize(Individual individual, IParameter weightParameter, double? minWeight, double? maxWeight, IEnumerable<IParameter> allIndividualParameters, RandomGenerator randomGenerator);
```
msevestre: Passing the RandomGenerator as parameter is a way to ensure that we can reproduce the same population outputs (comes from the population)

src/PKSim.Core/Services/DistributedMuSigma.cs
```diff
@@ -159,10 +149,7 @@ public ConstantMuSigma(double value)
          _value = value;
       }
 
-      public double DefaultValue
-      {
-         get { return _value; }
-      }
+      public double DefaultValue => _value;
```
msevestre: Only some formatting in this file

src/PKSim.Core/Snapshots/Compound.cs
```diff
@@ -12,13 +12,13 @@ public class Compound : ParameterContainerSnapshotBase
...
       public bool IsSmallMolecule { get; set; }
       public string PlasmaProteinBindingPartner { get; set; }
-      public List<Alternative> Lipophilicity { get; set; } = new List<Alternative>();
-      public List<Alternative> FractionUnbound { get; set; } = new List<Alternative>();
-      public List<Alternative> Solubility { get; set; } = new List<Alternative>();
-      public List<Alternative> IntestinalPermeability { get; set; } = new List<Alternative>();
-      public List<Alternative> Permeability { get; set; } = new List<Alternative>();
-      public List<PkaType> PkaTypes { get; set; } = new List<PkaType>();
-      public List<CompoundProcess> Processes { get; set; } = new List<CompoundProcess>();
+      public Alternative[] Lipophilicity { get; set; }
```
msevestre: Moved to array because it is the natural interface when using Tasks

src/PKSim.Core/Snapshots/Mappers/CompoundMapper.cs
```diff
@@ -32,53 +33,55 @@ public class CompoundMapper : ParameterContainerSnapshotMapperBase<ModelCompound
...
 
-      public override SnapshotCompound MapToSnapshot(ModelCompound compound)
+      public override async Task<SnapshotCompound> MapToSnapshot(ModelCompound compound)
       {
-         return SnapshotFrom(compound, snapshot =>
-         {
-            snapshot.CalculationMethods = _calculationMethodCacheMapper.MapToSnapshot(compound.CalculationMethodCache);
-
-            snapshot.Lipophilicity = mapAlternatives(compound, COMPOUND_LIPOPHILICITY);
-            snapshot.FractionUnbound = mapAlternatives(compound, COMPOUND_FRACTION_UNBOUND);
```
msevestre: Calling await from withing the action works but results in loss of stack trace when catching exceptions or results in exceptiuon being swollen. The way to make it work is to do all the await .... after the SnapshotFrom method

src/PKSim.Core/Snapshots/Mappers/ExtendedPropertyMapper.cs
```diff
@@ -25,47 +26,34 @@ public override SnapshotExtendedProperty MapToSnapshot(ModelExtendedProperty ext
...
 
-      public override ModelExtendedProperty MapToModel(SnapshotExtendedProperty snapshot)
+      public override Task<ModelExtendedProperty> MapToModel(SnapshotExtendedProperty snapshot)
       {
-
-         ModelExtendedProperty property;
-
          if (snapshot.Type == typeof(string))
-         {
-            property = addOptionsToExtendedProperty(snapshot, option => option.ToString());
```
msevestre: @rwmcintosh Small refactoring to return right away

src/PKSim.Core/Snapshots/Mappers/ParameterContainerSnapshotMapperBase.cs
```diff
@@ -46,17 +48,20 @@ protected virtual bool ParameterHasChanged(IParameter parameter)
...
+      protected Task UpdateParametersFromSnapshot(TSnapshot snapshot, IContainer container, string containerDesciptor)
       {
+         var tasks = new List<Task>();
          foreach (var snapshotParameter in snapshot.Parameters)
          {
             var modelParameter = container.Parameter(snapshotParameter.Name);
 
             if (modelParameter == null)
-               throw new SnapshotParameterNotFoundException(snapshotParameter.Name, containerDesciptor);
+               return FromException(new SnapshotParameterNotFoundException(snapshotParameter.Name, containerDesciptor));
```
msevestre: This methos (Task.FromException) is defined in ,NET 4.6. I had to create my own for now until we upgrade

src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -16,55 +17,59 @@ public ProjectMapper(IExecutionContext executionContext)
...
-            snapshot.Protocols = mapBuildingBlocksToSnapshots<Protocol>(project.All<Model.Protocol>());
-            snapshot.Populations = mapBuildingBlocksToSnapshots<Population>(project.All<Model.Population>());
-         });
+         var snapshot = await SnapshotFrom(project);
+         snapshot.Individuals = await mapBuildingBlocksToSnapshots<Individual>(project.All<Model.Individual>());
+         snapshot.Compounds = await mapBuildingBlocksToSnapshots<Compound>(project.All<Model.Compound>());
+         snapshot.Events = await mapBuildingBlocksToSnapshots<Event>(project.All<PKSimEvent>());
+         snapshot.Formulations = await mapBuildingBlocksToSnapshots<Formulation>(project.All<Model.Formulation>());
+         snapshot.Protocols = await mapBuildingBlocksToSnapshots<Protocol>(project.All<Model.Protocol>());
+         snapshot.Populations = await mapBuildingBlocksToSnapshots<Population>(project.All<Model.Population>());
```
msevestre: @rwmcintosh Observed data export is missing here.  can you add that when you do import?

src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -16,55 +17,59 @@ public ProjectMapper(IExecutionContext executionContext)
...
-         allBuildingBlocks.AddRange(mapSnapshotsToBuildingBlocks(snapshot.Protocols));
-         allBuildingBlocks.AddRange(mapSnapshotsToBuildingBlocks(snapshot.Populations));
-         return allBuildingBlocks;
+         var tasks = new List<Task<object>>();
+         tasks.AddRange(mapSnapshotsToBuildingBlocks(snapshot.Individuals));
+         tasks.AddRange(mapSnapshotsToBuildingBlocks(snapshot.Compounds));
+         tasks.AddRange(mapSnapshotsToBuildingBlocks(snapshot.Events));
+         tasks.AddRange(mapSnapshotsToBuildingBlocks(snapshot.Formulations));
+         tasks.AddRange(mapSnapshotsToBuildingBlocks(snapshot.Protocols));
+         tasks.AddRange(mapSnapshotsToBuildingBlocks(snapshot.Populations));
```
msevestre: @rwmcintosh and obviously add observed data from import

tests/PKSim.Tests/Core/DataColumnMapperSpecs.cs
```diff
@@ -8,16 +9,17 @@
 
 namespace PKSim.Core
 {
-   public abstract class concern_for_DataColumnMapper : ContextSpecification<DataColumnMapper>
```
msevestre: @rwmcintosh 
Those tests use a lot of MustHaveBeenCalled etc.. 
Can you refactor to verify that that snapshot property is set propertly instead?
eg.
_snapshot.DataInfo.ShouldBeEqualTo(_snapshotDataInfo)

instead of saying that the mapper map method should have been called. 
In general, tests gets less brittle when we do not rely too much on what calls are being made.

tests/PKSim.Tests/Extensions/AsyncContextSpecification.cs
```diff
@@ -0,0 +1,97 @@
+﻿using System;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using OSPSuite.BDDHelper.Extensions;
+using OSPSuite.Core.Domain;
+using OSPSuite.Utility.Extensions;
+
+namespace OSPSuite.BDDHelper
+{
+   /// <summary>
```
msevestre: This will be moved to BDDHelper once I am sure of the implemtation

src/PKSim.Core/Model/RandomPopulationFactory.cs
```diff
@@ -165,17 +164,17 @@ private bool tryRandomize(Individual individual, RandomPopulationSettings popula
...
+         PathCache<IDistributedParameter> allBaseDistributedParamters, 
+         Gender currentGender,
+         RandomGenerator randomGenerator)
       {
          //clone of the original origin data
          currentIndividual.OriginData = populationSettings.BaseIndividual.OriginData.Clone();
          currentIndividual.OriginData.Gender = currentGender;
 
-         //create a new seed for the individual in order to randomize correctly our values
-         currentIndividual.GenerateSeed();
```
msevestre: @Yuri05 I have removed this call here. The RandomGenerator of the Population is now used using the seed

src/PKSim.Core/Snapshots/Mappers/PopulationMapper.cs
```diff
@@ -1,89 +1,85 @@
...
-         throw new NotImplementedException();
+         if (snapshotAdvancedParameters == null)
+            return;
+
+         population.RemoveAllAdvancedParameters();
+         var parameterCache = _parameterTask.PathCacheFor(population.AllIndividualParameters());
+         var tasks = snapshotAdvancedParameters.Select(x => _advancedParameterMapper.MapToModel(x, parameterCache));
+         var advancedParameters = await Task.WhenAll(tasks);
+
+         advancedParameters.Each(x => population.AddAdvancedParameter(x, generateRandomValues: true));
```
msevestre: @Yuri05 Right now there is no way to ensure that the same randomValues are used for advanced parameters (it all depends on the order with which they are added to the collection etc..)
if we really want to also have the same values for those parameters, we need to actually save the double values.
Thoughts?

Yuri05: Think it's OK as is.

msevestre: ok cool

msevestre:
@rwmcintosh I merged the changes. You can still review the comments if you need.

PR #334, opened 2017-09-18
Fixes #281 export of simulation snapshot

### LINKED ISSUE #281: Export of simulation snapshot (json)

msevestre:


msevestre:
I am not sure quite yet if what I saved will be sufficient to restore the simulation completely. I tried to saved the minimum info that I think is required but I might have to change the export while the import is being implemented

src/PKSim.Core/Model/SimulationProperties.cs
```diff
@@ -52,27 +37,38 @@ public virtual void ClearCompoundPropertiesList()
 
       public virtual Origin Origin { get; set; }
 
+      public SimulationProperties()
```
msevestre: That was just moved up to be consistent with our usual conventions

src/PKSim.Core/Snapshots/Mappers/OutputIntervalMapper.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.Threading.Tasks;
+using SnapshotOutputInterval = PKSim.Core.Snapshots.OutputInterval;
+using ModelOutputInterval = OSPSuite.Core.Domain.OutputInterval;
+
+
+namespace PKSim.Core.Snapshots.Mappers
+{
+   public class OutputIntervalMapper : ParameterContainerSnapshotMapperBase<ModelOutputInterval, SnapshotOutputInterval>
```
msevestre: Special mapper for this one so that I can use the parameter mapping implementation from base. Nothing to do here

src/PKSim.Core/Snapshots/Mappers/ParameterContainerSnapshotMapperBase.cs
```diff
@@ -38,16 +37,10 @@ protected override Task<TSnapshot> SnapshotFrom(TModel model, Action<TSnapshot>
 
       protected virtual Task AddModelParametersToSnapshot(TModel model, TSnapshot snapshot)
       {
-         return AddParametersToSnapshot(model.AllParameters(ParameterHasChanged), snapshot);
-      }
-
-      protected virtual bool ParameterHasChanged(IParameter parameter)
```
msevestre: I have moved this method as an extension method. This will probably be merged with the ValueDiffersFromDefault when we are sure of how we will implement this method

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
```diff
@@ -0,0 +1,130 @@
...
+using OSPSuite.Core.Domain.Repositories;
+using OSPSuite.Utility.Extensions;
+using PKSim.Core.Model;
+using PKSim.Core.Repositories;
+using SnapshotSimulation = PKSim.Core.Snapshots.Simulation;
+using ModelSimulation = PKSim.Core.Model.Simulation;
+
+namespace PKSim.Core.Snapshots.Mappers
+{
+   public class SimulationMapper : ObjectBaseSnapshotMapperBase<ModelSimulation, SnapshotSimulation, PKSimProject>
```
msevestre: This is a best of a mapper. Some stuff could be put in specific mapper such as event, but also event mapping is only used in the context of a simulation and is a one liner. So in the spirit of keeping it simple, no mapper for now

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -105,6 +105,14 @@ public static bool ValueDiffersFromDefault(this IParameter parameter)
          return !ValueComparer.AreValuesEqual(parameter.Value, parameter.DefaultValue.Value, CoreConstants.DOUBLE_RELATIVE_EPSILON);
       }
 
+      //TODO Use ValueOrigin state when implemented. It should be != than PKSim default and probably mereged with method above
+      // This is just a dymmy method for now to satisfy requalification project
```
rwmcintosh: I know that's going to be removed as it's todo but dymmy => dummy if you are in here again.

PR #337, opened 2017-09-18
Fixes #336 Opening project with classification alwyas results in ques…

### LINKED ISSUE #336: Opening project with classification alwyas results in question "Do you want to save project"
even if nothing was changed
This is due to change in project conversion which does not set the flag to false upon loading anymore. 
This is new i 7.2

msevestre:
…tion Do you want to save project

### DIFF
--- a/src/PKSim.Infrastructure/Serialization/ORM/Mappers/ProjectMetaDataToProjectMapper.cs
+++ b/src/PKSim.Infrastructure/Serialization/ORM/Mappers/ProjectMetaDataToProjectMapper.cs
@@ -1,4 +1,5 @@
 using System;
+using OSPSuite.Core;
 using OSPSuite.Utility;
 using OSPSuite.Utility.Extensions;
 using OSPSuite.Utility.Visitor;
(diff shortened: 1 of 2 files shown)


src/PKSim.Infrastructure/Serialization/ORM/Mappers/ProjectMetaDataToProjectMapper.cs
```diff
@@ -70,6 +72,10 @@ public IPKSimProject MapFrom(ProjectMetaData projectMetaData)
          //Once reference to dynamic meta data was added, deserialize the project itself
          _serializationManager.Deserialize(project, projectMetaData.Content.Data);
 
+         //if the project DB Version is the same as the current project, the project did not change
+         if (projectMetaData.Version == ProjectVersions.Current)
```
msevestre: We used to always set this flag to false, thus bypassing any information from project conversion (the message was nit shown EVEN IF project conversion happened). Now the logic is different so this needs to be adapted. The mapper is the perfect place

tests/PKSim.Tests/Infrastructure/ProjectMetaDataToProjectMapperSpecs.cs
```diff
@@ -15,29 +15,62 @@ public abstract class concern_for_ProjectMetaDataToProjectMapper : ContextSpecif
...
       {
-         sut.MapFrom(new ProjectMetaData()).ShouldBeAnInstanceOf<PKSimProject>();
+         _project = sut.MapFrom(_projectMetaData);
       }
 
       [Observation]
-      public void should_thrown_an_exception_for_an_unknown_type()
+      public void should_not_have_reset_the_changed_flag()
       {
-         The.Action(() => sut.MapFrom(A<ProjectMetaData>.Ignored)).ShouldThrowAn<Exception>();
```
msevestre: That test was non sense

PR #339, opened 2017-09-18
Fixes #332 - import observed data

### LINKED ISSUE #274: Import of observed data folder structure from snapshot (json)
Structure plus reference to observed data classification

### LINKED ISSUE #332: Import of observed data snapshot (json)

rwmcintosh:


src/PKSim.Core/Snapshots/Mappers/DataInfoMapper.cs
```diff
@@ -38,9 +41,28 @@ private Task<SnapshotExtendedProperties> mapExtendedProperties(ModelExtendedProp
...
+
+         if(snapshot.ExtendedProperties != null)
+            dataInfo.ExtendedProperties.AddRange(await extendedPropertiesFrom(snapshot));
+
+         return dataInfo;
+      }
+
+      private async Task<IEnumerable<IExtendedProperty>> extendedPropertiesFrom(SnapshotDataInfo snapshot)
+      {
+         return await _extendedPropertiesMapper.MapToModel(snapshot.ExtendedProperties);
```
msevestre: if you do not use the return value, async and await adds complexity for nothing here.
Simply returns the task and remove async and await alltogether

src/PKSim.Core/Snapshots/Mappers/ObservedDataMapper.cs
```diff
@@ -45,9 +47,21 @@ private Task<SnapshotExtendedProperties> mapExtendedProperties(ModelExtendedProp
...
-         throw new System.NotImplementedException();
+         var dataRepository = new ModelDataRepository
+         {
+            Name = snapshot.Name,
+            Description = snapshot.Description
+         };
+
+         dataRepository.Add(await _dataColumnMapper.MapToModel(snapshot.BaseGrid, dataRepository));
+
+         snapshot.Columns.Each(async x => dataRepository.Add(await _dataColumnMapper.MapToModel(x, dataRepository)));
```
msevestre: this is going to cause all mapping to wait for each other. In general, never do an await in a loop. Instead use Tasks.WhenAll 
(in our case, there is no io etc.. so it will effectively be the same but why not  use the expected pattern)

src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -26,9 +27,18 @@ public ProjectMapper(IExecutionContext executionContext)
...
+         snapshot.ObservedData = await mapObservedDataToSnapshots(project.AllObservedData);
+
          return snapshot;
       }
 
+      private async Task<DataRepository[]> mapObservedDataToSnapshots(IReadOnlyCollection<OSPSuite.Core.Domain.Data.DataRepository> allObservedData)
+      {
+         var snapshotMapper = _executionContext.Resolve<ObservedDataMapper>();
+         var tasks = allObservedData.Select(datarepository => snapshotMapper.MapToSnapshot(datarepository));
+         return await Task.WhenAll(tasks);
```
msevestre: see above. This is the correct way to do it
That being said, you do not need the async and await here again

src/PKSim.Infrastructure/Services/ObservedDataTask.cs
```diff
@@ -62,6 +66,13 @@ public void ExportToPkml(DataRepository observedData)
          _observedDataPersistor.Save(observedData, file);
       }
 
+      public async Task<IReadOnlyList<DataRepository>> LoadFromSnapshot()
+      {
+         var observedData = (await _snapshotTask.LoadFromSnapshot<DataRepository>()).ToList();
+         observedData.Each(AddObservedDataToProject);
+         return observedData.ToList();
```
msevestre: why ToList here? this is already a list
What does LoadFromSnapshot returns

msevestre: actually you probably do not need the returned value here as this is typically called from a UICommand? Simply return Task no?

src/PKSim.Core/Snapshots/Mappers/ObservedDataMapper.cs
```diff
@@ -45,9 +47,21 @@ private Task<SnapshotExtendedProperties> mapExtendedProperties(ModelExtendedProp
          return _extendedPropertiesMapper.MapToSnapshot(extendedProperties);
       }
 
-      public override Task<ModelDataRepository> MapToModel(SnapshotDataRepository snapshot)
+      public override async Task<ModelDataRepository> MapToModel(SnapshotDataRepository snapshot)
       {
-         throw new System.NotImplementedException();
+         var dataRepository = new ModelDataRepository
```
msevestre: There is the MapSNapshotToModel method or sthg like that that does that for you

rwmcintosh:
I've been using the wrong issue number here. Changed to #332

src/PKSim.Infrastructure/Services/ObservedDataTask.cs
```diff
@@ -66,11 +66,10 @@ public void ExportToPkml(DataRepository observedData)
          _observedDataPersistor.Save(observedData, file);
       }
 
-      public async Task<IReadOnlyList<DataRepository>> LoadFromSnapshot()
+      public async Task LoadFromSnapshot()
       {
          var observedData = (await _snapshotTask.LoadFromSnapshot<DataRepository>()).ToList();
```
msevestre: probably does not need the TOList here either no?

PR #356, opened 2017-09-28
Fixes #274 - import observed data folder structure

### LINKED ISSUE #274: Import of observed data folder structure from snapshot (json)

rwmcintosh:


src/PKSim.Core/Snapshots/Classifiable.cs
```diff
@@ -0,0 +1,7 @@
+namespace PKSim.Core.Snapshots
+{
+   public class Classifiable : SnapshotBase
```
rwmcintosh: I ended up taking the classifiables out from the nested structure in JSON. Instead, these are serialized with the name of the subject and the path of the parent classification

src/PKSim.Core/Snapshots/Mappers/BaseClassifiableMapper.cs
```diff
@@ -0,0 +1,23 @@
...
+         return SnapshotFrom(model, snapshot =>
+         {
+            snapshot.ClassificationPath = model.Parent?.Path;
+            snapshot.Name = model.Name;
+         });
+      }
+
+      public override Task<T> MapToModel(Classifiable snapshot)
+      {
+         return Task.FromResult(new T());
```
rwmcintosh: When you deserialize the classifiable, the path and name are not really properties that can be set on the newly created model so no properties are set here.

src/PKSim.Core/Snapshots/Mappers/ClassificationSnapshotService.cs
```diff
@@ -0,0 +1,110 @@
+using System.Collections.Generic;
```
rwmcintosh: There was a lot of arranging of classifiables and classifications within the project mapper. The service is concerned with how to arrange classifications and classifiables instead.

rwmcintosh:
Let's go over it tomorrow and I'll explain the trouble I had deserializing JSON where the classifiable is a sub-node of the classification.

src/PKSim.Core/Snapshots/Mappers/ClassificationMapper.cs
```diff
@@ -8,56 +7,53 @@
...
-      {
-         var root = await mapClassificationToSnapshot(classification);
 
-         var childClassifications = childClassificationsFrom(classification, classifications);
-         root.Classifications = await mapChildClassifications(classifications, classifiables, childClassifications);
+         var childClassifications = childClassificationsFrom(classification, context.Classifications);
+         var childClassifiables = childClassifiablesFrom(classification, context.Classifiables);
 
-         root.Classifiables = childClassifiablesFrom(classification, classifiables).ToArray();
+         root.Classifications = childClassifications.Any() ? await Task.WhenAll(childClassifications.Select(x => mapTreeFrom(x, context))) : null;
```
msevestre: those objects are null per consturction. Only assign when any using a simple if
```
if(childClassifications.Any())
 root.Classifications =await ....
```

src/PKSim.Core/Snapshots/Mappers/ClassificationMapper.cs
```diff
@@ -8,56 +7,53 @@
...
+         root.Classifiables = childClassifiables.Any() ? childClassifiables.ToArray() : null;
```
msevestre: to array not required if the method returns an array (right now it returns  a list)

src/PKSim.Core/Snapshots/Mappers/ClassificationMapper.cs
```diff
@@ -8,56 +7,53 @@
...
 
          return root;
       }
 
       private IReadOnlyList<string> childClassifiablesFrom(ModelClassification classification, IReadOnlyList<IClassifiableWrapper> classifiables)
       {
-         return classifiables.Where(x => Equals(x.Parent, classification)).AllNames();
+         return childrenFrom(classification, classifiables).Select(x => x.Name).ToList();
```
msevestre: change it to ToArray and you do not need the ToARray above

src/PKSim.Core/Snapshots/Mappers/ClassificationSnapshotTask.cs
```diff
@@ -0,0 +1,88 @@
...
+         where TClassifiable : Classifiable<TSubject>, new() where TSubject : IObjectBase
+      {
+         var model = await _classificationMapper.MapToModel(snapshot, classificationType);
+         model.Parent = parent;
+         project.AddClassification(model);
+
+         snapshot.Classifiables?.Each(snapshotClassifiable =>
+         {
+            var dataRepository = subjects.FirstOrDefault(x => string.Equals(x.Name, snapshotClassifiable));
+            var classifiable = project.GetOrCreateClassifiableFor<TClassifiable, TSubject>(dataRepository);
```
msevestre: that should not be named dataRepository.

src/PKSim.Core/Snapshots/Mappers/ClassificationSnapshotTask.cs
```diff
@@ -0,0 +1,88 @@
...
+            project.RemoveClassifiable(classifiable);
+            classifiable.Parent = model;
+            project.AddClassifiable(classifiable);
```
msevestre: The add and remove seems to be required to update the parent. I am not sure why this is like that internally. If this is required, we need a commment to explain why this is done like that

msevestre:
```
 "ObservedDataClassifications": [
    {
      "Name": "A",
      "Classifications": [
        {
          "Name": "B",
          "Classifiables": [
            "Baneyx 2013, Rifampin po 600mg MD"
          ]
        },
        {
          "Name": "C"
        }
      ]
    }
  ],
````
That looks so much better. Great job

src/PKSim.Core/Snapshots/Mappers/ClassificationSnapshotTask.cs
rwmcintosh: I had seen this required at one stage before, but when I tested that today, it seems that only setting the parent is required. I removed the remove/add pairing

PR #358, opened 2017-09-28
Fixes #357 zoom in population chart broken

### LINKED ISSUE #357: Zoom in population chart broken
Probably due to refactoring for charts. This only affects 7.2

msevestre:
This was indeed due to our refactoring for chart


src/PKSim.UI/Binders/ChartDataBinder.cs
```diff
@@ -76,6 +72,10 @@ public void Bind(ChartData<TX, TY> chartData, PopulationAnalysisChart analysisCh
          {
             var dummySeries = new Series("dummy", _viewType);
             _view.Chart.Series.Add(dummySeries);
+
+            //Zoom initialization should be performed with an existing Diagram
+            if (_allowZoom)
```
msevestre: Moved this line to place with existing diagram. This initialized everything needed

PR #380, opened 2017-10-11
Fixes #287 Batch Tool: User Interface for creating projects based on snapshots

### LINKED ISSUE #287: Batch Tool: User Interface for creating projects based on snapshots
Project will be created in the folder where the json file resides. 
Either selection of one input folder that will be parse recursively
Or list of json files such as 

image

msevestre:
Update start application to use console application

First step: Move CLI stuff to own component to ensure proper separation of concerns between UI and console

Split command and options

Add single folder UI for snapshot

Implementation UI done. some tests required

Update logger to display separator

Add missing tests for runner


msevestre:
Lot's of changes here. Trying to fit CLI and UI into the same tool was starting to bring a lot of frictions so I decided to split stuff
PKSim.BatchTool is the UI for the CLI
PKSim.CLI is the actual CLI
PKSim.CLI.Core is the logic for the CLI

That way we will be able to configure for example logger etc. differently based on use case

PR #381, opened 2017-10-11
Fixes #359 - ExtendedProperties should be simplified

### LINKED ISSUE #359: DataRepository mapping to snapshot for ExtendedProperties should be simplified
```
"ExtendedProperties": [
        {
          "Name": "Source",
          "Type": "System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
          "Value": "C:\\Users\\gfrqb\\Desktop\\Digoxin\\Rifampin, Peloquin, 600mg po SD.xlsx.Peloquin",
          "DisplayName": "Source",
          "ReadOnly": false
        },
        {
          "Name": "File",
          "Type": "System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
          "Value": "Rifampin, Peloquin, 600mg po SD",
          "DisplayName": "File",
          "ReadOnly": false
        },
```

I think we could simplify it greatly by saving only the stuff that can be different

* DisplayName should be saved only if it differs from Name 
* Type should not be saved as .NET Type!! instead use the name only and map it accordingly
* ReadOnly is almost never true. Only save boolean if it's true

rwmcintosh:


src/PKSim.Core/Snapshots/Mappers/ExtendedPropertyMapper.cs
```diff
@@ -17,12 +17,10 @@ public override Task<SnapshotExtendedProperty> MapToSnapshot(ModelExtendedProper
          {
             snapshot.ListOfValues = extendedProperty.ListOfValuesAsObjects.Any() ? extendedProperty.ListOfValuesAsObjects.ToList() : null;
             snapshot.Value = extendedProperty.ValueAsObject;
-            snapshot.Type = extendedProperty.Type;
             snapshot.Description = SnapshotValueFor(extendedProperty.Description);
             snapshot.Name = extendedProperty.Name;
-            snapshot.FullName = extendedProperty.FullName;
-            snapshot.DisplayName = extendedProperty.DisplayName;
```
msevestre: What's the difference between fullName and displayName?

src/PKSim.Core/Snapshots/Mappers/ExtendedPropertyMapper.cs
```diff
@@ -17,12 +17,10 @@ public override Task<SnapshotExtendedProperty> MapToSnapshot(ModelExtendedProper
...
-            snapshot.ReadOnly = extendedProperty.ReadOnly;
+            snapshot.FullName = string.Equals(extendedProperty.FullName, extendedProperty.Name) ? null : extendedProperty.FullName;
```
msevestre: Here you can use the SnapshotValueFor which takes a second parameter( default)

src/PKSim.Core/Snapshots/Mappers/ExtendedPropertyMapper.cs
```diff
@@ -42,16 +40,14 @@ private Task<ModelExtendedProperty> mapExtendedProperty<T>(SnapshotExtendedPrope
 
       public override Task<ModelExtendedProperty> MapToModel(SnapshotExtendedProperty snapshot)
       {
-         if (snapshot.Type == typeof(string))
-            return mapExtendedProperty(snapshot, option => option.ToString());
-
-         if (snapshot.Type == typeof(double))
+         if (double.TryParse(snapshot.Value.ToString(), out double _))
```
msevestre: why don't you define a parameter in the try parse that you can use if the parse works instead of parsing again?

src/PKSim.Core/Snapshots/Mappers/ExtendedPropertyMapper.cs
rwmcintosh: DisplayName returns FullName if not null, otherwise name. 
DisplayName is readonly

src/PKSim.Core/Snapshots/Mappers/ExtendedPropertyMapper.cs
rwmcintosh: Actually, I think I can leave this with SnapshotValueFor without second parameter because FullName can be null

rwmcintosh: T must be non-nullable to use this default value version right?

msevestre: there is an overload for string

msevestre: right but it does not take default parameter :)

rwmcintosh: Yeah, I think it's ok to use the standard one for string. If FullName was set, then it should be preserved, otherwise it's blank

PR #394, opened 2017-10-13
Fixes #392 MoBi simulation should not be exported to snapshot

### LINKED ISSUE #392: MoBi simulation should not be exported to snapshot

msevestre:


tests/PKSim.Tests/Core/SimulationMapperSpecs.cs
```diff
@@ -194,6 +194,23 @@ protected override Task Context()
       }
    }
 
+   public class When_exporting_a_simulation_thatt_does_not_come_from_pksim_to_snapshot : concern_for_SimulationMapper
```
rwmcintosh: can you change the name? thatt has an extra t

PR #400, opened 2017-10-17
Fixes #297 behavior with unknown parameters

### LINKED ISSUE #297: Behavior when importing a snapshot containing unknow parameter
See discussion here
https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/296#discussion_r136382658

The current implementation throws an exception with the name of the unknown parameter.

Another option would be to show the information in some kind of a log file. I favor the exception as loading a snapshot with parameters that are unknown can only lead to issues and wrong expectation. The end user should NEVER trust a snapshot that does not conform to what PK-Sim is expected

msevestre:


msevestre:
@Yuri05 instead of throwing as soon as something is not found, the load snapshot action is now performed with the log view visible. That way any problem or error is seen right away

src/PKSim.BatchTool/ApplicationStartup.cs
```diff
@@ -34,16 +32,6 @@ public static void Start()
...
-
-         initializeLogger();
-      }
-
-      private static void initializeLogger()
-      {
-         var loggerFactory = IoC.Resolve<ILoggerFactory>();
-
-         loggerFactory
-            .AddPresenter();
```
msevestre: The log presenter is now added by default when using the PKSim.UI

src/PKSim.Core/Snapshots/Mappers/CompoundMapper.cs
```diff
@@ -158,7 +158,7 @@ private IEnumerable<IParameter> changedGroupParameters(IContainer container, str
       {
          var alternativeGroup = compound.ParameterAlternativeGroup(alternativeGroupName);
          var alteratives = await _alternativeMapper.MapToSnapshots(alternativeGroup.AllAlternatives);
-         var definedAlternatives = alteratives?.Where(x => x != null).ToArray();
+         var definedAlternatives = alteratives?.ToArray();
```
msevestre: This is not required anymore. I have changed the MapToSnapshots to NOT return any null object

msevestre: This is not required anymore. I have changed the MapToSnapshots to NOT return any null object

src/PKSim.Core/Snapshots/Mappers/CurveMapper.cs
```diff
@@ -39,12 +43,24 @@ public CurveMapper(CurveOptionsMapper curveOptionsMapper, IDimensionFactory dime
          curve.CurveOptions.UpdateFrom(curveOptions);
 
          var yData = findCurveWithPath(snapshot.Y, simulationAnalysisContext.DataRepositories);
+         if (yData == null)
+         {
+            _logger.AddToLog(PKSimConstants.Error.CouldNotFindQuantityWithPath(snapshot.Y));
```
msevestre: That's the only thing that needs to be done when something should be logged. Grab the logger and write

src/PKSim.Core/Snapshots/Mappers/ParameterMapper.cs
```diff
@@ -57,6 +75,12 @@ public override Task<SnapshotParameter> MapToSnapshot(IParameter modelParameter)
          if (snapshot.Value == null)
             return parameter;
 
+         var displayUnit = ModelValueFor(snapshot.Unit);
+         if (!parameter.Dimension.HasUnit(displayUnit))
```
msevestre: This can happen either if snapshot is edited by hand OR using some old projects where unit was sometime missing

src/PKSim.Presentation/Services/ProjectTask.cs
```diff
@@ -224,34 +222,16 @@ public void LoadProjectFromSnapshot()
...
-         var fileName = _dialogCreator.AskForFileToOpen(message, Constants.Filter.JSON_FILE_FILTER, Constants.DirectoryKey.REPORT);
-         if (string.IsNullOrEmpty(fileName))
-            return;
-
          closeProject();
 
-         _heavyWorkManager.Start(LoadSnapshotIntoWorkspace, PKSimConstants.UI.LoadingSnapshot);
-
-         void LoadSnapshotIntoWorkspace()
+         using (var presenter = _applicationController.Start<ILoadProjectFromSnapshotPresenter>())
```
msevestre: Starts the UI instead of doing everything in HeavyWorkManager

src/PKSim.UI/BootStrapping/ApplicationStartup.cs
```diff
@@ -69,6 +71,16 @@ public void InitializeForStartup()
...
+
+         configureLogger(container);
+      }
+
+      private void configureLogger(IContainer container)
+      {
+         var loggerFactory = container.Resolve<ILoggerFactory>();
+         loggerFactory
+            .AddPresenter()
+            .AddDebug();
```
msevestre: Also add the debug logger. That way we always see what's happening in the debug windown

msevestre:
Need to merge that to get going with next feature. Feel free to review when you have time

PR #401, opened 2017-10-17
Fixes #385 Parameter Export to snapshot. Exported value should be rou…

### LINKED ISSUE #385: Parameter Export to snapshot. Exported value should be rounded
Attached a simple project with a MW compound set to 450
The exported json shows

```
  {
      "Name": "Molecular weight",
      "Value": 449.99999999999994,
      "Unit": "g/mol"
    },
```

This cannot be always reproduced and I think this happen when exporting a saved project, but we should do something about it.

msevestre:
…nded


src/PKSim.Infrastructure/Serialization/Json/NullabeDoubleJsonConverter.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using Newtonsoft.Json;
+
+namespace PKSim.Infrastructure.Serialization.Json
+{
+   public class NullabeDoubleJsonConverter : JsonConverter
+   {
+      private const int DOUBLE_PRECISION = 10;
```
msevestre: here I am saying that I am rounding each value to 10 digits before serializing it. That should be more than enough to avoid any side effects

tests/PKSim.Tests/Infrastructure/SnapshotSerializerSpecs.cs
```diff
@@ -0,0 +1,106 @@
...
+         _deserialiedParameter.Value.ShouldBeEqualTo(_parameter.Value);
+      }
+   }
+
+   public class When_serializing_a_parameter_whose_value_was_tainted_with_numerical_noise : concern_for_SnapshotSerializer
+   {
+      protected override async Task Context()
+      {
+         await base.Context();
+         _parameter.Value = 449.9999999999991;
```
msevestre: This is the kind of stuff that this PR aims at solving

msevestre:
@Yuri05 actually i think there is probably a better way to do it.
Do you remember we had at some point a method that was converting a double to a string and back to double for that very reason. Any idea where the code is?

msevestre:
@Yuri05 Could you review that today?

tests/PKSim.Tests/Infrastructure/SnapshotSerializerSpecs.cs
```diff
@@ -0,0 +1,136 @@
...
```
Yuri05: would make it e.g. **449.99999999996;**
(10 times 9 after decimal point)

tests/PKSim.Tests/Infrastructure/SnapshotSerializerSpecs.cs
Yuri05: >Can you think of any more tests that I should write?

Maybe smthg in scientific notation? E.g. 199.999999998E-2

msevestre:
test updated :)

tests/PKSim.Tests/Infrastructure/SnapshotSerializerSpecs.cs
```diff
@@ -0,0 +1,79 @@
...
+         await base.GlobalCleanup();
+         FileHelper.DeleteFile(_fileName);
+      }
+   }
+
+  
+   public class When_serializing_a_parameter_with_a_value_set: concern_for_SnapshotSerializer
+   {
+      [TestCase(null)]
+      [TestCase(1.123456789)]
```
msevestre: Much easier to understand what's happening instead of trying to find names that make sense :)

msevestre:
Merging now that I have added more tests.

src/PKSim.Infrastructure/Serialization/Json/NullabeDoubleJsonConverter.cs
```diff
@@ -1,20 +1,30 @@
...
+
       public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
       {
          var d = (double?) value;
          if (!d.HasValue)
             return;
 
-         var rounded = Math.Round(d.Value, DOUBLE_PRECISION);
-         writer.WriteValue(rounded);
+         var formatted = _doubleFormatter.Format(d.Value);
```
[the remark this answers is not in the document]
Yuri05: Hmm, will we could get trouble e.g. with solver tolerances.
AbsTol <1e-10 is not unusual. I think we will need special handling for that.

e.g. 
`if fabs(d.Value)<10^(-DOUBLE_PRECISION) write/read unformatted value`

Issue #368, opened 2017-10-04
Integrate pregnant population


Yuri05:
Integrate new pregnant population based on the [latest publications](https://www.ncbi.nlm.nih.gov/pubmed/?term=dallmann%2C+ince)

* Population name: _Pregnant (Dallmann et al. 2017)_
* Age range: [30 .. 30.75] years

PR #422, opened 2017-11-01
Yuri05 #368 integrate pregnant population

msevestre:
@Yuri05 This is your PR but I fixed a few issues that were breaking tests


msevestre:
Ok I am going to create  a branch for that and we can both push on it if required instead of this issues with PR. WE should merge this branch when the work is done

PR #432, opened 2017-11-06
Fixes #420 project conversion for pregancy

### LINKED ISSUE #420: Project conversion
Add new parameters to Individual:

* Weight   of blood organs
* Weight   of tissue organs

msevestre:


msevestre:
@Yuri05 I do not think that we need to add those parameters are described in the  issue. Instead the calculation method is missing and needs to be added to the origin data

PR #438, opened 2017-11-08
Fixes #435 project corrupteed

### LINKED ISSUE #435: Project can potentially become corrupted after loading a compound from template
To reproduce with 7.2
* Create a project
* Add compound to user database
* Load compound from database. Give it another name
* Save project
* load project

msevestre:


src/PKSim.Core/Commands/AddBuildingBlockToProjectCommand.cs
```diff
@@ -23,6 +23,7 @@ protected override void ExecuteWith(IExecutionContext context)
       {
          var project = context.CurrentProject;
          project.AddBuildingBlock(_buildingBlockToAdd);
+         _buildingBlockToAdd.HasChanged = true;
```
msevestre: This is now required since the flag is not set anymore automatically upon loading (problem with multiple conversion message). 
By setting the flag in the AddBuildingBlockToProjectCommand we should be good to go

PR #443, opened 2017-11-09
Fixes #368 Integrate pregnant population

### LINKED ISSUE #368: Integrate pregnant population

Yuri05:
Once this PR is merged, pregnancy branch can be merged back into develop

### DIFF
--- a/tests/PKSim.Tests/IntegrationTests/SimulationSpecs.cs
+++ b/tests/PKSim.Tests/IntegrationTests/SimulationSpecs.cs
@@ -19,6 +19,7 @@ using OSPSuite.Core.Domain.Formulas;
 using OSPSuite.Core.Domain.Services;
 using IContainer = OSPSuite.Core.Domain.IContainer;
 using IFormulaFactory = PKSim.Core.Model.IFormulaFactory;
+using System;
 
 namespace PKSim.IntegrationTests
 {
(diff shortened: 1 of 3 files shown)


Yuri05:
No code changes here, just all missing data/formulas for pregnancy population added.

Integration tests extended so that all populations and all population simulations are created once.

tests/PKSim.Tests/IntegrationTests/SimulationSpecs.cs
```diff
@@ -159,6 +166,54 @@ public void should_be_able_to_simulate_the_simulation()
       }
    }
 
+   public class When_creating_a_population_simulation_with_the_standard_building_block_and_iv_bolus_Asian_Tanaka_1996 : When_creating_a_population_simulation_with_the_standard_building_block_and_iv_bolus
```
msevestre: hum. Why not just iterate over all population instead of creating a test for each population explicitely? SImply use the PopulationRepositgory no?

Yuri05: Just because I wanted to have an explicit message for every pop which fails.
Well could also iterate over the pops and collect error messages

msevestre: >Well could also iterate over the pops and collect error messages
Yes that would be the same. The advantage is that it would work for any other population automatically without having to write code.

If you want to leave it like that, this is also ok.

Yuri05:
OK I have replaced single tests with iteration over pops. Indeed is better

PR #450, opened 2017-11-13
Fixes #320 merge batch to snapshot

### LINKED ISSUE #320: Batch functionalities to merge with Snapshot
There is a lot of duplication and same behavior. Probably a good idea would be to keep the batch structure but use functionalities from the snapshot component instead

msevestre:


src/PKSim.CLI.Core/Batch/Compound.cs
```diff
@@ -2,13 +2,13 @@
 
 namespace PKSim.Core.Batch
 {
-   public class PkaType
```
msevestre: I have marked all those classes as internal for now. There is another entry to remove them once we are ready to convert all batch to snapshot (this cannot be done for now because we have no way to only export parameters that have actually changed. This will be possible once Model Documentation is done and we have ValueOrigin parameters)

src/PKSim.CLI.Core/Batch/Mapper/OriginDataMapper.cs
```diff
@@ -88,13 +88,13 @@ public Model.OriginData MapFrom(BatchOriginData batchOriginData)
 
          var meanWeightParameter = _individualModelTask.MeanWeightFor(originData);
          originData.Weight = valueFrom(batchOriginData.Weight, meanWeightParameter.Value).Value;
-         originData.WeightUnit = meanWeightParameter.Dimension.BaseUnit.Name;
+         originData.WeightUnit = meanWeightParameter.Dimension.DefaultUnit.Name;
```
msevestre: This are all display units (what is selected by the user)

src/PKSim.CLI.Core/Services/BatchConverterRunner.cs
```diff
@@ -0,0 +1,142 @@
...
+using PKSim.Core.Commands;
+using PKSim.Core.Extensions;
+using PKSim.Core.Model;
+using PKSim.Core.Services;
+using PKSim.Core.Snapshots.Services;
+using Simulation = PKSim.Core.Batch.Simulation;
+
+namespace PKSim.CLI.Core.Services
+{
+   internal class BatchConverterRunner : IBatchRunner<BatchConverterRunOptions>
```
msevestre: This is the CLI command to convert batch to snapshot. This can be deleted once we are happy with the conversion

src/PKSim.Core/Snapshots/OriginData.cs
```diff
@@ -0,0 +1,81 @@
...
+
+      /// <summary>
+      ///    Age of individual to create in [year]. This is a mandatory input for age dependent species. If not defined, set to NaN
+      /// </summary>
+      public Parameter Age { get; set; }
+
+      /// <summary>
+      ///    Gestational age of individual to create in [weeks]. If not defined, set to 40
+      /// </summary>
+      public Parameter GestationalAge { get; set; }
```
msevestre: is is possible to use such a class in Matlab? Or does it have to be only primitive type (double, string etc..)

PR #499, opened 2017-12-05
498 matlab tests

msevestre:


msevestre:
This PR moves all matlab related tests into a separate assemblies. We were not able to run them on TC before. It kind of work by coincidence on Appveyor (order of tests or sthg) but started failing again after renaming some files. Not acceptable.

PR #525, opened 2018-01-12
512 creating building block should set value origin

msevestre:


msevestre:
@Yuri05 First leg of implementation as discussed. IS_INPUT is not used yes anywhere but this is the next step.

PR #535, opened 2018-01-16
Fixes #532 Propagation of fields for value description and value origin mode

### LINKED ISSUE #532: Propagation of fields for "value description" and "value origin mode"
see https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/228

msevestre:
Update smartxls to latest version

Ensure the the Default value is set based on database value

Add some more test to ensure that values are set propertly when coming from database

Ensure that value origin is reset when resetting a default parameter

Fixes #532 Propagation of fields for value description and value origin mode


msevestre:
@Yuri05 This is the code that ensures that ValueOrigin are being read from the database and propagated to the building blocks. Also the update commit logic was updated to ensure that ValueOrigin is updated after update commit.

src/PKSim.Core/Commands/BuildingBlockChangeCommand.cs
```diff
@@ -78,7 +78,7 @@ protected BuildingBlockChangeCommand(T buildingBlock)
       protected override void PerformExecuteWith(IExecutionContext context)
       {
          ObjectType = context.TypeFor(_buildingBlock);
-         context.UpdateBuildinBlockProperties(this, _buildingBlock);
+         context.UpdateBuildinBlockPropertiesInCommand(this, _buildingBlock);
```
msevestre: no change in this code. Just a rename of a method

src/PKSim.Core/Commands/EditParameterCommand.cs
```diff
@@ -51,36 +54,40 @@ protected void UpdateDependenciesOnParameter(IExecutionContext context)
...
       {
-         if (!IsBuildingBlockParameterInSimulation(parameter))
+         if (!isBuildingBlockParameterInSimulation(parameter))
             return null;
 
          return context.Get<IParameter>(parameter.Origin.ParameterId);
       }
 
-      protected bool IsBuildingBlockParameterInSimulation(IParameter parameter)
+      private bool isBuildingBlockParameterInSimulation(IParameter parameter)
```
msevestre: Looking at this logic, I simply reorganized the if statements and updated the comments that were ...well outdated as always :)

src/PKSim.Core/Commands/EditParameterCommand.cs
```diff
@@ -104,6 +111,64 @@ private void setBuildingBlockAlteredFlag(IExecutionContext context)
...
+      {
+         _oldValueOrigin = parameter.ValueOrigin.Clone();
+      }
+
+      protected virtual void UpdateParameter(IExecutionContext context)
+      {
+         var bbParameter = OriginParameterFor(_parameter, context);
+         UpdateParameter(_parameter, context);
+         UpdateParameter(bbParameter, context);
+         UpdateValueOriginInParameters(_parameter, bbParameter);
```
msevestre: I am always updating teh VAlueOrigin in both parmaeter in corresponding bbParameter in Simulation

src/PKSim.Core/Commands/EditParameterCommand.cs
```diff
@@ -104,6 +111,64 @@ private void setBuildingBlockAlteredFlag(IExecutionContext context)
...
+         if (!parameter.ValueOrigin.Default)
+            return;
+
+         SaveValueOriginFor(parameter);
+         markDefaultValueOriginAsChanged(parameter.ValueOrigin);
+      }
+
+      private static bool valueOriginShouldBeUpdatedAutomatically(ValueOrigin valueOrigin)
+      {
+         return valueOrigin.Source == ValueOriginSources.Undefined &&
```
msevestre: To be discussed. This will only be updated if both source and method are undefined. Otherwise, the user changed something already so this should not be resetted

src/PKSim.Core/Commands/PKSimCommand.cs
```diff
@@ -42,9 +42,9 @@ public PKSimMacroCommand(IEnumerable<ICommand> subCommands) : this()
          subCommands.Each(Add);
       }
 
-      public new IEnumerable<IPKSimCommand> All()
+      public new IEnumerable<IOSPSuiteCommand> All()
```
msevestre: No need to be so specific here

src/PKSim.Core/Commands/ResetParameterCommand.cs
```diff
@@ -22,7 +24,25 @@ protected override void ExecuteUpdateParameter(IExecutionContext context)
...
-         parameter?.ResetToDefault();
+         if (parameter == null) return;
+         parameter.ResetToDefault();
+         SaveValueOriginFor(parameter);
+         resetValueOriginForDefaultParameter(parameter, context);
+      }
+
+      private void resetValueOriginForDefaultParameter(IParameter parameter, IExecutionContext context)
+      {
+         var parametersInContainerRepository = context.Resolve<IParametersInContainerRepository>();
```
msevestre: In case of a reset command,  I did implement the concept of retrieving the default value origin if available. The update is only performed for default parameter

src/PKSim.Core/Commands/SetRelativeExpressionFromNormalizedCommand.cs
```diff
@@ -33,6 +33,9 @@ public override void Execute(IExecutionContext context)
          Add(new SetParameterValueCommand(_normalizedParameter, _newNormalizedValue));
 
          base.Execute(context);
+
+         //clear references
+         _normalizedParameter = null;
```
msevestre: This was a memory leak here

src/PKSim.Core/Commands/UpdateParameterValueOriginCommand.cs
```diff
@@ -0,0 +1,47 @@
+﻿using OSPSuite.Core.Commands;
+using OSPSuite.Core.Domain;
+using OSPSuite.Utility.Extensions;
+using PKSim.Core.Model;
+
+namespace PKSim.Core.Commands
+{
+   public class UpdateParameterValueOriginCommand : PKSimMacroCommand
```
msevestre: I have encapsulated the action in a command for now. That means that updating the value origin will result in an entry in the history

src/PKSim.Core/Extensions/ValueOriginExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using OSPSuite.Core.Domain;
+
+namespace PKSim.Core.Extensions
+{
+   public static class ValueOriginExtensions
```
msevestre: This will be moved to core soon

src/PKSim.Core/Services/ParameterSetUpdater.cs
```diff
@@ -143,11 +150,16 @@ public ICommand UpdateValuesByName(IEnumerable<IParameter> sourceParameters, IEn
...
+      public ICommand UpdateValue(IParameter sourceParameter, IParameter targetParameter, bool updateParameterOriginId=true)
       {
-         _parameterIdUpdater.UpdateParameterId(sourceParameter, targetParameter);
+         if(updateParameterOriginId)
+            _parameterIdUpdater.UpdateParameterId(sourceParameter, targetParameter);
+
+         var command = _parameterUpdater.UpdateValue(sourceParameter, targetParameter);
+
+         //this line ensure that value origin is updated for all parameters, even the one for which the value were not changed
          targetParameter.ValueOrigin.UpdateFrom(sourceParameter.ValueOrigin);
```
msevestre: With this line, even PARAMETER whose value have not changed will get the updated value origin. I think this is the best way to implement the feature. 
Drawback. In case of update/commit and undo, some ValueOrigin that may have changed without value changed might be outdated. I don't think this happens really often. On the other hand, when a ValueOrigin is updated in say a simulation, it is important to have a way to update it in the building block without having to change the value of the parameter.

src/PKSim.Core/Services/ParameterTask.cs
```diff
@@ -434,13 +434,7 @@ public ICommand SetParameterValue(IParameter parameter, double value, ISimulatio
...
       public ICommand SetParameterValueOrigin(IParameter parameter, ValueOrigin newValueOrigin)
       {
-         //TODO MBD
-         //no command required here
-         parameter.ValueOrigin.UpdateFrom(newValueOrigin);
-         var originParameter = _executionContext.Get<IParameter>(parameter.Origin.ParameterId);
-         originParameter?.ValueOrigin.UpdateFrom(newValueOrigin);
-
-         return new PKSimEmptyCommand();
+         return new UpdateParameterValueOriginCommand(parameter, newValueOrigin).Run(_executionContext);
```
msevestre: Command after all for now

src/PKSim.Core/Services/ParameterUpdater.cs
```diff
@@ -73,18 +74,40 @@ public ICommand UpdateValue(IParameter sourceParameter, IParameter targetParamet
...
+      }
+
+      private ICommand updateDistributedTableFormula(IParameter targetParameter, IDistributedParameter sourceDistributedParameter)
+      {
+         return withUpdatedValueOrigin(_parameterTask.UpdateDistributedTableFormula(targetParameter, sourceDistributedParameter), sourceDistributedParameter, targetParameter);
+      }
+
+      private ICommand withUpdatedValueOrigin(ICommand command, IParameter sourceParameter, IParameter targetParameter)
+      {
+         if (sourceParameter.ValueOrigin.IsIdenticalTo(targetParameter.ValueOrigin))
```
msevestre: I am only creating the command if the values are indeed different. I don't want to have the history explode with unnecessary commands. Alternatively, this could be done in the SetParameterVAlueOriginTask (but this would go against what we do otherwise)

src/PKSim.Core/Services/SimulationParametersToBuildingBlockUpdater.cs
```diff
@@ -59,17 +59,16 @@ public ICommand UpdateParametersFromSimulationInBuildingBlock(Simulation simulat
...
       }
 
-
-
-      private IPKSimCommand updateParameterValues(PathCache<IParameter> sourceParameters, PathCache<IParameter> targetParameters)
+      private IOSPSuiteCommand updateParameterValues(PathCache<IParameter> simulationParameters, PathCache<IParameter> templateParameters)
       {
-         return _parameterSetUpdater.UpdateValues(sourceParameters, targetParameters).DowncastTo<IPKSimCommand>();
+         //we do not want to update parameter origin id here since we are updateing building block parameters
+         return _parameterSetUpdater.UpdateValues(simulationParameters, templateParameters, updateParameterOriginId: false);
```
msevestre: This was  bug in current PK-Sim implementation. Without this flag (new flag) the OriginParameterId of a template building block was set to the ParameterId of teh parameter in the simulation after Commit. This could have lead potentially to wrong references being updated

src/PKSim.Core/Commands/EditParameterCommand.cs
```diff
@@ -51,36 +54,40 @@ protected void UpdateDependenciesOnParameter(IExecutionContext context)
...
       }
 
-      protected bool IsBuildingBlockParameterInSimulation(IParameter parameter)
+      private bool isBuildingBlockParameterInSimulation(IParameter parameter)
       {
-         //a parameter defined in a simulation? Not a building block parameter
-         if (parameter.BuildingBlockType == PKSimBuildingBlockType.Simulation) return false;
+         //Building block id is not set => This is the case of parameter defined in a building block
+         if (string.IsNullOrEmpty(parameter.Origin.BuilingBlockId))
+            return false;
```
Yuri05: Typo: Buil**d**ingBlockId

msevestre: Well it's a property from the object. I'd rather not have to do project conversion to be honest

msevestre: been like that since forever :)

PR #536, opened 2018-01-17
Fixes #532 changes in propagation of set value origin command

### LINKED ISSUE #532: Propagation of fields for "value description" and "value origin mode"

msevestre:


msevestre:
This is what we discussed this morning

src/PKSim.Core/Commands/EditParameterCommand.cs
```diff
@@ -73,21 +73,21 @@ private bool isBuildingBlockParameterInSimulation(IParameter parameter)
...
-         //The simulation id is not set. The parameter is defined in a building block
-         if (string.IsNullOrEmpty(parameter.Origin.SimulationId))
-            return false;
-
-         //a parameter defined in a simulation? Not a building block parameter in a  simulation
-         if (parameter.BuildingBlockType == PKSimBuildingBlockType.Simulation)
+         //THIS SHOULD NOT HAPPEN and should be deleted
+         if (string.IsNullOrEmpty(parameter.Origin.SimulationId) || parameter.BuildingBlockType == PKSimBuildingBlockType.Simulation)
+         {
+            Debug.Print($"This should not happen for parameter {parameter.Name}");
```
msevestre: If we don't see it happen, we should simply remove this check in the future. Keep for now

src/PKSim.Core/Commands/ResetParameterCommand.cs
```diff
@@ -17,21 +16,23 @@ public ResetParameterCommand(IParameter parameter) : base(parameter)
...
       {
          if (parameter == null) return;
          parameter.ResetToDefault();
-         SaveValueOriginFor(parameter);
          resetValueOriginForDefaultParameter(parameter, context);
       }
 
       private void resetValueOriginForDefaultParameter(IParameter parameter, IExecutionContext context)
       {
+         //TODO MBD: That does not work quite yet
```
msevestre: Das funktioniert nicht...

src/PKSim.Core/Commands/UpdateParameterValueOriginCommand.cs
```diff
@@ -1,46 +1,39 @@
...
+﻿using OSPSuite.Core.Commands.Core;
 using OSPSuite.Core.Domain;
-using OSPSuite.Utility.Extensions;
-using PKSim.Core.Model;
+using Command = OSPSuite.Assets.Command;
 
 namespace PKSim.Core.Commands
 {
-   public class UpdateParameterValueOriginCommand : PKSimMacroCommand
+   public class UpdateParameterValueOriginCommand : EditParameterCommand
```
msevestre: Need to write my own command and cannot use the one in core. Just too much stuff to do to syncrhonize building block otherwise. SImply not worth it

src/PKSim.Core/Services/ParameterSetUpdater.cs
```diff
@@ -150,16 +153,27 @@ public ICommand UpdateValuesByName(IEnumerable<IParameter> sourceParameters, IEn
...
+         return withUpdatedValueOrigin(_parameterUpdater.UpdateValue(sourceParameter, targetParameter), sourceParameter, targetParameter);
+      }
+
+      private ICommand withUpdatedValueOrigin(ICommand command, IParameter sourceParameter, IParameter targetParameter)
+      {
+         if (sourceParameter.ValueOrigin.IsIdenticalTo(targetParameter.ValueOrigin))
+            return command;
+
+         var updateValueOriginCommand = _parameterTask.SetParameterValueOrigin(targetParameter, sourceParameter.ValueOrigin);
+         if (command == null)
```
msevestre: command may be null if ONLY value origin has changed. In that case return the valueorigincommand

src/PKSim.Infrastructure/ORM/Repositories/FlatParameterInContainsRepository.cs
```diff
@@ -0,0 +1,19 @@
...
+using PKSim.Infrastructure.ORM.Core;
+using PKSim.Infrastructure.ORM.Mappers;
+
+namespace PKSim.Infrastructure.ORM.Repositories
+{
+   public interface IFlatParameterInContainsRepository : IMetaDataRepository<ParameterMetaData>
+   {
+   }
+
+   public class FlatParameterInContainsRepository : MetaDataRepository<ParameterMetaData>, IFlatParameterInContainsRepository
```
msevestre: Jsut moved in its own file

src/PKSim.Infrastructure/ORM/Repositories/ParametersInContainerRepository.cs
```diff
@@ -43,19 +30,23 @@ public ParameterMetaData ParameterMetaDataFor(IParameter parameter)
          if (parameter == null)
             return null;
 
-         return ParameterMetaDataFor(_entityPathResolver.PathFor(parameter));
-      }
+         if (parameter.IsDistributed())
```
msevestre: I thought I was clever with this implementation but this does not work. We need to know CM or PVV depending on the parameter type... problem is that we only have the parameter in the command

src/PKSim.Core/Repositories/IParametersInContainerRepository.cs
```diff
@@ -1,12 +1,10 @@
 ﻿using OSPSuite.Core.Domain;
-using OSPSuite.Utility.Collections;
 using PKSim.Core.Model;
 
 namespace PKSim.Core.Repositories
 {
-   public interface IParametersInContainerRepository : IStartableRepository<ParameterMetaData>
+   public interface IParametersInContainerRepository
    {
```
Yuri05: why it should not be startable anymore?

src/PKSim.Infrastructure/ORM/Repositories/ParametersInContainerRepository.cs
```diff
@@ -43,19 +30,23 @@ public ParameterMetaData ParameterMetaDataFor(IParameter parameter)
...
+            return parameterMetaDataFrom(_parameterDistributionRepository, parameter);
 
-      protected override void DoStart()
-      {
-         base.DoStart();
-         _parameterMetaDataList.Each(x => _parameterMetaDataCache.Add($"{x.ParentContainerPath}{ObjectPath.PATH_DELIMITER}{x.ParameterName}", x));
+         if (!parameter.Formula.IsConstant())
+            return parameterMetaDataFrom(_parameterRateRepository, parameter);
```
Yuri05: will it also work for parameters, which are defined as rates but have a constant formula (e.g. "0" or "1")?

src/PKSim.Core/Repositories/IParametersInContainerRepository.cs
msevestre: As mentioned in previous comment, this whole repository does not work. It is actually more a service than a repository because it needs to retrieve, for a live parameter `parameter` the corresponding `ParameterMetaDatra`. 


The discussed implementation by path simply cannot work.

src/PKSim.Infrastructure/ORM/Repositories/ParametersInContainerRepository.cs
msevestre: See above. This strategy does not work. If we want to be able to retrieve ParmaeterMEtaData from a PArameter, we will have to think about something else. This function is new and at the moment only used by the ResetParameterCommand.

src/PKSim.Core/Repositories/IParametersInContainerRepository.cs
msevestre: I think that an easy way to achieve what we want is to save the ValueOriginId from the database (integer, costs nothing) in the ValueOrigin. This will not be used in MoBi but will allow PK-Sim to grab the original value origin from the database at any time independent from parameter path, formulat type etc..
At any rate this has nothing to do with this PR.

PR #546, opened 2018-01-26
Fixes #543 Some parameters should be set to IsDefault=true in Core BuildingBlock when simulation is created

### LINKED ISSUE #543: Some parameters should be set to IsDefault=true in Core BuildingBlock when simulation is created
## EventBuildingBlock
StartTime  (application and event)
Dose

## Spatial Structure
Height
Age
GA

msevestre:


tests/PKSim.Tests/Core/DomainHelperForSpecs.cs
```diff
@@ -120,12 +120,13 @@ public static Individual CreateIndividual()
          return individual;
       }
 
-      public static IParameter ConstantParameterWithValue(double value)
+      public static IParameter ConstantParameterWithValue(double value=10, bool isDefault = false)
```
msevestre: Enhance the helper method to define a default value and a default state so that I do not have to repeat it all the time :)

src/PKSim.Core/Services/ParameterDefaultStateUpdater.cs
```diff
@@ -14,22 +15,44 @@ public class ParameterDefaultStateUpdater : IParameterDefaultStateUpdater
    {
       public void UpdateDefaultFor(ISpatialStructure spatialStructure)
```
msevestre: Super simple...
is it too simple?

src/PKSim.Core/Services/ParameterDefaultStateUpdater.cs
```diff
@@ -14,22 +15,44 @@ public class ParameterDefaultStateUpdater : IParameterDefaultStateUpdater
...
-      private void updateDefaultStateForOriginDataParameters(IContainer spatialStructureOrganism)
+      private void updateDefaultStateForOriginDataParameters(IContainer organism)
       {
-         new[] { CoreConstants.Parameter.AGE, Constants.Parameters.GESTATIONAL_AGE, CoreConstants.Parameter.HEIGHT }.Each(x =>
+         var parameterNames = new[]
          {
-            spatialStructureOrganism.Parameter(x).IsDefault = true;
-         });
-      }
+            CoreConstants.Parameter.AGE,
```
msevestre: The only 3 constant parameters that I could think of in Individual. We might need to  add more as we test it

src/PKSim.Core/Services/ParameterDefaultStateUpdater.cs
```diff
@@ -14,22 +15,44 @@ public class ParameterDefaultStateUpdater : IParameterDefaultStateUpdater
...
       {
-         //TODO implement
+         eventGroupBuildingBlock.Each(updateDefaultStateForEventParameters);
       }
+
+      private void updateDefaultStateForEventParameters(IEventGroupBuilder eventGroupBuilder)
+      {
+         var parameterNames = new[]
+         {
+            CoreConstants.Parameter.DOSE,
```
msevestre: These are all the event parameters that I could think of as well .. We might need more also

PR #551, opened 2018-01-30
Fixes #550 Percentile Parameter should have the flag IsDefault = fals…

### LINKED ISSUE #550: Percentile Parameter should have the flag IsDefault = false when exporting from PKSim

msevestre:
…e when exporting from PKSim

### DIFF
--- a/src/PKSim.Core/Services/ParameterDefaultStateUpdater.cs
+++ b/src/PKSim.Core/Services/ParameterDefaultStateUpdater.cs
@@ -22,17 +22,29 @@ namespace PKSim.Core.Services
 
       private void updateDefaultStateForOriginDataParameters(IContainer organism)
       {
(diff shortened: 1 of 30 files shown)


src/PKSim.Core/Services/ParameterDefaultStateUpdater.cs
```diff
@@ -22,17 +22,29 @@ public void UpdateDefaultFor(ISpatialStructure spatialStructure)
...
             CoreConstants.Parameters.HEIGHT
          };
 
-         parameterNames
+         organismParameters
             .Select(organism.Parameter)
             .Where(p => p != null)
             .Each(resetDefault);
+
+         var distributionParameters = new[]
```
msevestre: Those guys need to be reset as well since we are not scaling them

PR #557, opened 2018-02-02
Fixes #556 Use IsDefault to decide if a parameter should be updated t…

### LINKED ISSUE #556: Use IsDefault to decide if a parameter should be updated to snapshot
and also Update OSPSuite.Core to remove redundant classes

msevestre:
…o snapshot


src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -105,13 +105,7 @@ public static bool ValueDiffersFromDefault(this IParameter parameter)
...
       }
 
-      //TODO Use ValueOrigin state when implemented. It should be != than PKSim default and probably mereged with method above
-      // This is just a dymmy method for now to satisfy requalification project
-      public static bool ParameterHasChanged(this IParameter parameter)
-      {
-         var canBeEdited = parameter.Visible && parameter.Formula.IsConstant();
-         return parameter.ValueDiffersFromDefault() || canBeEdited && parameter.Value != 0;
-      }
+      public static bool ParameterHasChanged(this IParameter parameter) => parameter.Visible && parameter.IsDefault == false;
```
msevestre: This is really the one difference. Only take parameter that are visible at the moment. Not sure if this is required or not

msevestre:
@Yuri05: Merging.... almost everything is related to stuff moved from PKSim into Core. Otherwise one real line

src/PKSim.Core/Model/ParameterExtensions.cs
Yuri05: I would remove `parameter.Visible` check.
E.g. parameters hidden in PK-Sim could be changed in MoBi; after reimport of a simulation back in PK-Sim those changes should appear under changed parameters

PR #563, opened 2018-02-09
Fixes #486 Hidden parameter with a value changed by the user (IsFixed…

### LINKED ISSUE #486: Hidden parameter with a value changed by the user (IsFixedValue = true) should be shown
This can happen when a hidden parameter is edited in MoBi and the simulation is imported back into PKSim. 
There is no way at the moment to see this change in PK-Sim which can lead to serious frustration when trying to understand why something is not working as expected

msevestre:
…Value = true) should be shown

### DIFF
--- a/src/PKSim.Presentation/Presenters/Parameters/ParameterSetPresenter.cs
+++ b/src/PKSim.Presentation/Presenters/Parameters/ParameterSetPresenter.cs
@@ -1,17 +1,17 @@
 using System.Collections.Generic;
 using System.Linq;
-using OSPSuite.Utility.Events;
-using OSPSuite.Utility.Extensions;
-using PKSim.Core.Model;
-using PKSim.Core.Services;
-using PKSim.Presentation.DTO.Parameters;
-using PKSim.Presentation.Services;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Formulas;
 using OSPSuite.Core.Events;
 using OSPSuite.Presentation.DTO;
 using OSPSuite.Presentation.Views;
+using OSPSuite.Utility.Events;
+using OSPSuite.Utility.Extensions;
 using PKSim.Core;
+using PKSim.Core.Model;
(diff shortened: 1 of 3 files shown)


src/PKSim.Presentation/Presenters/Parameters/ParameterSetPresenter.cs
```diff
@@ -149,10 +150,17 @@ public bool IsFormulaNotFixed(IParameterDTO parameterDTO)
 
       public virtual void Edit(IEnumerable<IParameter> parameters)
       {
-         _visibleParameters = parameters.Where(x => x.Visible).ToList();
+         _visibleParameters = parameters.Where(shouldShowParameter).ToList();
          _pathCache = _parameterTask.PathCacheFor(_visibleParameters);
       }
 
+      private bool shouldShowParameter(IParameter parameter)
```
msevestre: I have just created a function and extended the logic

tests/PKSim.Tests/Presentation/MultiParameterEditPresenterSpecs.cs
```diff
@@ -250,6 +248,59 @@ public void should_hide_the_distribution_if_unvisible()
...
+      private IParameter _hiddenIsFixedParameter;
+      private IParameter _hiddenIsNotDefaultParameter;
+      private List<ParameterDTO> _allParameterDTO;
+      private List<IParameter> _displayedParameters;
+
+      protected override void Context()
+      {
+         base.Context();
+         _visibleParameter = new PKSimParameter {Visible = true};
+         _hiddenUnchangedParameter = new PKSimParameter {Visible = false, IsDefault = true};
```
msevestre: Our different types of parameters here
Visible
Hidden unchanged
Hidden IsFixedValue = true (Formula was changed to constant....this should result in IsDefault false I believe)
Hidden IsDefault = false

msevestre:
@Yuri05 Going to merge that... Please review

PR #566, opened 2018-02-13
Fixes #564 order of favorites editable

### LINKED ISSUE #564: The order of Favorites should be editable
see request here https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/237

msevestre:


src/PKSim.Presentation/Presenters/Parameters/FavoriteParametersPresenter.cs
```diff
@@ -17,27 +17,49 @@ public interface IFavoriteParametersPresenter : ICustomParametersPresenter,
       IListener<FavoritesLoadedEvent>
 
    {
+      void MoveUp();
+      void MoveDown();
    }
 
-   public class FavoriteParametersPresenter : MultiParameterEditPresenter, IFavoriteParametersPresenter
+   public class FavoriteParametersPresenter : AbstractCommandCollectorPresenter<IFavoriteParametersView, IFavoriteParametersPresenter>, IFavoriteParametersPresenter
```
msevestre: Composition over inheritance. Should really be the goal anyways

src/PKSim.Presentation/Presenters/Parameters/FavoriteParametersPresenter.cs
```diff
@@ -47,25 +69,39 @@ where _allParameterCache.Contains(favorite)
...
+
+      public void Handle(FavoritesLoadedEvent eventToHandle) => updateParameters();
+
+      public void MoveUp() => moveFavorites(x => x.MoveUp);
+
+      public void MoveDown() => moveFavorites(x => x.MoveDown);
+
+      private void moveFavorites(Func<Favorites, Action<IEnumerable<string>>> moveActionFunc)
       {
+         var selectedParameters = _multiParameterEditPresenter.SelectedParameters;
```
msevestre: When parameter are being moved, we get them from the view. Then move them in the favorite list and select them again

msevestre: There are probably more efficient way to do this, but with list of favorites <=30, no problem

src/PKSim.Presentation/Presenters/PopulationAnalyses/PopulationAnalysisParameterFieldsPresenter.cs
```diff
@@ -74,25 +73,18 @@ public void AddCovariate(string covariate)
...
-            ParameterFieldSelected(this, new ParameterFieldSelectedEventArgs(parameterFromField(parameterField), parameterField));
-            return;
-         }
-
-         var covariateField = populationAnalysisField as PopulationAnalysisCovariateField;
-         if (covariateField != null)
-         {
-            CovariateFieldSelected(this, new CovariateFieldSelectedEventArgs(covariateField.Covariate, covariateField));
-            return;
+            case PopulationAnalysisParameterField parameterField:
```
msevestre: This new switch by type syntax is so much nicer. Liskov says we don't need it but when you do, men it's better

src/PKSim.UI/Views/Parameters/MultiParameterEditView.cs
```diff
@@ -48,14 +47,15 @@ public MultiParameterEditView(IToolTipCreator toolTipCreator, IImageListRetrieve
       {
          _pathElementsBinder = pathElementsBinder;
          InitializeComponent();
-         Initialize(gridViewParameters);
+         Initialize(gridView);
          initRepositories();
          //Allow cell selection
-         gridViewParameters.OptionsSelection.EnableAppearanceFocusedRow = true;
```
msevestre: naming it sthg else than gridview makes identifying similar code harder

src/PKSim.UI/Views/Parameters/MultiParameterEditView.cs
```diff
@@ -239,11 +239,32 @@ public void FixParameterColumnWidth(int parameterWitdh)
...
       }
 
-      public IEnumerable<ParameterDTO> SelectedParameters => gridViewParameters.DataController.GetAllFilteredAndSortedRows().Cast<ParameterDTO>();
+      public IEnumerable<ParameterDTO> AllVisibleParameters => gridView.DataController.GetAllFilteredAndSortedRows().Cast<ParameterDTO>();
+
+      public IReadOnlyList<ParameterDTO> SelectedParameters
+      {
+         get { return gridView.GetSelectedRows().Select(rowHandle => _gridViewBinder.ElementAt(rowHandle)).ToList(); }
+         set
+         {
```
msevestre: Little bit hugly here to select rows programmatically. I need to find the row handles , select the rows and then set the focus row handle. Does not work if rows are not continuously selected but this is not possible in this view

PR #577, opened 2018-02-20
Fixes #573 value origin bb not updated

### LINKED ISSUE #573: ValueOrigin: BB in simulation state not updated
If a ValueOrigin is changed in the simulation for a BuildingBlock-Parameter, Building block is not marked as changed in the simulation tree.
Update/Commit actions however are available in the context menu - so just an icon is nicht updated.
clipboard01

msevestre:


src/PKSim.Core/Services/ParameterTask.cs
```diff
@@ -483,7 +483,7 @@ public ICommand UpdateParameterValueOrigin(IParameter parameter, ValueOrigin val
 
       public ICommand SetParameterValueOrigin(IParameter parameter, ValueOrigin newValueOrigin)
       {
-         return setParameterValueOrigin(parameter, newValueOrigin, shouldChangeVersion: false);
+         return setParameterValueOrigin(parameter, newValueOrigin, shouldChangeVersion: true);
```
msevestre: I had implemented the action as NOT change version at the beginning and forgot to update the code once we changed our mind

PR #578, opened 2018-02-21
Fixes #560 Value Origin export so snaphsot should not save value orig…

### LINKED ISSUE #560: Value Origin export so snaphsot should not save value origin if value origin can be loaded from db
Id defined and ValueOrigin == DBValueOrigin. In that case, the info does not need to be saved

msevestre:
…in if value origin can be loaded from db

Also fixed a bunch of small issue related to how we moved from Value has changed to IsDefault flag

### DIFF
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -874,6 +874,8 @@ namespace PKSim.Assets
          public static string SnapshotFileMismatch(string desiredType) => $"Snapshot file cannot be used to load a {desiredType.ToLowerInvariant()}.";
 
          public static string CannotLoadRelatedItemAsObjectAlreadyExistInProject(string objectType, string objectName) => $"Cannot load related item into project. A {objectType.ToLower()} named '{objectName}' already exists.";
+
+         public static string CompoundGroupNotFoundFor(string compoundGroup, string compoundName) => $"Cannot find compound group '{compoundGroup}' for compound '{compoundName}'";
       }
 
       public static class Information
@@ -1486,7 +1488,6 @@ namespace PKSim.Assets
          public static readonly string EventFolder = Events;
          public static readonly string AdministrationProtocolFolder = "Administration Protocols";
          public const string Value = "Value";
-         public const string ValueDescription = "Value Description";
          public static readonly string Percentage = "Percentage";
          public static readonly string Container = "Container";
          public static readonly string Percentile = "Percentile";
--- a/src/PKSim.Core/Snapshots/Mappers/IndividualMapper.cs
+++ b/src/PKSim.Core/Snapshots/Mappers/IndividualMapper.cs
@@ -45,7 +45,7 @@ namespace PKSim.Core.Snapshots.Mappers
 
       private Task<LocalizedParameter[]> allParametersChangedByUserFrom(ModelIndividual individual)
       {
-         var changedParameters = individual.Organism.GetAllChildren<IParameter>(x => x.ValueDiffersFromDefault());
+         var changedParameters = individual.Organism.GetAllChildren<IParameter>(x => x.ParameterHasChanged());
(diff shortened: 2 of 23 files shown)


src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -874,6 +874,8 @@ public static string MapToSnapshotNotSupportedWithoutContext(string snapshotType
          public static string SnapshotFileMismatch(string desiredType) => $"Snapshot file cannot be used to load a {desiredType.ToLowerInvariant()}.";
 
          public static string CannotLoadRelatedItemAsObjectAlreadyExistInProject(string objectType, string objectName) => $"Cannot load related item into project. A {objectType.ToLower()} named '{objectName}' already exists.";
+
+         public static string CompoundGroupNotFoundFor(string compoundGroup, string compoundName) => $"Cannot find compound group '{compoundGroup}' for compound '{compoundName}'";
```
msevestre: Unsure if those kind of error should be logged or throw exception. We should decide of the best strategy in those cases.
This happens if the alternative selected in the simulation does not exist in the compound. Typically editing the json by hand

src/PKSim.Core/Snapshots/Mappers/IndividualMapper.cs
```diff
@@ -45,7 +45,7 @@ private Task<Molecule[]> allMoleculesFrom(ModelIndividual individual)
 
       private Task<LocalizedParameter[]> allParametersChangedByUserFrom(ModelIndividual individual)
       {
-         var changedParameters = individual.Organism.GetAllChildren<IParameter>(x => x.ValueDiffersFromDefault());
+         var changedParameters = individual.Organism.GetAllChildren<IParameter>(x => x.ParameterHasChanged());
```
msevestre: this was not using the right method

src/PKSim.Core/Snapshots/Mappers/OutputIntervalMapper.cs
```diff
@@ -24,6 +24,12 @@ public override Task<SnapshotOutputInterval> MapToSnapshot(ModelOutputInterval o
          });
       }
 
+      protected override bool ShouldExportParameterToSnapshot(IParameter parameter)
+      {
+         //we want to ensure that start time and end time are always exported
+         return parameter.NameIsOneOf(Constants.Parameters.START_TIME, Constants.Parameters.END_TIME) || base.ShouldExportParameterToSnapshot(parameter);
```
msevestre: Start time and end time should always be exported because their state is set to IsDefault = true when created

src/PKSim.Core/Snapshots/Mappers/ParameterMapper.cs
```diff
@@ -88,7 +88,10 @@ private void updateParameterValue(SnapshotParameter snapshot, double parameterBa
          var snapshotValueInBaseUnit = parameter.ConvertToBaseUnit(snapshot.Value);
 
          if (!ValueComparer.AreValuesEqual(baseValue, snapshotValueInBaseUnit))
```
msevestre: Not convinced that this is required anymore. If the parameter is in the snapshot, then the value should be overwrittten. I don't think this is wrong however. TO discuss

src/PKSim.Core/Snapshots/Mappers/ValueOriginMapper.cs
```diff
@@ -21,6 +29,12 @@ public override Task<SnapshotValueOrigin> MapToSnapshot(ModelValueOrigin valueOr
          });
       }
 
+      private bool valueOriginComesFromDb(ModelValueOrigin valueOrigin)
+      {
+         var dbValueOrigin = _valueOriginRepository.FindBy(valueOrigin.Id);
```
msevestre: if there is a value origin with the id, then only save if not the same. That works also for id not defined becuase it returns the default value origin (undefined and empty description)

PR #585, opened 2018-02-22
Filled some value origins

Yuri05:
Didn't define any tests for that so far; will be added with next portion of value origins
(any good ideas for tests here? just exemplaric picks or percentage of filled value origins or ...?)


msevestre:
> just exemplaric picks
Yes that's what I would do. At that stage, we cannot test the whole system

PR #587, opened 2018-02-23
Fixes #586 messed up relative expressions

### LINKED ISSUE #586: Messed up relative expression
1) Create a new PK-Sim project. Load the individual "Standard European Male for DDI" from template.
2) Create any compound and administration protocol and create a simulation.
3) Compare the expression of the enzyme AADAC in the individual BB and the simulation. They are identical.
4) In the Individual-BB, change the relative expression of AADAC in vascular endothelium to 2.

relexp_bb

5) Update the simulation.
6) Compare thge expression of the enzyme in the BB and simulation again. The expression in the simulation is totally messed up.

relexp_sim

It seems that some "normalized expression" values from the BB are treated like "relative expression" in the simulation - e.g., the liver expression values. But the simulation seems to normalize the expression once again!

The expression values in intestinal mucosa are absolutely unclear - where do the values of 0.0684 come from??

msevestre:


msevestre:
This is quite the bug. It has been in since the very beginning :). When we update relative expression, we were triggering the scaling of all other relative expression as well that were then being updated as part of the update process etc... so some stuff was scaled twice. 
In this implementation, my suggestion is to NEVER scale parameter as part of the ParameterUpdater call. This is call to synchronize a set of parameters and as such, the caller will always update RelativeExpression AND RelativeExpression_norm. I tested it for quite some time and it is working as expected

PR #588, opened 2018-02-28
 Fixes #580 Add Parameter Identification to snapshot 

### LINKED ISSUE #580: Add Parameter Identification to snapshot

msevestre:


msevestre:
@Yuri05 This is a big commit here that should be merged with Squash and merge

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -105,7 +106,22 @@ public static bool ValueDiffersFromDefault(this IParameter parameter)
          return !ValueComparer.AreValuesEqual(parameter.Value, parameter.DefaultValue.Value, CoreConstants.DOUBLE_RELATIVE_EPSILON);
       }
 
-      public static bool ParameterHasChanged(this IParameter parameter) => parameter.IsDefault == false;
+      public static bool ShouldExportToSnapshot(this IParameter parameter)
```
msevestre: I realized that this was not behaving as expected. We cannot export NaN parameters to json either

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -105,7 +106,22 @@ public static bool ValueDiffersFromDefault(this IParameter parameter)
...
+      {
+         if (parameter.IsDefault)
+            return false;
+
+         try
+         {
+            var v = parameter.Value;
```
msevestre: It would be nice to have a method TryGetValue instead of always having to check

msevestre:
@Yuri05 I need you to review this code before I move on. Can you do that Friday?

msevestre:
@Yuri05 Ok going to merge this because I need to push my next changes. Please review

src/PKSim.Core/Snapshots/IdentificationParameter.cs
```diff
@@ -0,0 +1,12 @@
+﻿using OSPSuite.Core.Domain;
+
+namespace PKSim.Core.Snapshots
+{
+   public class IdentificationParameter : ParameterContainerSnapshotBase
+   {
+      public Scalings Scaling { get; set; }
+      public bool? UseAsFactor { get; set; }
+      public bool? IsFixed { get; set; }
+      public string[] LinkedParameters { get; set; }
```
Yuri05: Why are `UseAsFactor  `and `IsFixed `defined as nullable?

msevestre: The way the snapshot engine works is that it allows to NOT specify default value (since the goal was to only write what was set by the user). Typically, UseAsFactor and IsFixed are not used (in probably 90% of the cases and even more). No need to save a value. When deserializing, missing value is interpreted as false

So instead of true or false, we are using the concept of truthy and falsy (falsy = false, null in that case)

Yuri05: The question for me is what should happen if we change a default.
In case of e.g. a physiological parameter the new default should be used.

But here I'm not quite sure. If we change the default for *UseAsFactor*: should the PI recreated from snapshot use this new default or the previous value? (and I would vote for the previous one)

Anyway, default values for the settings above will probably never be changed, so we can keep it as is.

msevestre: @Yuri05 I believe this is the correct behavior. If we change the default!, we expect the snapshot to run with the default. No result is saved so the PI would result in the same behavior as IF the snapshot was created from scratch. 
Now to be honest, this does not really matter for those guys and we can do whatever . But this logic of sticking with App default is present everywhere and just not here as per requirement. (Only save, what was set by the user)

PR #595, opened 2018-04-03
Fixes  #540 CSV import export

### LINKED ISSUE #540: CSV: Export parameters using ""
I know this will raise a lot of discussions regarding compatibility with older scripts but I think we should discuss the issue anyways. 

At the moment, exporting a population to csv and re-importing it in PKSim is just a **massive pain** because our csv are separated with ";" and not with "," which is how excel works.

We cannot move to "," because we have some parameters with "," in names

My suggestion would be to look at parameter name systematically and 
* replace , with _
* replace " " with _

so a parameter like `logP (oil/water,pKa)` would be renamed to `logP_(oil/water_pKa)`

This would ensure that we can always export to csv without any issues in excel and also prevent empty space in path.

Display name would be adjusted so that _ is repalced with " " 
so `logP (oil/water,pKa)` would become `logP (oil/water pKa)`

With simple logic, it should be possible to propose an error message if a parameter is not found and also it would allow a user to systematically know how a parameter is named in MoBi or in Matlab from its display value in PKSim (simply replace " " with _)

msevestre:


msevestre:
@Yuri05 Basically updating all nuget packages that I had to change. THe main logic change is really in the IndividualImporter where the new authorized , is defined. Trimming "" happens automatically in header when using the CSV importer so it works out of the box

src/PKSim.Infrastructure/Services/IndividualPropertiesCacheImporter.cs
```diff
@@ -28,21 +28,22 @@ public IndividualPropertiesCacheImporter(IGenderRepository genderRepository, IPo
          _populationRepository = populationRepository;
       }
 
-      private const int PARAMETER_PATH = 0;
-      private readonly Char[] _allowedDelimiters = {';', '\t'};
+      private static readonly Char[] ALLOWED_DELIMITERS = {',', ';', '\t'};
```
msevestre: First change: Support for ','

src/PKSim.Infrastructure/Services/IndividualPropertiesCacheImporter.cs
```diff
@@ -52,28 +53,23 @@ public IndividualPropertiesCache ImportFrom(string fileFullPath, IImportLogger l
...
-
          using (var reader = new CsvReaderDisposer(fileFullPath, delimiter))
          {
             var csv = reader.Csv;
             var headers = csv.GetFieldHeaders();
             if (headers.Contains(Constants.Population.INDIVIDUAL_ID_COLUMN))
-               loadNewCSVFormat(individualPropertiesCache, csv, headers);
-            else
-            {
-               loadOldCSVFormat(individualPropertiesCache, csv);
```
msevestre: Remove loading of very old PKSim csv file format. (not used since 2013!)

src/PKSim.Infrastructure/Services/CsvReaderDisposer.cs
```diff
@@ -10,24 +10,19 @@ public class CsvReaderDisposer : IDisposable
...
       private readonly CsvReader _csv;
 
-      public CsvReader Csv
-      {
-         get { return _csv; }
-      }
+      public CsvReader Csv => _csv;
 
-      public CsvReaderDisposer(string fileFullPath, char delimiter = ';')
+      public CsvReaderDisposer(string fileFullPath, char delimiter = ',')
```
msevestre: default is now ,

msevestre:
@Yuri05 A couple tests failing I see. I'll fix them tomorrow. Please review the PR if you can and I'll merge after that :)

msevestre:
@Yuri05 Ok this should be good to merge

PR #598, opened 2018-04-04
Fixes #529 Adding new metabolization process - metabolite selection i…

### LINKED ISSUE #529: Adding new metabolization process - metabolite selection ignored
1) Open the "Paracetamol"-example, or any other PK-Sim project with metabolizing enzyme expression and at least two compounds (in the latter case, open the respective compounds in the next steps)
2) Open the "Paracetamol"-compound and navigate to the tab "ADME"
3) Select "Add metabolizing Enzyme". Select any expressed enzyme, and some other compound under "Metabolite". As "process type", select "In vitro clearance - First Order" (important step to reproduce).

image

4) After pressing "OK", no metabolite is selected.
image

msevestre:
…gnored

### DIFF
--- a/src/PKSim.Presentation/Presenters/Compounds/CreateEnzymaticProcessPresenter.cs
+++ b/src/PKSim.Presentation/Presenters/Compounds/CreateEnzymaticProcessPresenter.cs
@@ -8,6 +8,7 @@ using PKSim.Presentation.DTO.Mappers;
 using PKSim.Presentation.Presenters.Parameters;
 using PKSim.Presentation.Views.Compounds;
 using OSPSuite.Assets;
+using PKSim.Presentation.DTO.Compounds;
 
 namespace PKSim.Presentation.Presenters.Compounds
 {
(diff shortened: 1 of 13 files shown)


src/PKSim.Presentation/Presenters/Compounds/CreateEnzymaticProcessPresenter.cs
```diff
@@ -42,16 +44,33 @@ private IEnumerable<Compound> allPossibleMetabolites()
          return _buildingBlockRepository.All<Compound>()
             .OrderBy(x => x.Name);
       }
+      
+      protected override void Rebind(EnzymaticProcess enzymaticProcess, Compound compound)
```
msevestre: This is the actual code here. It is required to save the metabolite name before updating the view with thew new selected template

msevestre: This is really the only change in this code. All the rest is code reformatting using terser syntax and tests

tests/PKSim.Tests/Presentation/CreatePartialProcessPresenterSpecs.cs
```diff
@@ -83,5 +92,11 @@ public void should_not_reset_the_name_and_protein_entered_by_the_user()
          _process2DTO.MoleculeName.ShouldBeEqualTo("Protein");
          _process2DTO.DataSource.ShouldBeEqualTo("Lab");
       }
+      
+      [Observation]
+      public void should_not_reset_the_name_of_the_metabolite()
+      {
+         _enzymaticPartialProcessDTO.Metabolite.ShouldBeEqualTo(_template1.MetaboliteName);
```
msevestre: That's the test here. Before the Metabolite was empty. With the implementation, the Metabolite is whatever was set before the update

PR #599, opened 2018-04-06
Fixes #481 CSV Import of population should allow txt files

### LINKED ISSUE #323: Renaming parameter in SA should also rename the parameter in the SA analysis
Right now it is required to RERUN the analysis to update the caption!

### LINKED ISSUE #481: CSV Import of population should allow txt files
Importing a file exported from PK-Sim works like a charmed but this scenario is of course useless.
Usually a population is:

1. exported to a csv file 
1. which is opened in Excel and values are e.g. modified, added removed etc..
1. the file is saved (which creates a lot of problems if you are not careful)
1. the file is imported in PKSim

Step 2. is already somewhat a pain because PK-Sim csv does not support the default csv with "," separated values. Instead the ";" is used as separator, which Excel does not recognized automatically

Step 3. is critical. Once a file is opened in Excel and saved, the "," are BACK in the file as separator and the file cannot be used in PK-Sim anymore. Instead one needs to save the file as "TAB" separated file, which is a txt file by default

Step 4. The txt files exported previously cannot be found by default. Instead one need to enter the wildcard search *.* to find the tab delimited file

Suggestion: At least make Step 4 easier and allow for *.txt extensions to be found out of the box

msevestre:
Also Fixes #323 by virtue of updating DLL

### DIFF
--- a/src/PKSim.Assets/PKSim.Assets.csproj
+++ b/src/PKSim.Assets/PKSim.Assets.csproj
@@ -68,8 +68,8 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="OSPSuite.Assets, Version=7.3.0.47, Culture=neutral, processorArchitecture=x86">
-      <HintPath>..\..\packages\OSPSuite.Assets.7.3.0.47\lib\net462\OSPSuite.Assets.dll</HintPath>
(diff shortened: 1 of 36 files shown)


msevestre:
@Yuri05 Mostly update of dll again. Simply extended the filter to support *.txt

Yuri05:
1 test failed

PR #610, opened 2018-04-18
 Fixes #248 solubility as table 

### LINKED ISSUE #248: Define solubility as a table (X=pH, Y = Sol)
* Values can be edited by hand
* Values can be loaded from file

similar to Table Formulation UI

msevestre:


msevestre:
@Yuri05 Big pull request that should basically support Table solubility in PK-Sim
There is a test that creates a simulation using Table Solubility. This test WILL fail as long as the implementation is not supported in SimModel.

msevestre:
@Yuri05 
Table is exported as follow:
```
 "UseDerivedValues": true,
        "Points": [
          {
            "X": 5.0,
            "Y": 22.0,
            "RestartSolver": false
          },
          {
            "X": 10.0,
            "Y": 50.0,
            "RestartSolver": false
          }

```

Is this correct (I do not know what `UseDerivedValues` means anymore)

Yuri05:
>Is this correct

No, must be `"UseDerivedValues": false`
Otherwise dY/dX (here: dSolubility/dpH) will be used in the simulation

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -2157,6 +2160,7 @@ public static class UI
          public static readonly string UseWatermark = "Use watermark in charts when exporting to clipboard?";
          public static readonly string WatermarkText = "Text";
          public static readonly string WatermarkProperties = "Watermark Properties";
+         public static readonly string CreateTableSolubilityAlternative = "Create as pH-Sol table";
```
Yuri05: Think _Create as pH-Sol**ubility** table_ would be better

src/PKSim.Core/Model/FormulaFactory.cs
```diff
@@ -311,6 +311,9 @@ private IFormula createFormula(RateKey rateKey)
          if (rateKey.IsTableWithOffsetFormula)
```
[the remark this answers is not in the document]
Yuri05: might be required for old projects (unless we don't convert old **TableWithOffset** formulas into **TableWithXArg**)

src/PKSim.Assets/PKSimConstants.cs
msevestre: :)
image

changed

src/PKSim.Core/Model/FormulaFactory.cs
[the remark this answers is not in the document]
msevestre: No this is only used when creating  a new Formula using the RateKey (e..g from DB).
That should never happen anymore. TableWithOffset will still be available as an object for backward compatibility

src/PKSim.Core/Model/FormulaFactory.cs
```diff
@@ -348,6 +351,27 @@ private IFormula createFormula(RateKey rateKey)
...
+         formula.AddTableObjectPath(tableObjectPath);
+         formula.AddXArgumentObjectPath(xArgument);
+
+         //Table formula with reference has the same dimension as its referenced table object
+         formula.Dimension = tableObjectPath.Dimension;
+
+         return formula;
+      }
+
       private IFormula createTableFormulaWithOffset(RateKey rateKey)
```
msevestre: All of that should go to

src/PKSim.Assets/PKSimConstants.cs
msevestre: done

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -2238,13 +2238,25 @@ public static string ImportFormulationDescription
...
+               sb.AppendLine("The Excel table for formulation import should have <b>two colums</b>.");
+               sb.AppendLine(" -  The first column represents the Time (unit Time e.g. in hours).");
+               sb.AppendLine(" -  The second column represents the fraction of the dose released at t.");
                return sb.ToString();               
             }
-         } 
-         
+         }
+
+         public static string ImportSolubilityTableDescription
```
msevestre: Added a description too

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -2238,13 +2238,25 @@ public static string ImportFormulationDescription
             get
             {
                var sb = new StringBuilder();
-               sb.AppendLine("The Excel table for formulation import should have a format of two colums.");
```
msevestre: this was not not english

src/PKSim.Core/Model/FormulaFactory.cs
```diff
@@ -260,9 +261,10 @@ public IFormula RateFor(RateKey rateKey, IFormulaCache formulaCache)
          return formulaCache[rateKey];
       }
 
-      public TableFormula CreateTableFormula()
+      public TableFormula CreateTableFormula(bool useDerivedValues = true)
```
msevestre: add the parameter. Default is true (same behavior as previous)

src/PKSim.Core/Model/ParameterAlternativeFactory.cs
```diff
@@ -59,11 +59,10 @@ public ParameterAlternative CreateTableAlternativeFor(ParameterAlternativeGroup
          var alternative = CreateAlternativeFor(compoundParameterGroup);
 
          var tableParameter = alternative.Parameter(tableParameterName);
-         var tableFormula = _formulaFactory.CreateTableFormula();
+         var tableFormula = _formulaFactory.CreateTableFormula(useDerivedValues:false);
```
msevestre: creates with false in this case

PR #612, opened 2018-04-18
Fixes #602 Portable version: locating PK-Sim DB

### LINKED ISSUE #602: Portable version: locating PK-Sim DB
Search order for the PK-Sim DB should be:
1. Application Folder
2. ProgramData folder

This would enable using portable and installed version at the same time

msevestre:


msevestre:
Straight forward changes to account for update of core.

PR #614, opened 2018-04-19
Fixes #504 Allow multiselect when importing templates from template d…

### LINKED ISSUE #504: Allow multiselect when importing templates from template database
Right now, only one item at a time can be imported.

msevestre:
…atabase

### DIFF
--- a/src/PKSim.Presentation/Presenters/TemplatePresenter.cs
+++ b/src/PKSim.Presentation/Presenters/TemplatePresenter.cs
@@ -1,21 +1,22 @@
 using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
-using PKSim.Assets;
+using OSPSuite.Assets;
+using OSPSuite.Core;
+using OSPSuite.Core.Domain.Services;
 using OSPSuite.Core.Services;
+using OSPSuite.Presentation.Core;
 using OSPSuite.Presentation.Nodes;
+using OSPSuite.Presentation.Presenters;
+using OSPSuite.Presentation.Presenters.ContextMenus;
+using OSPSuite.Presentation.Views;
 using OSPSuite.Utility.Collections;
+using OSPSuite.Utility.Extensions;
+using PKSim.Assets;
 using PKSim.Core.Model;
 using PKSim.Core.Services;
 using PKSim.Presentation.Nodes;
 using PKSim.Presentation.Views;
-using OSPSuite.Core;
-using OSPSuite.Assets;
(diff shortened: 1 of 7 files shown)


src/PKSim.Presentation/Presenters/TemplatePresenter.cs
```diff
@@ -36,9 +37,9 @@ public interface ITemplatePresenter : IDisposablePresenter, IPresenterWithContex
       IEnumerable<Template> AllTemplates();
 
       /// <summary>
-      ///    This method is called whenever a node is being selected in the view
+      ///    This method is called whenever a nodes are being selected in the view
       /// </summary>
-      void ActivateNode(ITreeNode node);
+      void ActivateNodes(IReadOnlyList<ITreeNode> treeNodes);
```
msevestre: Multiselect enabled=> many nodes can be selected

src/PKSim.Presentation/Presenters/TemplatePresenter.cs
```diff
@@ -159,18 +161,24 @@ private void updateIcon(TemplateType templateType)
 
       private void updateView()
       {
-         _view.OkEnabled = (_selectedTemplate != null);
-         _view.Description = _selectedTemplate == null ? string.Empty : _selectedTemplate.Description;
+         var numberOfTemplateSelected = _selectedTemplates.Count;
+         _view.OkEnabled = numberOfTemplateSelected > 0;
+
+         _view.Description =
```
msevestre: All those changes are required to take multiple selection into account

src/PKSim.Presentation/Services/CompoundTask.cs
```diff
@@ -37,42 +37,6 @@ public class CompoundTask : BuildingBlockTask<Compound>, ICompoundTask
...
-         if (loadedCompounds.Count <= 1)
-            return loadedCompounds;
-
-         //one compound and one or more metabolites
-         warnUserIfMetabolitesDoNotMatch(loadedCompounds);
-
-         return loadedCompounds;
-      }
-
-      private void warnUserIfMetabolitesDoNotMatch(IReadOnlyList<Compound> loadedCompounds)
```
msevestre: There is no easy way to actually check that. This was only required when importing a metabolite that already existed in the project. In that case, the metabolite was renamed but not the name of metabolite in the parent compound . Absolute edge case that probably never happens. Now that we import multiple compound at the same time, that code does not work and it would be quite hard to find a logic that make sense. => REMOVED

PR #618, opened 2018-04-23
Fixes #513 multi select export template, Fixes #316 Fixes #513

### LINKED ISSUE #316: Favorite parameters that should not be variable should be ignored in PI when creating a PI automatically
Right now, if a parameter has the flag for CanBeVaried = false and is set as favorite, it will be added automatically to a PI. The `IParameterAnalysableParameterSelector` should be used to ensure that favorite can indeed be added.

### LINKED ISSUE #513: Allow multiselect when exporting templates to template database

msevestre:


src/PKSim.Presentation/Presenters/ContextMenus/MultipleEventNodeContextMenu.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Collections.Generic;
+using OSPSuite.Presentation.Nodes;
+using OSPSuite.Presentation.Presenters;
+using OSPSuite.Presentation.Presenters.ContextMenus;
+using PKSim.Core;
+using PKSim.Core.Model;
+
+namespace PKSim.Presentation.Presenters.ContextMenus
+{
+   public class MultipleEventNodeContextMenu : MultipleBuildingBlockNodeContextMenu<PKSimEvent>
```
msevestre: I had to create one context menu for each type as a special call was required for Compound. That way, we can change the multiselect menu based on building block types if we feel like iit

msevestre:
@Yuri05 need to merge

PR #622, opened 2018-04-24
Fixes #621 reintegrate changes from hotfix

### LINKED ISSUE #621: Reintegrate changes from hotfix 7.2.3

msevestre:


msevestre:
I have added every single commit from hotfix into this branch. Please review quickly and let me know if you think something is missing :)

src/PKSim.Core/Services/ParticleApplicationCreator.cs
```diff
@@ -77,6 +75,12 @@ private void addDrugStartFormulaForParticles(IApplicationBuilder applicationBuil
             setupParticleBin(applicationBuilder, i, binRadius[i - 1], binNumberOfParticlesFactor[i - 1], formulaCache);
          }
 
+         //// add insoluble molecule for particle bin
```
msevestre: ?? Why is this still there?

tests/PKSim.Tests/Infrastructure/DomainFactoryForSpecs.cs
```diff
@@ -47,6 +48,24 @@ public static Protocol CreateStandardIVProtocol()
...
+      }
+
+      public static Formulation CreateParticlesFormulation(int numberOfBins)
+      {
+         var formulationRepository = IoC.Resolve<IFormulationRepository>();
+         var formulation = formulationRepository.FormulationBy(CoreConstants.Formulation.Particles);
+         formulation.Parameter(CoreConstants.Parameters.NUMBER_OF_BINS).Value = numberOfBins;
+         
+         //set mono/polydisperse property (0=mono, 1=poly).
+         formulation.Parameter(CoreConstants.Parameters.PARTICLE_DISPERSE_SYSTEM).Value = (numberOfBins > 1) ? 1 : 0;
```
msevestre: We have constant for that
         public const int MONODISPERSE = 0;
         public const int POLYDISPERSE = 1;

tests/PKSim.Tests/IntegrationTests/FlatParameterInContainsRepositorySpecs.cs
```diff
@@ -33,7 +34,11 @@ public class When_checking_the_database_consistency_regarding_parameter_flags :
       [Observation]
       public void all_hidden_parameters_should_be_read_only()
       {
-         var parameters = _allParameters.Where(x => !x.Visible).Where(x => !x.ReadOnly).ToList();
+         var parameters = _allParameters.Where(x => !x.Visible)
+            .Where(x => !x.ReadOnly)
+            .Where(x => x.ParameterName != CoreConstants.Parameters.SOLUBILITY_TABLE)
```
msevestre: This is new. This tests was failing because by defaut, solubility table is hidden but not readonly

src/PKSim.Core/Services/ParticleApplicationCreator.cs
Yuri05: Once we have adjusted our application/formulation structure, we can think about reducing the number of state variables by moving Insoluble drug fro Bin-Level to Application (or better: formulation) level.
Then those lines would be used.

Yuri05:
I don't see the database in the list of changed files?

msevestre:
Database is in 7.3 branch already. Nothing to change indeed

Yuri05:
looks good. All failing tests are because of missing implementation of TableWithXargument in SimModel

PR #632, opened 2018-04-27
Fix particles integration tests

Yuri05:


tests/PKSim.Tests/IntegrationTests/SimulationWithParticlesFormulationSpecs.cs
```diff
@@ -488,6 +488,8 @@ private void checkValuesMonotonous(float[] values, bool checkDecreasing)
             if (checkDecreasing)
                diff *= -1;
 
+            diff += (100 * _simulation.Solver.AbsTol).ToFloat(); //ignore "small" negative differences
+
             diff.ShouldBeGreaterThanOrEqualTo(0);
```
Yuri05: while checking that values are increasing: add a small offset (based on the absolute tolerance) to the difference.
Thus small negative differences will not prevent the test from passing

PR #633, opened 2018-04-27
 Fixes #629 issue with conversion snapshot 

### LINKED ISSUE #629: Issue with convertion from batch file to snapshot files

msevestre:


src/PKSim.CLI.Core/Batch/Mapper/CompoundMapper.cs
```diff
@@ -181,7 +181,16 @@ private void setValue(ModelCompound compound, string alternativeName, string par
       {
          var alternative = compound.ParameterAlternativeGroup(alternativeName).DefaultAlternative;
          var parameter = alternative.Parameter(parameterName);
+         setValue(parameter, value);
+      }
+
+      private void setValue(IParameter parameter, double value)
```
msevestre: Defining this method in all mappers was key to get the expected snapshot after convesion. Most of this code will be removed once we have an agreement on all snapshot anyways

src/PKSim.CLI.Core/Batch/Mapper/IndividualMapper.cs
```diff
@@ -79,15 +77,22 @@ public Model.Individual MapFrom(Individual batchIndividual)
...
          return newMolecule.DowncastTo<TMolecule>();
       }
+
+
+      private void setValue(IParameter parameter, double value)
+      {
+         if (ValueComparer.AreValuesEqual(parameter.Value, value))
+            return;
+
+         parameter.Value = value;
```
msevestre: same code everywhere. Because code will be deleted soon, don't care for once

src/PKSim.Core/Services/ParameterTask.cs
```diff
@@ -69,22 +69,34 @@ public interface IParameterTask : ISetParameterTask
...
       ///    Sets the value in the parameter. Value will be take as is
       /// </summary>
       /// <param name="parameter">Parameter</param>
       /// <param name="value">Value in kernel unit</param>
-      ICommand SetParameterValue(IParameter parameter, double value);
+      /// <param name="shouldUpdateDefaultStateAndValueOriginForDefaultParameter">
+      ///    If set to <c>true</c> default, default state
+      ///    and value origin of parameter will be updated as well
+      /// </param>
+      ICommand SetParameterValue(IParameter parameter, double value, bool shouldUpdateDefaultStateAndValueOriginForDefaultParameter = true);
```
msevestre: It was required to add this extra parameter. True by default no avoid behavior change
But when creating a simulation based on building block, I need to use this flag to avoid updating the IsDefault flag to false when not required

src/PKSim.Core/Services/ParameterUpdater.cs
```diff
@@ -121,6 +121,9 @@ private ICommand resetParameterValue(IParameter targetParameter)
 
       private ICommand setParameterValue(IParameter sourceParameter, IParameter targetParameter, bool forceUpdate)
       {
+         //If the parameter we are updating from is a default parameter, the target parameter should either remain as is (default or not default)
+         bool shouldUpdateDefaultState = !sourceParameter.IsDefault;
```
msevestre: The source parameter form the update is a default parmaeter. Updating the target parameter form the source parameter should not update the default state in the target parameter to false. Instead we decide here if we need to update

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
```diff
@@ -119,10 +126,72 @@ private Task<CompoundProperties[]> usedCompoundsFrom(ModelSimulation simulation,
...
+
+         return _parameterMapper.LocalizedParametersFrom(allParametersToExport);
+      }
+
+      private static bool canFilterUnchangedParameters(PKSimBuildingBlockType buildingBlockType)
+      {
+         return buildingBlockType.IsOneOf(PKSimBuildingBlockType.Compound, PKSimBuildingBlockType.Formulation, PKSimBuildingBlockType.Individual, PKSimBuildingBlockType.Event);
+      }
+
+      private IEnumerable<IParameter> parametersChangedFromBuildingBlock(IGrouping<PKSimBuildingBlockType, IParameter> parametersByBuildingBlockType, ModelSimulation simulation, PKSimProject project)
```
msevestre: Code written quickly tested but no unit test yet. This ensures that I do not write parameters that haven't changed from building block to snapshot

msevestre: Protocol excluded because we still need to refactor application to avoid parameter duplication. This should be done in next itereation in my opinion

src/PKSim.CLI.Core/Batch/Mapper/ApplicationProtocolMapper.cs
```diff
@@ -27,10 +27,17 @@ public Protocol MapFrom(ApplicationProtocol batchProtocol)
...
+         simpleProtocol.EndTimeParameter.IsDefault = false;
          simpleProtocol.Dose.DisplayUnit = simpleProtocol.Dose.Dimension.Unit(batchProtocol.DoseUnit);
          simpleProtocol.Dose.Value = simpleProtocol.Dose.Dimension.UnitValueToBaseUnitValue(simpleProtocol.Dose.DisplayUnit, batchProtocol.Dose);
+         simpleProtocol.Dose.IsDefault = false;
          simpleProtocol.DosingInterval = DosingIntervals.ById(dosingIntervalId);
 
+         if (!simpleProtocol.IsSingleDosing)
+         {
+            simpleProtocol.EndTimeParameter.IsDefault = false;
+         }
```
Yuri05: ??? 
Isn't `simpleProtocol.EndTimeParameter.IsDefault = false` anyway because of line 30?

src/PKSim.Core/Services/ParameterTask.cs
```diff
@@ -273,9 +285,9 @@ public ICommand SetParameterDisplayValueAsStructureChange(IParameter parameter,
          return SetParameterValueAsStructureChange(parameter, parameter.ConvertToBaseUnit(valueToSetInGuiUnit));
       }
 
-      public ICommand SetParameterValueAsStructureChange(IParameter parameter, double value)
+      public ICommand SetParameterValueAsStructureChange(IParameter parameter, double value, bool shouldUpdateDefaultStateAndValueOriginForDefaultParameter = true)
       {
-         return withUpdatedDefaultStateAndValue(new SetParameterValueStructureChangeCommand(parameter, value).Run(_executionContext), parameter);
+         return withUpdatedDefaultStateAndValue(new SetParameterValueStructureChangeCommand(parameter, value).Run(_executionContext), parameter, shouldUpdateDefaultStateAndValueOriginForDefaultParameter: shouldUpdateDefaultStateAndValueOriginForDefaultParameter);
       }
```
Yuri05: `shouldUpdateDefaultStateAndValueOriginForDefaultParameter:shouldUpdateDefaultStateAndValueOriginForDefaultParameter);` 
named parameter not really required here

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
Yuri05: What about **Population**-BB? Why it's also excluded?

src/PKSim.CLI.Core/Batch/Mapper/ApplicationProtocolMapper.cs
msevestre: No. For non singledosing, end parameter is visible and should not be exported to snapshot (e.g.IsDefault false)
What's the problem?

src/PKSim.Core/Services/ParameterTask.cs
msevestre: No. There is another default parameter so name parameter is required

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
msevestre: I don't think we have parameter in population BB but I can add it for completion

src/PKSim.CLI.Core/Batch/Mapper/ApplicationProtocolMapper.cs
Yuri05:
```
simpleProtocol.EndTimeParameter.IsDefault = false;
...
if (!simpleProtocol.IsSingleDosing)
{
    simpleProtocol.EndTimeParameter.IsDefault = false;
}

```
`simpleProtocol.EndTimeParameter.IsDefault` was already set to false, so the if case will not change anything.

msevestre: Oh I see. I'll double-check tomorrow. All this code is a one time thing anyways and will be deleted as soon as we have agreement on batch files

msevestre: Oh I see. I'll double-check tomorrow. All this code is a one time thing anyways and will be deleted as soon as we have agreement on batch files

msevestre: done

msevestre:
@Yuri05 Merged so that I can move forward. Further changes will be implemented in other PR

PR #635, opened 2018-04-29
Fixes #631 Cloned simulations corrupted in 7.3

### LINKED ISSUE #631: Cloned simulations corrupted in 7.3
Path to the solubility table is not properly cloned.
[Test4Clone.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/1954496/Test4Clone.zip)

E.g in the project attached: clone simulation **S1** into **S2** and try to configure S2: the error below comes.

clipboard02

msevestre:


src/PKSim.Core/Services/RenameAbsolutePathVisitor.cs
```diff
@@ -45,7 +45,9 @@ private bool isAbsolutePath(string pathAsString)
 
       private void renameAbsolutePathIn(IFormula formula)
       {
-         if (formula == null || !formula.IsExplicit()) return;
```
msevestre: this was only checking path for explicit formula. It should actually do it for all formula except constants

PR #638, opened 2018-05-01
Fixes #637 relative expressions not saved 

### LINKED ISSUE #637: Relative Expression not saved into snapshot

msevestre:


src/PKSim.CLI.Core/Services/SimulationExporter.cs
```diff
@@ -137,15 +132,14 @@ public Task Export(Simulation simulation, string outputFolder, SimulationExportM
 
       private async Task exportSimulationResultsToCsv(Simulation simulation, string outputFolder)
       {
-         var simulationResultFile = CoreConstants.DefaultResultsExportNameFor(simulation.Name);
-         var simulationResultFileFullPath = csvPathUnder(outputFolder, simulationResultFile);
+         var simulationResultFileFullPath = csvPathUnder(outputFolder, simulation.Name);
```
msevestre: This change to make sure that the csv file name is the same as before (otherwise comparison using your fav tool won't be possible :)

src/PKSim.CLI.Core/Services/SimulationExporter.cs
```diff
@@ -137,15 +132,14 @@ public Task Export(Simulation simulation, string outputFolder, SimulationExportM
...
          await _simulationExportTask.ExportResultsToCSVAsync(simulation, simulationResultFileFullPath);
          _logger.AddDebug($"Exporting simulation results to '{simulationResultFileFullPath}'");
       }
 
       private void exportParameters(string outputFolder, IndividualSimulation individualSimulation)
       {
-         var parameterReportFileName = csvPathUnder(outputFolder, $"{individualSimulation.Name}-parameters");
+         var parameterReportFileName = csvPathUnder(outputFolder, $"{individualSimulation.Name}_parameters");
```
msevestre: same with those guys. We used to export parameters with _parameters

src/PKSim.Core/Commands/SetRelativeExpressionCommand.cs
```diff
@@ -12,6 +12,13 @@ public SetRelativeExpressionCommand(IParameter parameter, double valueToSet)
          ObjectType = PKSimConstants.ObjectTypes.Molecule;
       }
 
+      protected override void UpdateParameter(IParameter parameter, IExecutionContext context)
+      {
+         base.UpdateParameter(parameter, context);
+         if (parameter == null) return;
+         parameter.IsDefault = (parameter.Value == 0);
```
msevestre: This is the big change here. We are doing a bunch of extra stuff for relative exprssions (normalizing etc..) . The easiest way to get a consistent behavio was to simply say that IsDefault means value is zero

src/PKSim.Core/Model/IndividualTransporterFactory.cs
```diff
@@ -47,6 +42,18 @@ public IndividualTransporter UndefinedLiverTransporterFor(Individual individual)
          return transporter;
       }
 
+      public IndividualMolecule CreateFor(ISimulationSubject simulationSubject, TransportType transporterType)
```
msevestre: Another bug that I found trying to compare old and new. old batch always created an efflux, no matter what. Here we will create the real process type

src/PKSim.Core/Snapshots/Mappers/ExpressionContainerMapper.cs
```diff
@@ -55,18 +51,7 @@ private void mapTransporterExpressionProperties(ExpressionContainer snapshot, Tr
          if (transporterExpressionContainer == null)
             return;
 
-         snapshot.MembraneLocation = SnapshotValueFor(transporterExpressionContainer.MembraneLocation, _defaultMembraneLocation);
-      }
-
-      private bool shouldMapContainer(IParameter expressionParameter, TransporterExpressionContainer transportedExpressionContainer)
```
msevestre: This was not accurate. We should always export the membrane location . But RelativeExpression IsDefault means nothing to export

src/PKSim.Core/Snapshots/Mappers/MoleculeMapper.cs
```diff
@@ -100,7 +100,12 @@ private void updateMoleculePropertiesToMolecule(IndividualMolecule molecule, Mol
 
       private IndividualMolecule createMoleculeFrom(Molecule molecule, ISimulationSubject simulationSubject)
       {
-         return factoryFor(molecule).CreateFor(simulationSubject);
+         var moleculeFactory = factoryFor(molecule);
+         var transporterFactory = moleculeFactory as IIndividualTransporterFactory;
+         if (transporterFactory == null || molecule.TransportType == null)
+            return moleculeFactory.CreateFor(simulationSubject);
+
+         return transporterFactory.CreateFor(simulationSubject, molecule.TransportType.Value);
```
msevestre: This is how I can call the transporterFactory with the REAL transportertype

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
```diff
@@ -181,7 +181,9 @@ private IEnumerable<IParameter> parametersChangedFromBuildingBlock(IGrouping<PKS
 
       private static bool parameterDiffersFromTemplate(IParameter templateBuildinglockParameter, IParameter parameter)
       {
-         return templateBuildinglockParameter == null || !ValueComparer.AreValuesEqual(parameter, templateBuildinglockParameter);
+         return templateBuildinglockParameter == null || 
+                !templateBuildinglockParameter.ValueIsDefined() ||
```
msevestre: This is a bug that was introduce when trying to not export too many paraemters. Sometimes, parameter in BB cannot be evaluated. Need to check that before comparing

src/PKSim.Presentation/Presenters/Individuals/IndividualMoleculePropertiesPresenter.cs
```diff
@@ -57,12 +57,12 @@ public void Edit(IndividualMolecule molecule, TSimulationSubject simulationSubje
          var parameters = new[] {molecule.ReferenceConcentration, molecule.HalfLifeLiver, molecule.HalfLifeIntestine};
          _moleculeParametersPresenter.Edit(parameters);
          _ontogenySelectionPresenter.Edit(molecule, simulationSubject);
-         _view.AdjustHeight();
       }
 
       public void RefreshView()
       {
          _ontogenySelectionPresenter.RefreshView();
+         _view.AdjustHeight();
```
msevestre: sometime, the view was squished. I need to refresh the view after it is visible

src/PKSim.Presentation/Presenters/Individuals/MoleculesPresenter.cs
```diff
@@ -184,6 +184,9 @@ public void ActivateNode(ITreeNode node)
          _activePresenter.MoleculeParametersVisible = _simulationSubject.IsAnImplementationOf<Individual>();
          _activePresenter.ActivateMolecule(moleculeFrom(node));
          _view.ActivateView(_activePresenter.BaseView);
+
+         //needs to be done onec the view has been activated to ensure propery display
+         _activePresenter.RefreshView();
```
msevestre: Which is here! Visible, refresh

msevestre:
@Yuri05 Need to merge this. Please review

src/PKSim.CLI.Core/Services/SimulationExporter.cs
Yuri05: :man_dancing:

src/PKSim.Core/Snapshots/Mappers/ExpressionContainerMapper.cs
```diff
@@ -92,7 +77,7 @@ private bool shouldMapContainer(IParameter expressionParameter, TransporterExpre
 
          var species = individual.Species.Name;
          var transporterExpressionContainer = expressionContainer.DowncastTo<TransporterExpressionContainer>();
-         var membraneLocation = ModelValueFor(snapshot.MembraneLocation, _defaultMembraneLocation);
+         var membraneLocation = ModelValueFor(snapshot.MembraneLocation, MembraneLocation.Basolateral);
```
Yuri05: can u explain this change?

msevestre: There is no member _defaultMembraneLocation. This member was avoid exporting the membrane location and reduced the number of entry. But the logic was not ok. Right now MembraneLocation will be exported as soon as the RelExp is !=0. I cannot enforce the presence of the member however because the same object is used for non transporter container
The default value for a contaienr in DB is in general MembraneLocation.Basolateral that;s why I am using it/
Alternatively, we could throw a warning if the location is not set. Find either way

Yuri05: ok. I thik we don't need a warning in this case.

PR #657, opened 2018-05-02
Fixes #648 load from snapshot quantity type

### LINKED ISSUE #648: Load from snapshot reports error and will not load
After loading one of the projects where "Save to shnapshot" works, I get the following error message:

Information: Loading Project from snapshot file 'M:\Test\Sorafenib\Sorafenib.json'
Warning: Snapshot file cannot be used to load a project.

Value "Drug, Observer" is not defined in enum. Path 'ParameterIdentifications[3].Analyses[1].DataRepositories[0].Columns[0].RelatedColumns[0].QuantityInfo.Type'.
Value "Drug, Observer" is not defined in enum. Path 'ParameterIdentifications[3].Analyses[1].DataRepositories[0].Columns[0].QuantityInfo.Type'.

For more information or questions, please visit the Open Systems Pharmacology Forum (forum.open-systems-pharmacology.org).

msevestre:


src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -222,16 +220,18 @@ private void addQualificationPlanToProject(ModelProject project, Model.Qualifica
 
       private Task<Model.Simulation[]> allSmulationsFrom(Simulation[] snapshotSimulations, ModelProject project) => _simulationMapper.MapToModels(snapshotSimulations, project);
 
-      private Task<IPKSimBuildingBlock[]> allBuidingBlocksFrom(SnapshotProject snapshot)
+      private async Task<IEnumerable<IPKSimBuildingBlock>> allBuidingBlocksFrom(SnapshotProject snapshot)
       {
-         var tasks = new List<Task<object>>();
```
msevestre: This was running all those tasks in //. Great until I realized that there is one instance of the mapper used per building block, which is good. But until you realized that loading many populations at the same time kinds of destroy other populations. That way, each building block is loaded one at at time. No issue with concurrent load

src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -249,6 +249,23 @@ private IEnumerable<Task<object>> mapSnapshotsToModels(IEnumerable<object> snaps
...
+      {
+         var models = new List<IPKSimBuildingBlock>();
+
+         if (snapshots == null)
+            return models;
+
+         //do not run tasks in parallel as the same mapper instance may be used concurrently to load two different snapshots
+         foreach (var snapshot in snapshots)
+         {
+            var buildingBlock = await snapshotMapper.MapToModel(snapshot);
```
msevestre: one at a time

src/PKSim.Core/Snapshots/Mappers/QuantityInfoMapper.cs
```diff
@@ -14,15 +15,16 @@ public override Task<SnapshotQuantityInfo> MapToSnapshot(ModelQuantityInfo quant
          {
             snapshot.OrderIndex = SnapshotValueFor(quantityInfo.OrderIndex);
             snapshot.Path = SnapshotValueFor(quantityInfo.PathAsString);
-            snapshot.Type = SnapshotValueFor(quantityInfo.Type, QuantityType.Undefined);
+            snapshot.Type = quantityInfo.Type != QuantityType.Undefined ? quantityInfo.Type.ToString() : null;
```
msevestre: Json Schema does nto support validation of Flag Enum (E.g. Drug|Observer) and would fail.
I am starting to wonder if we should not disable validation with schema especially considering the fact that those json files are created by PKSIm and are not coming from 3rd party tools

src/PKSim.Core/Snapshots/QuantityInfo.cs
```diff
@@ -1,11 +1,11 @@
-using OSPSuite.Core.Domain;
-
 namespace PKSim.Core.Snapshots
 {
    public class QuantityInfo : SnapshotBase
    {
       public string Path { set; get; }
-      public QuantityType? Type { set; get; }
+
+      //Cannot user QuantityType here as Snapshot validation will fail for composed Type such as "Drug, Observer"
```
Yuri05: cannot use (typo in comment)

src/PKSim.Core/Snapshots/Mappers/QuantityInfoMapper.cs
Yuri05: Unless we start e.g. exchanging compounds via snapshots...

src/PKSim.Core/Snapshots/QuantityInfo.cs
msevestre: fix in another PR

src/PKSim.Core/Snapshots/Mappers/QuantityInfoMapper.cs
Yuri05: >Or we could turn out schema validation for compound only when we are ready.

Sounds reasonable for me

PR #672, opened 2018-05-03
Fixes #616 Create Simmodel-XML from PKML in Matlab interface

### LINKED ISSUE #616: Create Simmodel-XML from PKML in Matlab interface
In PKSim.Matlab: add function, which takes a pkml file and saves it to SimModel-XML file.

msevestre:


msevestre:
There is one test failign that I will correct in next PR

PR #696, opened 2018-05-08
Fixes #687 Project created with version earler than 7.3: Warning when…

### LINKED ISSUE #687: Project created with version earler than 7.3: Warning when exporting to snapshot
If a project was created with version <7.3 and user selects "Export project to snapshot": a warning should be produced, explaining that a snapshot would contain a lot of garbage. After that the user could either continue with export or cancel.

See discussion in #674

msevestre:
… exporting to snapshot

### DIFF
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2166,7 +2166,15 @@ namespace PKSim.Assets
          public static readonly string WatermarkText = "Text";
          public static readonly string WatermarkProperties = "Watermark Properties";
          public static readonly string CreateTableSolubilityAlternative = "Create as pH-Solubility table";
-         public static readonly string DoYouWantToProceedWithExportToSnapshotWithChangedSimulation = "WARNING: Some simulations are in a changed state (red icon) and may not be re-imported correctly.\nDo you wish to continue?";
+
+         public static string DoYouWantToProceed(params string[] messages) => $"WARNING:\n{messages.ToString("\n")}\n\nDo you wish to continue?";
+
+         private static readonly string _snapshotOfProjectWithChangedSimulationText = "Some simulations are in a changed state (red icon) and may not be re-imported correctly.";
+         private static readonly string _snapshotOfProjectCreatedWithEarlierVersionText = "Project was created with an older version of PK-Sim (earlier than 7.3.0). Export to snapshot will create a file with a lot of noise.";
+
+         public static readonly string SnapshotOfProjectWithChangedSimulation = DoYouWantToProceed(_snapshotOfProjectWithChangedSimulationText);
+         public static readonly string SnapshotOfProjectCreatedWithEarlierVersion = DoYouWantToProceed(_snapshotOfProjectCreatedWithEarlierVersionText);
+         public static readonly string SnapshotOfProjectCreatedWithEarlierVersionAndWithChangedSimulation = DoYouWantToProceed(_snapshotOfProjectCreatedWithEarlierVersionText, _snapshotOfProjectWithChangedSimulationText);
 
          public static string NumberOfIndividualsToExtract(int count, string populationName) => $"{count} {"individual".PluralizeIf(count)} will be extracted from population {populationName}.";
 
--- a/src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
+++ b/src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
@@ -24,6 +24,7 @@ namespace PKSim.Core.Snapshots.Mappers
       private readonly QualificationPlanMapper _qualificationPlanMapper;
       private readonly IClassificationSnapshotTask _classificationSnapshotTask;
       private readonly ILazyLoadTask _lazyLoadTask;
+      private readonly ICreationMetaDataFactory _creationMetaDataFactory;
       private readonly ILogger _logger;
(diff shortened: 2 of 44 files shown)


src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -2166,7 +2166,15 @@ public static class UI
          public static readonly string WatermarkText = "Text";
          public static readonly string WatermarkProperties = "Watermark Properties";
          public static readonly string CreateTableSolubilityAlternative = "Create as pH-Solubility table";
-         public static readonly string DoYouWantToProceedWithExportToSnapshotWithChangedSimulation = "WARNING: Some simulations are in a changed state (red icon) and may not be re-imported correctly.\nDo you wish to continue?";
+
+         public static string DoYouWantToProceed(params string[] messages) => $"WARNING:\n{messages.ToString("\n")}\n\nDo you wish to continue?";
+
+         private static readonly string _snapshotOfProjectWithChangedSimulationText = "Some simulations are in a changed state (red icon) and may not be re-imported correctly.";
```
msevestre: These are the specific messages. Please review and create another issue if they need to be updated :)

src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -77,7 +80,13 @@ public class ProjectMapper : SnapshotMapperBase<ModelProject, SnapshotProject>
 
       public override async Task<ModelProject> MapToModel(SnapshotProject snapshot)
       {
-         var project = new ModelProject {Description = snapshot.Description};
+         var project = new ModelProject
```
msevestre: Creating project from Snapshot. We update the CreationMetaData to use the version that was saved in Snapshot

src/PKSim.Infrastructure/ProjectConverter/v7_3/Converter721To730.cs
```diff
@@ -48,6 +52,20 @@ public void Visit(Simulation simulation)
...
+      public void Visit(PKSimProject project)
+      {
+         convertProject(project);
+      }
+
+      private void convertProject(PKSimProject project)
+      {
+         //We do not know when the project was really created so we simply set the internal version to null
+         project.Creation = _creationMetaDataFactory.Create();
+         project.Creation.Version = string.Empty;
```
msevestre: I think this is a wise assumption here

src/PKSim.Presentation/Core/CreateProjectCommand.cs
```diff
@@ -10,30 +10,28 @@ namespace PKSim.Presentation.Core
...
-         _project = project;
          ObjectType = ObjectTypes.Project;
          CommandType = Command.CommandTypeAdd;
       }
 
       protected override void ExecuteWith(IExecutionContext context)
       {
-         _workspace.Project = _project;
+         var metaDataFactory = context.Resolve<ICreationMetaDataFactory>();
+         var project = new PKSimProject {Creation = metaDataFactory.Create()};
```
msevestre: Let's create the project here!!

src/PKSim.Presentation/Services/ProjectTask.cs
```diff
@@ -241,16 +241,30 @@ public void LoadProjectFromSnapshot()
...
-         {
-            var proceed = _dialogCreator.MessageBoxYesNo(PKSimConstants.UI.DoYouWantToProceedWithExportToSnapshotWithChangedSimulation);
-            if (proceed == ViewResult.No)
-               return Task.CompletedTask;
-         }
+         var project = _workspace.Project;
+         var anySimulationInChangedState = project.All<Simulation>().Any(x => _buildingBlockInSimulationManager.StatusFor(x) == BuildingBlockStatus.Red);
+         var projectExportWillCreateNoise = !_snapshotTask.IsVersionCompatibleWithSnapshotExport(project);
+
+         if (exitIf(anySimulationInChangedState && projectExportWillCreateNoise, PKSimConstants.UI.SnapshotOfProjectCreatedWithEarlierVersionAndWithChangedSimulation))
```
msevestre: I only want to show one message so I combine them depending on flags

src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -77,7 +80,13 @@ public class ProjectMapper : SnapshotMapperBase<ModelProject, SnapshotProject>
...
+         {
+            Description = snapshot.Description,
+            Creation = _creationMetaDataFactory.Create()
+         };
+         project.Creation.InternalVersion = snapshot.Version;
+         project.Creation.Version = ProjectVersions.FindBy(snapshot.Version)?.VersionDisplay;
```
Yuri05: Hard to say, if it's correct or not. 
On the one side, project is created with the **_current_** PK-Sim version and not with the **_snapshot_** version.
On the other side, if a snapshot was created with an older version of PK-Sim and contains all that noise, exporting a new snapshot from the recreated project will produce all the noise once again (and thus e.g. it would be good to get the export warning once again).

msevestre: That's the only way to track the original version however, unless we introduce yet another field for that

Yuri05: >That's the only way to track the original version however, unless we introduce yet another field for that

Yep, let's keep it as is for now

PR #702, opened 2018-05-09
Fixes #700 Import pkml simulation back to PK-Sim: relative expression…

### LINKED ISSUE #700: Import pkml simulation back to PK-Sim: relative expressions are gone
Created PK-Sim simulation using an individual with CYP3A4 relative expression enabled
Exported to Mobi
Added observer
reimport pkml to PK-Sim
relative expression node in simulation is available but empty
image

msevestre:
…s are gone

### DIFF
--- a/src/PKSim.Presentation/DTO/Mappers/ExpressionParametersToSimulationExpressionsDTOMapper.cs
+++ b/src/PKSim.Presentation/DTO/Mappers/ExpressionParametersToSimulationExpressionsDTOMapper.cs
@@ -76,7 +76,7 @@ namespace PKSim.Presentation.DTO.Mappers
          var expressionContainerDTO = new ExpressionContainerDTO();
          var moleculeName = relativeExpression.ParentContainer.Name;
          var simulation = _executionContext.Get<Simulation>(relativeExpression.Origin.SimulationId);
-         var molecule = simulation.Individual.MoleculeByName<IndividualMolecule>(moleculeName);
+         var molecule = simulation.Individual?.MoleculeByName<IndividualMolecule>(moleculeName);
          var isTransporter = moleculeIsTransporter(molecule);
 
          expressionContainerDTO.RelativeExpressionParameter = _containerParameterMapper.MapFrom(relativeExpression, expressionContainerDTO, x => x.RelativeExpression, x => x.RelativeExpressionParameter);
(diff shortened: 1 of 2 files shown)


src/PKSim.Presentation/DTO/Mappers/ExpressionParametersToSimulationExpressionsDTOMapper.cs
```diff
@@ -76,7 +76,7 @@ private ExpressionContainerDTO expressionContainerFor(IParameter relativeExpress
          var expressionContainerDTO = new ExpressionContainerDTO();
          var moleculeName = relativeExpression.ParentContainer.Name;
          var simulation = _executionContext.Get<Simulation>(relativeExpression.Origin.SimulationId);
-         var molecule = simulation.Individual.MoleculeByName<IndividualMolecule>(moleculeName);
+         var molecule = simulation.Individual?.MoleculeByName<IndividualMolecule>(moleculeName);
```
msevestre: Imported simulation do not have individual. hence the crash

src/PKSim.Presentation/DTO/Mappers/ExpressionParametersToSimulationExpressionsDTOMapper.cs
```diff
@@ -124,7 +124,7 @@ private ExpressionContainerDTO expressionContainerFor(IParameter relativeExpress
 
       private bool moleculeIsTransporter(IndividualMolecule molecule)
```
msevestre: There is no way to find (I believe) from the Global Molecule container in the model, that a container represents a transporter etc..

```
var container  = model.root.GetChildrenByName("CYP3A4")
container.MoleculeType == ENZYME <== that we do not have
```

At the moment, I am returning isTransporter false. But that would be better if we could actually retrieve the info from the container itself

tests/PKSim.Tests/IntegrationTests/ExpressionParametersToSimulationExpressionsDTOMapperSpecs.cs
```diff
@@ -97,4 +99,65 @@ public void should_map_one_parameter_for_each_realtive_expression_parameter_defi
...
+      private List<IParameter> _parameters;
+      private SimulationExpressionsDTO _result;
+      private IParameter _referenceConcentrationParam;
+      private IParameter _halfLifeLiverParameter;
+      private IParameter _halfLifeLiverIntestineParameter;
+
+      protected override void Context()
+      {
+         base.Context();
+         _simulation.RemoveAllBuildingBlockOfType(PKSimBuildingBlockType.Individual);
```
msevestre: Test without individual

PR #709, opened 2018-05-10
Fixes #708 Project Conversion

### LINKED ISSUE #708: Project Conversion
All parameter will get the flag IsDefault true except

* Parameter where IS_INPUT = true in DB
* Individual Parameter with Value!=DefaultVAlue
* Parmeter where IsFixedValue = true

msevestre:


src/PKSim.Infrastructure/ProjectConverter/v7_3/Converter721To730.cs
```diff
@@ -44,19 +53,44 @@ public Converter721To730(ICompoundFactory compoundFactory, ICloner cloner, ICrea
 
       public void Visit(Compound compound)
       {
+         resetAllParametersToIsDefaultIn(compound);
          convertCompound(compound);
       }
 
       public void Visit(Simulation simulation)
       {
+         simulation.AllBuildingBlocks<IPKSimBuildingBlock>().Each(resetAllParametersToIsDefaultIn);
```
msevestre: Converting all building blocks of the simulation as well as the simulation itself.

That being said, compound parameter IN the simulation (such as Lipo) will not show up as changed.
However newly created simulation with old compound will show the changes. I think this is good enough as there is NO way we can catch all parameters that have changed in the simulation anyways. Better show only the one that really changed in the simulation so that it's clearer

src/PKSim.Infrastructure/ProjectConverter/v7_3/Converter721To730.cs
```diff
@@ -44,19 +53,44 @@ public Converter721To730(ICompoundFactory compoundFactory, ICloner cloner, ICrea
...
+      {
+         resetAllParametersToIsDefault(container.GetAllChildren<IParameter>());
+      }
+
+      private void resetAllParametersToIsDefault(IEnumerable<IParameter> allParameters)
+      {
+         allParameters.Each(p =>
+         {
+            p.IsDefault = true;
+            if (p.ValueDiffersFromDefault())
```
msevestre: Checks for IsFixedValue AND for value!=DefaultValue

Yuri05: Where is the check for `IS_INPUT == true` done?

src/PKSim.Infrastructure/ProjectConverter/v7_3/Converter721To730.cs
```diff
@@ -44,19 +53,44 @@ public Converter721To730(ICompoundFactory compoundFactory, ICloner cloner, ICrea
...
+         resetAllParametersToIsDefaultIn(buildingBlock);
+      }
+
+      private void resetAllParametersToIsDefaultIn(IContainer container)
+      {
+         resetAllParametersToIsDefault(container.GetAllChildren<IParameter>());
+      }
+
+      private void resetAllParametersToIsDefault(IEnumerable<IParameter> allParameters)
+      {
```
Yuri05: would rename the function to smthg like `adjustDefaultStateOfAllParameters`
(because `resetAllParametersToIsDefault `lets me thinks that all parameters get `IsDefault=true`)

src/PKSim.Infrastructure/ProjectConverter/v7_3/Converter721To730.cs
msevestre: The check for IS_INPUT can only be done on a case by case basis. This is implemented for Compound where I have to load a template compound and compare.

src/PKSim.Infrastructure/ProjectConverter/v7_3/Converter721To730.cs
msevestre: Done

src/PKSim.Infrastructure/ProjectConverter/v7_3/Converter721To730.cs
Yuri05: Guess we will need this also for Formulation and Application Protocols

Yuri05:
Still has to be Converter72**2**To730.cs  ;)

PR #714, opened 2018-05-10
Fixes #711 Simulation: Save snapshot (dev only) visible in non-dev mode.

### LINKED ISSUE #711: Simulation: Save snapshot (dev only) visible in non-dev mode.
In the context menu of a simulation: **_Save snapshot (Developer only)_** action is visible in standard (non-developer) mode.

msevestre:


msevestre:
trivial

PR #716, opened 2018-05-11
Fixes #649 wrong default value in Scale Individual function

### LINKED ISSUE #649: wrong default value in "Scale Individual" function
when Scaling an individual, the "default value" is not conform with the current default value.
E.g. current Default value for liver half-life for CYP3A4 is 37h, when scaling the template individual, it says 72 h, see screenshot.

msevestre:


src/PKSim.Core/Services/IndividualExpressionsUpdater.cs
```diff
@@ -24,9 +27,20 @@ public void Update(Individual sourceIndividual, Individual targetIndividual)
          {
             var newMolecule = _cloner.Clone(molecule);
             targetIndividual.AddMolecule(newMolecule);
-            //we have to reset the new ontogeny for the molecule
+
+            //Make sure parameters that user defined parameters are reset to default to ensure proper scaling
+            resetMoleculeParametersToDefault(newMolecule);
```
msevestre: At the moment, molecule were cloned and added. When the user changed the default value, the "default value " of the target value was wrong. Parameter needs to be reset to default to see the scaling magic happen

PR #738, opened 2018-05-14
Fixes #723 Snapshot/Compound: Differences in Solubility

### LINKED ISSUE #723: Snapshot/Compound: Differences in Solubility
Steps to reproduce described in #720

Not sure, if it's really an issue here, but just to be complete: there are also some differences in solubility parameters shown.

In the compound there are 3 solubility alternatives: 
unbenannt

Comparison between original compound and compound reimported from snapshot shows following differences for the solubility: 

unbenannt

For the alternative **Sol3** (which is defined as table) `RefPh `and `GainPerCharge ` do not matter anyway.
Another difference in `RefPh `and `GainPerCharge ` (where empty Top Container is shown in comparison): no idea where it comes from. In the UI those values for the 2 other solubility alternatives are identical.

msevestre:


src/PKSim.Presentation/Services/CompoundAlternativeTask.cs
```diff
@@ -313,11 +313,12 @@ public TableFormula SolubilityTableForPh(ParameterAlternative solubilityAlternat
          var gainPerCharge = solubilityAlternative.Parameter(CoreConstants.Parameters.SOLUBILITY_GAIN_PER_CHARGE);
          var refSolubilityValue = refSolubility.Value;
 
+         var compoundForCalculation = _executionContext.Clone(compound);
```
msevestre: Creating a clone of the compound before using it to perform calculation. This ensure that we are not MUTATING the compound given as parameter (Template RefPh and Gain where changed, hence the difference when saving loading etc...) Very interesting

tests/PKSim.Tests/Presentation/CompoundAlternativeTaskSpecs.cs
```diff
@@ -341,20 +343,20 @@ protected override void Context()
...
-         _solubilityAlternative.Add(solubilityTable);
+
+         _compoundForCalculation = new Compound {refPh, solubilty, gainPerCharge, _solubility_pKa_pH_Factor};
+         _solubilityAlternative = new ParameterAlternative {refPh, solubilty, gainPerCharge, solubilityTable};
 
          var solubilityAlternativeGroup = new ParameterAlternativeGroup();
          solubilityAlternativeGroup.AddAlternative(_solubilityAlternative);
 
+
+         _originalRefPh = DomainHelperForSpecs.ConstantParameterWithValue(_refPhValue).WithName(CoreConstants.Parameters.REFERENCE_PH).WithDimension(DomainHelperForSpecs.NoDimension());
```
msevestre: This is the original parameter. Just testing that at the end of the test, it has the same value now because we are using a clone

PR #742, opened 2018-05-15
Fixes #733 Automatic switch of ValueOrigin to unknown

### LINKED ISSUE #733: Automatic switch of ValueOrigin to unknown
Value Origin should be automatically set to Unknown if:
* Parameter has state IsDefault and is overwritten by user
* Value Origin comes from PK-Sim DB (e.g. publication XY) and was not changed by user

msevestre:


src/PKSim.Core/Services/ParameterTask.cs
```diff
@@ -371,16 +371,7 @@ private ICommand withUpdatedDefaultStateAndValue(IOSPSuiteCommand executedComman
          macroCommand.Add(executedCommand);
          macroCommand.Add(new SetParameterDefaultStateCommand(parameter, isDefault: false) {ShouldChangeVersion = shouldChangeVersion}.Run(_executionContext).AsHidden());
 
-         if (!parameter.ValueOrigin.IsUndefined)
```
msevestre: This is the big code change here. Instead of only updating for undefined, we update all the time using the ValueOrigin.Unknown

tests/PKSim.Tests/Core/ParameterTaskSpecs.cs
```diff
@@ -670,9 +672,11 @@ public void should_set_the_default_flag_to_false()
       }
 
       [Observation]
-      public void should_not_set_the_value_origin_to_unknwon()
+      public void should_set_the_value_origin_to_unknwon()
       {
-         _parameter.ValueOrigin.Source.ShouldBeEqualTo(ValueOriginSources.Internet);
+         _parameter.ValueOrigin.Source.ShouldBeEqualTo(ValueOriginSources.Unknown);
+         _parameter.ValueOrigin.Method.ShouldBeEqualTo(ValueOriginDeterminationMethods.Undefined);
+         string.IsNullOrEmpty(_parameter.ValueOrigin.Description).ShouldBeTrue();
```
msevestre: Actual test

msevestre:
@Yuri05 Please review

PR #750, opened 2018-05-16
Fixes #747 and Fixes #749

### LINKED ISSUE #747: Load from Snapshot: usability
When loading project from snapshot: first the window "Load project from snapshot" is shown; after that the user has to click on "..." to open a file dialog with snapshot file selection.
The second action must be done always anyway, so I would suggest to open the file dialog first.

### LINKED ISSUE #749: Import from Snapshot: Cancel
If loading from snapshot is cancelled, both Building Blocks and Simulations windows are cleared, so that the user has to create a new project first.
aufnahme38

This is different from the usual behavior (e.g. if a project is closed, user can immediately create BBs/Simulations without explicitly creating a new project first)

msevestre:


msevestre:
@Yuri05 That should take care of those usability issues

PR #758, opened 2018-05-17
Fixes #754 individual enzyme refresh problem

### LINKED ISSUE #754: Individual enzymes: funny input field
After adding individual metabolizing enzymes unnecessary input field is shown

msevestre:


msevestre:
There is no tests for this presenter....that's why we did not caught the problem. I have no time to add the tests but something we should do in the future

PR #764, opened 2018-05-17
Fixes #763 protocol parameters should be variable in population simul…

### LINKED ISSUE #763: Add protocol parameters to Population user defined variability
That means that the flag has been set in MoBi. Parameter should be shown no matter what

msevestre:
…ation

### DIFF
--- a/src/PKSim.Core/Model/PopulationSimulation.cs
+++ b/src/PKSim.Core/Model/PopulationSimulation.cs
@@ -58,7 +58,8 @@ namespace PKSim.Core.Model
       /// <summary>
       ///    Returns all parameters that could potentially be defined as advanced parameters in a simulation 
       /// </summary>
-      public virtual IEnumerable<IParameter> AllPotentialAdvancedParameters => ParametersOfType(PKSimBuildingBlockType.Simulation | PKSimBuildingBlockType.Compound | PKSimBuildingBlockType.Event | PKSimBuildingBlockType.Formulation);
+      public virtual IEnumerable<IParameter> AllPotentialAdvancedParameters => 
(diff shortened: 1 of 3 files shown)


src/PKSim.Core/Model/PopulationSimulation.cs
```diff
@@ -58,7 +58,8 @@ public override TBuildingBlock BuildingBlock<TBuildingBlock>()
       /// <summary>
       ///    Returns all parameters that could potentially be defined as advanced parameters in a simulation 
       /// </summary>
-      public virtual IEnumerable<IParameter> AllPotentialAdvancedParameters => ParametersOfType(PKSimBuildingBlockType.Simulation | PKSimBuildingBlockType.Compound | PKSimBuildingBlockType.Event | PKSimBuildingBlockType.Formulation);
+      public virtual IEnumerable<IParameter> AllPotentialAdvancedParameters => 
+         ParametersOfType(PKSimBuildingBlockType.Simulation | PKSimBuildingBlockType.Compound | PKSimBuildingBlockType.Event | PKSimBuildingBlockType.Formulation | PKSimBuildingBlockType.Protocol);
```
msevestre: I added the protocol parameters. 
I think this absolutely makes sense. By default none of the parameter are CanBeVariedInPopulation so they don't show up anyways. But if you go to mobi and change the flag to true, they should appear there.

The next step would be to actually always set the dose for example as variable in pop... but this does not have to be done now

Yuri05: Unless somebody starts to vary application start times, should be ok

msevestre: Lol yeah. In that case, selber schuld

msevestre: Lol yeah. In that case, selber schuld

msevestre: Lol yeah. In that case, selber schuld

PR #772, opened 2018-05-18
Fixes #768 advanced protocol

### LINKED ISSUE #768: Advanced protocol cannot be imported from snapshot
Project created with 7.3.0.22

Trying to import from snapshot fails with:

```
Information: Loading Project from snapshot file 'C:\temp\AdvancedProtocol01.json'
Warning: Name 'Schema Item' is not unique in parent container 'Schema 1'
```

[01.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/2017503/01.zip)

msevestre:


src/PKSim.Core/Snapshots/Mappers/SchemaItemMapper.cs
```diff
@@ -30,7 +30,7 @@ public override Task<SnapshotSchemaItem> MapToSnapshot(ModelSchemaItem modelSche
       {
          var applicationType = ApplicationTypes.ByName(snapshot.ApplicationType);
          var schemaItem = _schemaItemFactory.Create(applicationType);
-         MapModelPropertiesToSnapshot(schemaItem, snapshot);
+         MapSnapshotPropertiesToModel(snapshot, schemaItem);
```
msevestre: Wrong method used.
Test worked because I was checking that name were equal.

msevestre:
@Yuri05 I also checked that MapModelPropertiesToSnapshot was never called in the MatToModel

PR #775, opened 2018-05-18
Fixes #771 and Fixes #773

### LINKED ISSUE #771: Old projects: applications not properly saved to snapshot
7.3.0.22
Open the project attached (created with 7.2.2) and save to snapshot.
Reload project from snapshot.
Dose and infusion time do not match
[7.2.2_AppTest.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/2017514/7.2.2_AppTest.zip)
clipboard01

Both parameters were not exported to snapshot even though they are both IS_INPUT=true and their values differ from defaults
```

  "Protocols": [
    {
      "Name": "App1",
      "ApplicationType": "Intravenous",
      "DosingInterval": "Single"
    }
  ],
```

### LINKED ISSUE #773: Snapshot for old projects: change warning text
Instead of `"... export to snapshot will create a file with a lot of noise"` should be something like
`"... exported snapshot might be buggy/incorrect. Use on your own risk"`

msevestre:


src/PKSim.Infrastructure/ProjectConverter/v7_3/Converter721To730.cs
```diff
@@ -92,6 +98,37 @@ public void Visit(Formulation formulation)
...
+
+      public void Visit(AdvancedProtocol protocol)
+      {
+         adjustDefaultStateOfAllParametersIn(protocol);
+         protocol.AllSchemas.Each(convertSchema);
+      }
+
+      private void convertSchema(Schema schema)
+      {
+         schema.NumberOfRepetitions.IsDefault = false;
```
msevestre: Those need to be changed by user anyways so not default no matter what

src/PKSim.Infrastructure/ProjectConverter/v7_3/Converter721To730.cs
```diff
@@ -152,7 +189,7 @@ private void updateIsInputStateByNameAndValue(IContainer containerToUpdate, ICon
          if (templateContainer == null)
             return;
 
-         foreach (var templateParameter in templateContainer.AllParameters(x => x.Visible && x.Editable && x.ValueIsComputable()))
```
msevestre: Some parameters are not editable such as dose. Not sure why and this should probably be changed
For the conversion, if parameters are visible and have changed, they should be Is_Default false. THis check was not required

PR #780, opened 2018-05-20
Fixes #776 Particles formulation monodisperse: parameters not hidden

### LINKED ISSUE #776: Particles formulation monodisperse: parameters not hidden
7.3
While creating particles dissolution formulation in **_monodisperse_** mode: parameters which should be visible only in **_polydisperse_** mode are not hidden

Must look like this:

(in 7.2.2 was OK)

msevestre:


src/PKSim.Presentation/Presenters/Formulations/FormulationSettingsPresenter.cs
```diff
@@ -152,9 +154,9 @@ private void adjustParameterVisibility()
          if (!Formulation.IsParticleDissolution)
             return;
 
-         var cache = new Cache<string, IParameter>(p => p.Name);
+         var cache = new Cache<string, IParameter>(p => p.Name, x => null);
          cache.AddRange(_formulationDTO.Parameters);
-         cache.Each(p => p.Visible = true);
```
msevestre: Hidding showing is not enough. We also need to update the IsDefault flag

msevestre:
@Yuri05 Merging. Please review

src/PKSim.Presentation/Presenters/Formulations/FormulationSettingsPresenter.cs
Yuri05: It's correct but a bit hard to understand what happens here.
It would be easier for understanding to show/hide parameters first and then for all parameters set 
`p.IsDefault=!p.Visible`

Yuri05: You can keep it as is, not a big deal. Might consider adding some comments explaining what's going on.
Maybe I have a bad day, but I had to read it 3 times to get it.

msevestre: No no I'll update. Nothing like the moment!-)

PR #781, opened 2018-05-20
Fixes #779 wrong dimension when creating new spec intestinal perm

### LINKED ISSUE #779: Wrong formula dimension in new specific intestinal permeability alternative
Problem and example described here: https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/722#issuecomment-390428276

msevestre:


src/PKSim.Core/Model/ParameterAlternativeFactory.cs
```diff
@@ -57,7 +58,7 @@ private void resetParameterToInput(IParameter parameter)
       {
          parameter.IsDefault = false;
          parameter.ValueOrigin.UpdateFrom(new ValueOrigin(), updateId:true);
-         parameter.Formula = _objectBaseFactory.Create<ConstantFormula>().WithValue(double.NaN);
+         parameter.Formula = _formulaFactory.ValueFor(double.NaN, parameter.Dimension);
```
msevestre: Instead of Nan, also sets the dimension to be the dimension of the parameter

msevestre:
@Yuri05 Please review

PR #793, opened 2018-05-23
Fixes #789 import from snapshot: advanced parameters differences

### LINKED ISSUE #789: Import from snapshot: differences in populations
With the example from #788:
* Open project and save population **Adults** as template
* Close project and load project from Snapshot
* Load population **Adults** from template and rename it to OLD_Adults
* Extract from both populations the same Individual (e.g. Individual 1) and compare both extracted Individuals **Adults-1** and **OLD_Adults-1**.

There is only one difference: **Lumen/LowerIleum/Length** (same happens if comparing other extracted individuals for these 2 populations)
aufnahme12

The reason for that is probably: **Lumen/LowerIleum/Length** was additionally distributed in the population **Adults** using **_Uniform_** distribution.

aufnahme11

Some other parameters were additionally distributed using, Normal, LogNormal and Constant distributions - all of them show no differences between original population and imported from snapshot.

msevestre:


src/PKSim.Core/Model/AdvancedParameter.cs
```diff
@@ -51,7 +51,7 @@ public int Seed
       /// <summary>
       ///    Generates a new random value based on the distribution type and the defined distribution parameters
       /// </summary>
-      public double GenerateRandomValue => DistributedParameter.RandomDeviateIn(_randomGenerator);
+      private double generateRandomValue => DistributedParameter.RandomDeviateIn(_randomGenerator);
```
msevestre: This should not be public. It it's ever called, the _randomGenerator will be skewed

src/PKSim.Core/Model/AdvancedParameter.cs
```diff
@@ -66,13 +66,13 @@ public double GetPercentileForValue(double value)
       ///    Generates an array of random values. The dimension of the array is equal to numberOfValues
       /// </summary>
       /// <param name="numberOfValues">Number of values to generate</param>
-      public virtual IEnumerable<RandomValue> GenerateRandomValues(int numberOfValues)
+      public virtual IReadOnlyList<RandomValue> GenerateRandomValues(int numberOfValues)
       {
          ResetGenerator();
```
msevestre: I am not sure if this makes sense or not. That means that a given adavned parameter will always create the same series of random values...
This makes however our use case possible

src/PKSim.Core/Model/AdvancedParameterFactory.cs
```diff
@@ -56,16 +61,14 @@ public AdvancedParameter Create(IParameter parameter, DistributionType distribut
...
-         distributionMetaData.DefaultUnit = parameter.DisplayUnit.Name;
 
-         distributionMetaData.ParameterName = "distribution";
          advancedParameter.DistributedParameter = _parameterFactory.CreateFor(distributionMetaData);
 
          //should set the distribution parameter as visible since they will need to be edited
          advancedParameter.DistributedParameter.AllParameters().Each(x =>
          {
             x.Visible = true;
+            x.IsDefault = false;
```
msevestre: This was probably missing here too. Those parameter are definitly input

src/PKSim.Core/Snapshots/AdvancedParameter.cs
```diff
@@ -2,6 +2,7 @@
 {
    public class AdvancedParameter : ParameterContainerSnapshotBase
    {
+      public int Seed { get; set; }
```
msevestre: See. Not saved to snapshot :)

tests/PKSim.Tests/Core/AdvancedParameterSpecs.cs
```diff
@@ -0,0 +1,39 @@
...
+   }
+
+   public class When_generationg_some_random_values_consecutively : concern_for_AdvancedParameter
+   {
+      private IReadOnlyList<double> _values1;
+      private IReadOnlyList<double> _values2;
+
+      protected override void Because()
+      {
+         _values1 = sut.GenerateRandomValues(10).Select(x => x.Value).ToList();
```
msevestre: I added this test just to convince myself that this was the behavior we actually had

tests/PKSim.Tests/Core/AdvancedParameterSpecs.cs
```diff
@@ -0,0 +1,39 @@
...
+         _values2 = sut.GenerateRandomValues(10).Select(x => x.Value).ToList();
+      }
+
+      [Observation]
+      public void should_return_the_same_values()
+      {
+         _values1.ShouldBeEqualTo(_values2);
```
msevestre: and indeed, the values are the same :)

src/PKSim.Core/Snapshots/AdvancedParameter.cs
Yuri05: To be discussed. Will create an issue for 7.4 for this.

src/PKSim.Core/Model/AdvancedParameterFactory.cs
Yuri05: Yep

src/PKSim.Core/Model/AdvancedParameter.cs
msevestre: If we do not reset, there is NO way we can recreate the same population parameters.
I think this is absolutely ok when you think about it, because EACH advanced parameter will have a different seed to they are independent

msevestre: and remove the advanced parameter and added it again will create another seed

PR #809, opened 2018-05-25
Fixes #808 Load observed data from snapshot should be for dev only

### LINKED ISSUE #808: Load observed data from snapshot should be for dev only

msevestre:


msevestre:
@Yuri05 FYI

Yuri05:
For the future: I think it would make sense to enable it also for non-developers (same as for compounds)

PR #810, opened 2018-05-25
Fixes #806 difference table

### LINKED ISSUE #806: Differences between snapshot-loaded simulation and identical working journal-loaded simulation 3: comparison function arguable
Similar to #803, I have built a simulation in v 7.3.0 with an adult scaled down to a preterm and a large molecule. I have exported it to a snapshot and to working journal. Then, I have opened a new project and loaded the snapshot-simulation and the working journal simulation (in the attached file called S2_Preterm_Protein and OLD_S2_Preterm_Protein, respectively).
When comparing both simulations, a breadth of differences is shown, some probably redundant or similar compared to #803. In addition to those shown in #803 and #805 the comparison function also yields differences for the property Y, which seems arguable, at least from the number of decimal point shown here...

image

Files are attached in #805

msevestre:


src/PKSim.Presentation/Presenters/Snapshots/LoadSimulationFromSnapshotPresenter.cs
```diff
@@ -0,0 +1,59 @@
...
+   public interface ILoadSimulationFromSnapshotPresenter : ILoadFromSnapshotPresenter<Simulation>
+   {
+      /// <summary>
+      ///    Simulation loaded from selected snapshot file. It is null if the user cancels the action or if the file was not
+      ///    loaded properly
+      /// </summary>
+      Simulation LoadSimulation();
+   }
+
+   public class LoadSimulationFromSnapshotPresenter : LoadFromSnapshotPresenter<Simulation>, ILoadSimulationFromSnapshotPresenter
```
msevestre: Also created an overload for Simulation so that we have the same UI even if it's only for debug

PR #811, opened 2018-05-25
Fixes #796 update display unit for concentration parameter created on…

### LINKED ISSUE #796: Cloning a simulation destroys Persistable-property
Once a simulation was run in PK-Sim: cloning this simulation will set a lot of quantities as persistable.
E.g. in project attached: clone simulation **S1** to **S1_clone** and compare both simulations
[CloneTest.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/2036413/CloneTest.zip)

aufnahme1

It happens only after original simulation was run. If cloned **before** the first run: all Persistable flags are cloned properly.

In the UI the differences cannot be seen directly
E.g. if **Define Settings and Run** dialog is opened in the cloned simulation: only the quantities which were selected for output in the _original_ simulation are also selected.

However exporting SimModel-XMLs for both simulations will also show the differences in persistable flag
(**which will result in much higher memory consumption of the cloned simulation during calculation, especially for population simulations and PIs if run in parallel on multiple cores**!)

[XMLs.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/2036437/XMLs.zip)

In the example above following quantities were set to persistable after cloning :
* ALL Observers (with 3 or 4 exceptions, hmm)
* 3 molecule amounts: **Gallbladder**, **Lumen|Feces** and **Organism|Kidney|Urine**

 (happens in 7.2 and 7.3)

msevestre:
… the fly

### DIFF
--- a/src/PKSim.Core/Model/ParameterFactory.cs
+++ b/src/PKSim.Core/Model/ParameterFactory.cs
@@ -224,6 +224,7 @@ namespace PKSim.Core.Model
          parameter.Info.Visible = false;
          parameter.Formula = _formulaFactory.ConcentrationFormulaFor(formulaCache);
          parameter.Dimension = _dimensionRepository.MolarConcentration;
+         parameter.DisplayUnit = _displayUnitRetriever.PreferredUnitFor(parameter);
          return parameter;
       }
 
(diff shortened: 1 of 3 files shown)


src/PKSim.Core/Model/ParameterFactory.cs
```diff
@@ -224,6 +224,7 @@ public IParameter CreateConcentrationParameterIn(IFormulaCache formulaCache)
          parameter.Info.Visible = false;
          parameter.Formula = _formulaFactory.ConcentrationFormulaFor(formulaCache);
          parameter.Dimension = _dimensionRepository.MolarConcentration;
+         parameter.DisplayUnit = _displayUnitRetriever.PreferredUnitFor(parameter);
```
msevestre: At least fixes the dimension issue for concentration parameter

PR #820, opened 2018-05-29
Fixes #798 Population simulation window: Name of statistics for displ…

### LINKED ISSUE #798: Population simulation window: Name of statistics for display not properly shown after (de)selection
In the Time Profile Dialog of Run Population Simulation, Tab Output --> Output: Select distribution statistics for display, the name of the different statistics changes when being ticked or unticked as shown below:

image

image

image

(Of note, the correct statistics are subsequently calculated, despite the wrong display here)
PK-Sim file:
[Test.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/2038544/Test.zip)

msevestre:
…ay not properly shown after (de)selection

### DIFF
--- a/src/PKSim.UI/Views/PopulationAnalyses/PopulationAnalysisStatisticsSelectionView.cs
+++ b/src/PKSim.UI/Views/PopulationAnalyses/PopulationAnalysisStatisticsSelectionView.cs
@@ -1,17 +1,17 @@
 ﻿using System.Collections.Generic;
-using OSPSuite.DataBinding.DevExpress;
-using OSPSuite.DataBinding.DevExpress.XtraGrid;
-using OSPSuite.UI.Extensions;
-using OSPSuite.UI.RepositoryItems;
-using OSPSuite.Assets;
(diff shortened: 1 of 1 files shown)


src/PKSim.UI/Views/PopulationAnalyses/PopulationAnalysisStatisticsSelectionView.cs
```diff
@@ -56,11 +54,10 @@ public override void InitializeBinding()
          _gridViewBinder.Changed += NotifyViewChanged;
       }
 
-      private RepositoryItem editRepository(StatisticalAggregation statisticalAggregation)
+      private RepositoryItem editRepository(StatisticalAggregation statisticalAggregation) => new RepositoryItemTextEdit
```
msevestre: This is change in DEvExpress a few version. Before we could use the same instance and change property of the actual instance. (which was kind of weird). Now we need to create a new instance

msevestre:
@Yuri05 Trivial. Merging

PR #821, opened 2018-05-29
Fixes #795 Renaming Time Profile Analysis tab only possible after clo…

### LINKED ISSUE #795: Renaming Time Profile Analysis tab only possible after closing and re-openening simulation window
In the attached file, I renamed the Time Profile Analysis 2 to "Plot 2". 
The new name is only shown after the simulation window is closed and re-opened, otherwise the old name is shown as in the figure below

image

[Test.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/2036244/Test.zip)

msevestre:
…sing and re-openening simulation window

### DIFF
--- a/src/PKSim.Presentation/Presenters/Charts/SimulationTimeProfileChartPresenter.cs
+++ b/src/PKSim.Presentation/Presenters/Charts/SimulationTimeProfileChartPresenter.cs
@@ -31,9 +31,9 @@ namespace PKSim.Presentation.Presenters.Charts
    }
 
    public class SimulationTimeProfileChartPresenter : ChartPresenter<SimulationTimeProfileChart, ISimulationTimeProfileChartView, ISimulationTimeProfileChartPresenter>,
-      ISimulationAnalysisPresenter<IndividualSimulation>,
-      ISimulationTimeProfileChartPresenter
-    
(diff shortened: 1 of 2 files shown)


src/PKSim.Presentation/Presenters/Charts/SimulationTimeProfileChartPresenter.cs
```diff
@@ -23,15 +23,17 @@ namespace PKSim.Presentation.Presenters.Charts
    public interface ISimulationTimeProfileChartPresenter : IChartPresenter<SimulationTimeProfileChart>,
       ISimulationAnalysisPresenter,
       IPKAnalysisWithChartPresenter,
+      IListener<RenamedEvent>,
```
msevestre: Added this event here to listen to global rename event

msevestre:
@Yuri05  Please review

src/PKSim.Presentation/Presenters/Charts/SimulationTimeProfileChartPresenter.cs
Yuri05: What about other profile types (Range Plots in Pop etc.)? Or was Time profile the only one where RenamedEvent was missing?

PR #836, opened 2018-06-04
Fixes #835 crash undo simulation

### LINKED ISSUE #835: Undo of configure simulation throws an exception
This is only in 7.3 after redesign of SerializationContext

msevestre:


msevestre:
@Yuri05 Bug found during ranorex tests. SerializationContext was refactored with a simpler API in 7.3. This use case was not taken into consideration (e.g. two simulations with the same id, but different instances). This happens only when deserializing a simulation that is already loaded with the same id (e.g. of configure)

src/PKSim.Infrastructure/Serialization/Xml/XmlSerializationManager.cs
```diff
@@ -55,7 +55,10 @@ public TObject Deserialize<TObject>(byte[] serializationBytes, SerializationCont
 
          var (element, originalVersion, conversionHappened) = getConvertedElementFrom(serializationBytes);
 
-         var context = serializationContext ?? _serializationContextFactory.Create();
+         var isSimulation = typeof(TObject).IsAnImplementationOf<Simulation>();
```
msevestre: When deserialzing simulations FROM byte array, do not add project simulations

src/PKSim.UI/Views/Compounds/CompoundTypeGroupView.cs
```diff
@@ -141,6 +141,6 @@ public void Repaint()
          gridView.LayoutChanged();
       }
 
-      public int OptimalHeight => layoutControlGroup.Height;
+      public int OptimalHeight => layoutControlGroup.Height + layoutControl.Margin.All;
```
msevestre: This was also a bug I saw during ranorex tests

src/PKSim.UI/Views/Compounds/MolWeightGroupView.cs
```diff
@@ -190,7 +190,7 @@ public void Repaint()
          _gridView.LayoutChanged();
       }
 
-      public int OptimalHeight => layoutControlGroup.Height;
+      public int OptimalHeight => layoutControlGroup.Height + layoutControl.Margin.All;
```
msevestre: This was also a bug I saw during ranorex tests.

msevestre:
@Yuri05 Merging. Please review

PR #842, opened 2018-06-06
 Fixes #841 PKSimDB location 

### LINKED ISSUE #841: Setup: PK-Sim DB
7.3.0.39
**PKSimDB.mdb** is installed both under 
`C:\Program Files (x86)\Open Systems Pharmacology\PK-Sim 7.3`
and
`C:\ProgramData\Open Systems Pharmacology\PK-Sim\7.3`
Must be only ProgramData

msevestre:


rakefile.rb
```diff
@@ -28,8 +29,6 @@
 		'packages/**/OSPSuite.TeXReporting/**/*.*',
 		'examples/**/*.{wxs,pksim5}',
 		'src/PKSim.Assets/Resources/*.ico',
-		'src/Db/PKSimDB.mdb',
```
msevestre: They do not need to be added explicitly as those guys are copied automatically

rakefile.rb
```diff
@@ -19,6 +19,7 @@
 	#Ignore files from automatic harvesting that will be installed specifically
 	harvest_ignored_files = [
 		'PKSim.exe',
+		'PKSimDB.mdb',
```
msevestre: Should be ignore from automatic harvesting however

PR #898, opened 2018-07-26
Fixes #897 missing tramsporters

### LINKED ISSUE #897: 7.4 Missing DrugRelease-Transports
[Human_SingleORAL_Weibull_comparison_7.3_vs_7.4.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/2230880/Human_SingleORAL_Weibull_comparison_7.3_vs_7.4.zip)

Comparing simulations for `Human_SingleORAL_Weibull.json` between 7.3 and 7.4: the only difference is that all DrugRelease-Transports (Tablet==>Lumen) are missing

aufnahme42

msevestre:


src/PKSim.Infrastructure/ORM/DAS/DASDataRow.cs
```diff
@@ -261,60 +261,6 @@ public void SelectFromDB()
          if (RowState != DataRowState.Detached && RowState != DataRowState.Added)
             AcceptChanges();
       }
-
-      /// <summary>
```
msevestre: Most of the DAS Stuff is not required. SLowly but surely removing what we don't need. Goal is to not use this abstraction anymore

src/PKSim.Infrastructure/ORM/DAS/DAS.cs
```diff
@@ -736,36 +557,27 @@ public void FillDataTable(DASDataTable dataTable, string sql)
...
          dataTable.ParameterValues = paramValues;
 
          dataTable.SQL = sql;
          dataTable.Rows.Clear();
-
-         switchNamedParametersToUnnamed(ref cmd);
          try
          {
-            dataTable.Load(cmd.ExecuteReader());
+            using (var adapter = _providerFactory.CreateDataAdapter())
```
msevestre: This is the actual change! Adapter instead of reader
BTW: We have leaks everywhere here. Reader etc.. are never disposed... what a mess

PR #905, opened 2018-08-01
Fixes #904 content data not deleted

### LINKED ISSUE #904: Content data not deleted in project
When deleting building blocks / simulations from a project, corresponding data is not removed from the `CONTENTS `database table. As a result, PK-Sim project files do not shrink when removing simulations / BBs.

msevestre:


src/PKSim.Infrastructure/Serialization/WorkspaceLayoutPersistor.cs
```diff
@@ -17,29 +17,35 @@ public class WorkspaceLayoutPersistor : IWorkspaceLayoutPersistor
...
          _workspaceLayoutMapper = workspaceLayoutMapper;
       }
 
       public void Save(IWorkspaceLayout workspaceLayout, ISession session)
       {
-
-         //first remove old layout if existing
-         var layoutsFromDb = session.CreateCriteria<WorkspaceLayoutMetaData>().List<WorkspaceLayoutMetaData>();
-         layoutsFromDb.Each(session.Delete);
+         var layoutFromDb = loadLayoutFromDb(session);
```
msevestre: I realized that the content was always deleted and added. A little stupid and also annoying when looking at what is changing when saving a project. That way, we update the existing layout if there is already one instead of saving it again

PR #962, opened 2018-10-04
Fixes #948 scale individual resets concentration

### LINKED ISSUE #948: Scale Individual: Manually entered reference concentration is reset
If I had an individual with metabolizing enzymes, and if I override the reference concentrations to values found in the literature or identified by PI, these manually set values will be reset to default when "Scale" is applied. However, the "Value Origin" field is left to the reference.

Since, to my understanding, there is an ontogeny factor which will do the scaling, I suggest to leave manually entered values unchanged in the scaling process.

msevestre:


src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -35,20 +35,23 @@ public static bool IsExpression(this IParameter parameter)
...
-      public static bool IsIndividualMolecule(this IParameter parameter)
+      public static bool IsExpressionOrOntogenyFactor(this IParameter parameter)
       {
          if (parameter.IsExpression())
             return true;
 
          if (CoreConstants.Parameters.OntogenyFactors.Contains(parameter.Name))
             return true;
 
-         if (parameter.IsNamed(CoreConstants.Parameters.REFERENCE_CONCENTRATION))
```
msevestre: I think the intent was right but the usage was wrong.
I have created two method `IsExpressionOrOntogenyFactor` and `IsIndividualMolecule`

The former `IsIndiviudalMolecule` implementation was I believe not OK

src/PKSim.Core/Model/Population.cs
```diff
@@ -228,10 +228,7 @@ public virtual IEnumerable<IParameter> AllConstantParameters(IEntityPathResolver
 
       private bool parameterShouldBeReturnedAsConstant(IParameter parameter)
       {
-         if (parameter.IsNamed(CoreConstants.Parameters.REFERENCE_CONCENTRATION))
```
msevestre: Here we were bypassing the implementation because of the wrong implemenation

src/PKSim.Core/Services/IndividualScalingTask.cs
```diff
@@ -43,7 +43,7 @@ where scalingRequiredFor(originParameter.Value, targetParameter)
 
       private bool scalingRequiredFor(IParameter originParameter, IParameter targetParameter)
       {
-         if (originParameter == null || targetParameter == null)
```
msevestre: cannot be null by construction

tests/PKSim.Tests/Core/IndividualScalingTaskSpecs.cs
```diff
@@ -122,9 +133,9 @@ public void should_leverage_the_scaling_method_task_to_retrieve_the_default_scal
       }
 
       [Observation]
-      public void should_not_return_the_reference_concentration_parameter()
```
msevestre: Interesting that we were EXPLICITELY excluding ref concentration from scaling...not sure why. Probably because someone said it did not make sense? Well now other someones want to have it :)

src/PKSim.Core/Model/Population.cs
msevestre: I checked all usage of previous `IsIndividualMolecule`. It will know perform ok when used to genereate Individual pdf (it was not the case before)

msevestre:
@Yuri05 Merging as the next pr is coming. Please review

PR #963, opened 2018-10-04
Fixes #951 change method to use source value for global molecule para…

### LINKED ISSUE #951: Scale Individual: Change method default to "Use Source Value"
Branch-off from issue #948: In the "Scaling Configuration" tab, the default "Scaling Method" should be changed to "Use Source Value". See discussion in #948.

msevestre:
…meters

### DIFF
--- a/src/PKSim.Core/CoreConstants.cs
+++ b/src/PKSim.Core/CoreConstants.cs
@@ -847,6 +847,14 @@ namespace PKSim.Core
          public static readonly string REL_EXP_VASC_ENDO = "Relative expression in vascular endothelium";
          public static readonly string REL_EXP_VASC_ENDO_NORM = NormParameterFor(REL_EXP_VASC_ENDO);
 
+         public static IReadOnlyList<string> AllGlobalMoleculeParameters = new[]
+         {
+            REFERENCE_CONCENTRATION,
+            HALF_LIFE_LIVER,
+            HALF_LIFE_INTESTINE
+         };
+
+
          public static IReadOnlyList<string> AllGlobalRelExpParameters = new[]
          {
             REL_EXP_BLOOD_CELL,
--- a/src/PKSim.Core/Model/ParameterExtensions.cs
+++ b/src/PKSim.Core/Model/ParameterExtensions.cs
@@ -48,9 +48,10 @@ namespace PKSim.Core.Model
 
       public static bool IsIndividualMolecule(this IParameter parameter)
       {
-         return IsExpressionOrOntogenyFactor(parameter) ||
-                parameter.NameIsOneOf(CoreConstants.Parameters.REFERENCE_CONCENTRATION, CoreConstants.Parameters.HALF_LIFE_LIVER, CoreConstants.Parameters.HALF_LIFE_INTESTINE);
+         return IsExpressionOrOntogenyFactor(parameter) || IsIndividualMoleculeGlobal(parameter);
       }
+
+      public static bool IsIndividualMoleculeGlobal(this IParameter parameter) => CoreConstants.Parameters.AllGlobalMoleculeParameters.Contains(parameter.Name);
 
       public static bool IsStructural(this IParameter parameter)
       {
--- a/src/PKSim.Core/Services/OverrideScalingMethodSpecification.cs
+++ b/src/PKSim.Core/Services/OverrideScalingMethodSpecification.cs
@@ -1,3 +1,6 @@
+using System.Linq;
+using PKSim.Core.Model;
+
 namespace PKSim.Core.Services
 {
    public class OverrideScalingMethodSpecification : IScalingMethodSpecification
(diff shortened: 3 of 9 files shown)


src/PKSim.Core/CoreConstants.cs
```diff
@@ -847,6 +847,14 @@ public static class Parameters
          public static readonly string REL_EXP_VASC_ENDO = "Relative expression in vascular endothelium";
          public static readonly string REL_EXP_VASC_ENDO_NORM = NormParameterFor(REL_EXP_VASC_ENDO);
 
+         public static IReadOnlyList<string> AllGlobalMoleculeParameters = new[]
```
msevestre: It's about time we have one central place to hold those parameters

src/PKSim.Core/Services/OverrideScalingMethodSpecification.cs
```diff
@@ -15,15 +18,11 @@ public bool IsSatisfiedBy(ParameterScaling parameterScaling)
          return true;
       }
 
-      public bool IsDefaultFor(ParameterScaling parameterSatifyingSpecification)
+      public bool IsDefaultFor(ParameterScaling parameterScaling)
       {
-         //never the default value
-         return false;
+         return parameterScaling.SourceParameter.IsIndividualMoleculeGlobal();
```
msevestre: It is now the default for those parameters

tests/PKSim.Tests/Core/OverrideScalingMethodSpecificationSpecs.cs
```diff
@@ -68,5 +74,11 @@ public void should_return_false_for_a_distributed_parameter()
       {
          sut.IsDefaultFor(_distributedScaling).ShouldBeFalse();
       }
+
+      [Observation]
+      public void should_return_true_for_a_global_molecule_parameter()
+      {
+         sut.IsDefaultFor(_globalMoleculeParameterScaling).ShouldBeTrue();
```
msevestre: testing here

msevestre:
@Yuri05 please review

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -48,10 +48,11 @@ public static bool IsExpressionOrOntogenyFactor(this IParameter parameter)
 
       public static bool IsIndividualMolecule(this IParameter parameter)
```
Yuri05: Should be renamed to ```IsIndividualMoleculeParameter```

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -48,10 +48,11 @@ public static bool IsExpressionOrOntogenyFactor(this IParameter parameter)
...
       {
-         return IsExpressionOrOntogenyFactor(parameter) ||
-                parameter.NameIsOneOf(CoreConstants.Parameters.REFERENCE_CONCENTRATION, CoreConstants.Parameters.HALF_LIFE_LIVER, CoreConstants.Parameters.HALF_LIFE_INTESTINE);
+         return IsExpressionOrOntogenyFactor(parameter) || IsIndividualMoleculeGlobal(parameter);
       }
 
+      public static bool IsIndividualMoleculeGlobal(this IParameter parameter) => CoreConstants.Parameters.AllGlobalMoleculeParameters.Contains(parameter.Name);
```
Yuri05: Should be renamed to ```IsIndividualMoleculeGlobalParameter```

src/PKSim.Core/Model/ParameterExtensions.cs
msevestre: Disagree. Those method are extension method and read as 

parameter.IsIndividualMolecule()


Adding parameter at the end is 1/ redundant 2/ not in accordance with other methods in the same file such as IsExpression etc..


I would simply leave it as is

src/PKSim.Core/Model/ParameterExtensions.cs
msevestre: Same explanation as above

src/PKSim.Core/Services/OverrideScalingMethodSpecification.cs
Yuri05: Seems to be wrong for me!
Default method for those parameters should be "Use Source Value", which corresponds to `KeepScalingMethodSpecification` (?) and not to `OverrideScalingMethodSpecification`

src/PKSim.Core/Model/ParameterExtensions.cs
msevestre: +1 :)

src/PKSim.Core/Services/OverrideScalingMethodSpecification.cs
msevestre: No you are seeing it backward:
We scale from source to target .From the target point of view, we can either KEEP the value (Use default) or OVERRIDE the value (USe source)

msevestre: This was the old nomenclature. We could probalby rename those classes to match what the UI is showing to avoid confusion.

Yuri05: Yep, renaming would be a good idea

PR #964, opened 2018-10-04
Fixes #952 use source value origin from source when using update from…

### LINKED ISSUE #952: Scale Individual: Select proper "Value Origin" field dependent on change method
Another branch-off from issue #948: The Value Origin field content (see tab "Expression") should be set dependent on the selection done in "Scaling Configuration". If the Source Value is selected as Scaling Method, its Value Origin should be transferred to the scaled individual. If the Default Value is selected, the corresponding default value origin entry should be used - currently this field is usually empty, but it may change in the future.

msevestre:
… source

### DIFF
--- a/src/PKSim.Core/Services/OverrideScalingMethod.cs
+++ b/src/PKSim.Core/Services/OverrideScalingMethod.cs
@@ -1,5 +1,6 @@
-using PKSim.Assets;
 using OSPSuite.Core.Commands.Core;
+using PKSim.Assets;
+using PKSim.Core.Commands;
 
 namespace PKSim.Core.Services
 {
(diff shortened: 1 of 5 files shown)


src/PKSim.Core/Services/OverrideScalingMethod.cs
```diff
@@ -19,7 +20,12 @@ public override double ScaledValueFor(ParameterScaling parameterScaling)
 
       protected override ICommand PerformScaling(ParameterScaling parameterScaling)
       {
-         return _parameterTask.SetParameterValue(parameterScaling.TargetParameter, parameterScaling.SourceParameter.Value);
+         var (sourceParameter, targetParameter) = parameterScaling;
```
msevestre: This is quite neat. Using parameter desctructure feature from C#7.

src/PKSim.Core/Services/OverrideScalingMethod.cs
```diff
@@ -19,7 +20,12 @@ public override double ScaledValueFor(ParameterScaling parameterScaling)
...
+
+         var updateValueCommand = _parameterTask.SetParameterValue(targetParameter, sourceParameter.Value);
+         var updateValueOriginCommand = _parameterTask.SetParameterValueOrigin(targetParameter, sourceParameter.ValueOrigin);
```
msevestre: This is where I am also updating the value origin. It was actually already working for the one use case of molecule because were were cloning the whole molecule

src/PKSim.Core/Services/ParameterScaling.cs
```diff
@@ -114,5 +94,11 @@ public void ResetTargetParameter()
       {
          TargetParameter.Value = _targetValueBeforeScaling;
       }
+
+      public void Deconstruct(out IParameter sourceParameter, out IParameter targetParameter)
```
msevestre: This is how you define a deconstruct

msevestre:
@Yuri05 Please review

Issue #941, opened 2018-09-13
Manual editing of Observed Data will not accept system decimal separator


UlrichSi:
Same as #902: Only decimal point is accepted here, not separator set by the Windows system.

Issue #902, opened 2018-07-31
Parameter Identification Charts: Axis Options will not accept system decimal separator


UlrichSi:
If I enter a value into the axis scaling fields of a PI chart, I can only use a dot as decimal separator, but not the comma, as defined in the nationalized setting. Entering a comma is displayed but ignored. This behavior is different from all other value fields in PK-Sim where both will work.

![grafik](https://user-images.githubusercontent.com/27352743/43443395-07c0fdce-94a1-11e8-8ffe-c699adca9c87.png)

PavelBal:
Same for MoBi - should we open a new issue?

msevestre:
This is probably due to the fact that Min Max can also be empty. I would not be surprised if this is also the case for Weight in PI.

The code is the same for both PKSim and MoBi so we will fix all issues at the same time

UlrichSi:
Sure enough, using a comma in PI weight in the Data tab, lower right part (for the individual data points) will also be ignored, i.e., 1,5 becomes 15. But it works fine in the upper right part (for the total weight of the outputs). 

@msevestre Since you mentioned that weight might be empty: the behavior of PK-Sim is a bit obnoxious if you really delete the value (usually 1) and try to leave it empty. Two exception errors are thrown, with the first displaying a for a typical user meaningful message: "Input string was not in a correct format." The second one is apparently only for the insiders. ;-)
We might leave it that way anyhow.

msevestre:
Fixed by #974

PR #974, opened 2018-10-16
Fixes #942 manual editing of observed data

### LINKED ISSUE #942: Manual editing of observed data: error after deleting value
When manually editing observed data and (mistakenly by pressing Del or Backspace) deleting a value and hitting enter, the error message below is thrown. For some reason, the empty cell is kept afterwards and not flagged red, as done for invalid entries.

grafik

grafik

msevestre:


msevestre:
also Fixes #941

msevestre:
also fixes Open-Systems-Pharmacology/PK-Sim#902

PR #978, opened 2018-10-17
Fixes #309 ontogeny factor visible in simulation

### LINKED ISSUE #309: Should Ontogeny factor, ontogeny factor GI be visible and editable in a simulation
Right  now, those molecule specific parameters are hidden explicitly in the simulation.
also they are set as Editable = false in the database. My suggestion would be to make them visible at least in advanced mode and maybe even editable

msevestre:
This was not consistent at all here.
Ontogeny Factor, Ontogeny factor GI, Ontogeny Factor Albumin and AGP were visible but not editable in the individual
For those parameters, the default value was not set properly

In simulation, only Albumin and AGP were visible. other hidden

That should make everything consistent. One can now edit the factor also (only advanced parmeter anyways).


src/PKSim.Core/Services/MoleculeOntogenyVariabilityUpdater.cs
```diff
@@ -55,8 +54,14 @@ public void UpdatePlasmaProteinsOntogenyFor(Population population)
...
          molecule.Ontogeny = ontogeny;
-         molecule.OntogenyFactorGI = _ontogenyRepository.OntogenyFactorFor(ontogeny, CoreConstants.Groups.ONTOGENY_DUODENUM, individual.OriginData);
-         molecule.OntogenyFactor = _ontogenyRepository.OntogenyFactorFor(ontogeny, CoreConstants.Groups.ONTOGENY_LIVER, individual.OriginData);
+         updateOntogenyParameter(molecule.OntogenyFactorGIParameter, _ontogenyRepository.OntogenyFactorFor(ontogeny, CoreConstants.Groups.ONTOGENY_DUODENUM, individual.OriginData));
+         updateOntogenyParameter(molecule.OntogenyFactorParameter, _ontogenyRepository.OntogenyFactorFor(ontogeny, CoreConstants.Groups.ONTOGENY_LIVER, individual.OriginData));
+      }
+
+      private void updateOntogenyParameter(IParameter parameter, double value)
+      {
+         parameter.DefaultValue = value;
```
msevestre: When an ontogeny parameter is updated, we set the VALUE And the default value. It was done for AGP and ALB but not for OF and OF GI

src/PKSim.Core/Services/SimulationModelCreator.cs
```diff
@@ -116,9 +116,6 @@ private void updateAllParametersFor(Simulation simulation, Individual individual
          var globalMoleculeContainer = simulation.Model.Root
             .GetSingleChildByName<IContainer>(molecule.Name);
 
-         hideParameter(globalMoleculeContainer.Parameter(CoreConstants.Parameters.ONTOGENY_FACTOR_GI));
```
msevestre: Not sure why they were explicitly hidden. Definitely not required...We were too conservative at the beginning :)

PR #989, opened 2018-11-09
added tools for database dump/comparison

Yuri05:
With every new PR where PKSimDB was changed, I'm now going to create 2 files:
1. Full text dump of the database in `src/Db/Dump/PKSimDB_dump.txt`
2. SQL diff in `src/Db/Diff/PKSimDB_diff.txt` (text file with SQL commands that would transform `OSP/develop/PKSimDB.sqlite` into `PR/PKSimDB.sqlite`)

Both does not happen automatically.
* For 1. `src/Db/Dump/dumpdb.bat` must be executed (could be automated e.g. with a Pre-Commit-Hook, but I think it's not required)
* For 2.
  a. [`OSP/develop/PKSimDB.sqlite`](https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/develop/src/Db/PKSimDB.sqlite) must be downloaded into `src/Db/Diff/` (@msevestre any idea how to automate this?)
  b. `src/Db/Diff/diffdb.bat` must be executed

Both actions should help in PR review process if PK-Sim DB was changed (additionally to unit tests).


msevestre:
PKSimDb in develop will be available for download using the following link
`https://github.com/Open-Systems-Pharmacology/PK-Sim/raw/develop/src/Db/PKSimDB.sqlite`

If you use something more advance that a bat file (!!!), for example rake, you could download the file before starting the comparison. The whole process would simply be sthg like rake dbdump

The bat file could call this task but that;s about it. We should go away from scripting using bat file

msevestre:
@Yuri05 I have spent the last 2 hours trying to get the dump fro sqlite3 working and I did not manage.
Even with your bat file it does not work. 
I am not sure what's happening

msevestre:
Fixes #991

PR #993, opened 2018-11-12
Fixes #992 download script

### LINKED ISSUE #992: Download develop version of DB with script
This is a follow up of  #991

@Yuri05 I'd like to try to implement it in rake for consistency's sake

Also I think we should delete the PKSimDB.mdb

msevestre:


.gitignore
```diff
@@ -270,3 +270,4 @@ paket-files/
 /src/Db/03_PKSimDB_TissueAndBloodWeight.mdb
 /src/Db/7.2_PregnantWoman.xlsx
 /src/Db/PKSimDB.ldb
+/src/Db/**/*.develop.sqlite
```
msevestre: This is the file that will be downloaded when the script is started

db_pre_commit.bat
```diff
@@ -0,0 +1,4 @@
+@echo off
+cls
+rake db_pre_commit
```
msevestre: Entry point just for you. The only think that a bat file should do in PKSim, **for consistency and maintainability**,  is call rake tasks

rakefile.rb
```diff
@@ -2,6 +2,7 @@
 require_relative 'scripts/copy-dependencies'
 require_relative 'scripts/utils'
 require_relative 'scripts/coverage'
+require_relative 'src/Db/db'
```
msevestre: This is the file where the logic for dump and diff is implemented

rakefile.rb
```diff
@@ -108,6 +109,11 @@
 	end
 end
 
+task :db_pre_commit do
+	Rake::Task['db:dump'].execute();
```
msevestre: Just calling the task one after the other

src/Db/db.rb
```diff
@@ -0,0 +1,75 @@
+require 'open-uri'
+require 'openssl'
+require 'open3'
+PKSIM_DB_FILE_NAME = 'PKSimDB.sqlite'
+
+namespace :db do
+	desc "Performs a dump of the database"
+	task :dump do
+		puts "Starting database dump..."
+		sh(sqlite3, "#{pksimDb}", ".output #{dump_file}", ".dump", ".exit" )
```
msevestre: I managed to get it right after all

src/Db/db.rb
```diff
@@ -0,0 +1,75 @@
...
+		puts "Database dump completed and exported under #{dump_file}"
+	end
+	
+	desc "Downloads the develop version of the PKSim Database and performes a diff with the current version of the database"
+	task :diff do
+		puts "Starting datbase diff..."
+		download_pksimdb_file
```
msevestre: download the file first

src/Db/db.rb
```diff
@@ -0,0 +1,75 @@
...
+		stdout, stderr, status = Open3.capture3(sqldiff, "--primarykey", developPKSimDB, pksimDb)
```
msevestre: then call sqldiff. I had to use capture to capture stdout and put it in a file

src/Db/db.rb
```diff
@@ -0,0 +1,75 @@
...
+private
+
+def create_diff_file(content)
+  File.open(diff_file, 'w') do |file| 
+      file.puts content
+  end
+end
+
+def download_pksimdb_file
+	uri = "https://github.com/Open-Systems-Pharmacology/PK-Sim/raw/develop/src/Db/PKSimDB.sqlite"
```
msevestre: This is the DB File in develop

src/Db/db.rb
```diff
@@ -0,0 +1,75 @@
+require 'open-uri'
+require 'openssl'
+require 'open3'
+PKSIM_DB_FILE_NAME = 'PKSimDB.sqlite'
+
+namespace :db do
+  desc "Performs a dump of the database"
+  task :dump do
+    puts "Starting database dump..."
+    sh(sqlite3, "#{pksimDb}", ".output #{dump_file}", ".dump", ".exit" )
```
msevestre: each . command with sqlite3 can be executed one after the other that way

src/Db/db.rb
```diff
@@ -0,0 +1,75 @@
...
+    puts "Database dump completed and exported under #{dump_file}"
+  end
+  
+  desc "Downloads the develop version of the PKSim Database and performes a diff with the current version of the database"
+  task :diff do
+    puts "Starting datbase diff..."
+    download_pksimdb_file
+    stdout, stderr, status = Open3.capture3(sqldiff, "--primarykey", developPKSimDB, pksimDb)
```
msevestre: This weird construct is required because of the need to capture stdout.

src/Db/db.rb
Yuri05: interesting

PR #996, opened 2018-11-12
994 unify observed data import

msevestre:


msevestre:
@Yuri05 I had to update SmartXLS as well because it expired today...

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -1103,10 +1103,7 @@ public static class MenuNames
          public static readonly string NewFormulation = "Add &Formulation...";
          public static readonly string NewEvent = "Add &Event...";
          public static readonly string AddObservedData = "Add &Observed Data...";
-         public static readonly string AddAmountObservedData = "Add &Amount Observed Data...";
```
msevestre: None of that is required. Only import observed data. The selection of dimension will set the way the data is being imported

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -1609,6 +1606,7 @@ public static class UI
          public static readonly string OntogenyFactor = "Ontogeny factor";
          public static readonly string Concentration = "Concentration";
          public static readonly string Amount = "Amount";
+         public static readonly string Measurement = "Measurement";
```
msevestre: Using the same nomenclature as in MoBi

src/PKSim.Infrastructure/Services/ImportObservedDataTask.cs
```diff
@@ -106,78 +93,42 @@ private void adjustMolWeight(DataRepository observedData)
...
-
-         return columns;
-      }
 
-      private IReadOnlyList<ColumnInfo> concentrationImportConfiguration()
-      {
-         var columns = new List<ColumnInfo>();
-         var timeColumn = createTimeColumn();
-         columns.Add(timeColumn);
+         var supportedDimensions = new[] {
```
msevestre: MoBi allows everyting to be imported. I am using just a subset of what we want to support in PKSim
That way we also get fraction error if we wish

src/PKSim.Infrastructure/Services/ImportObservedDataTask.cs
```diff
@@ -303,6 +259,7 @@ private void addPredefinedCompartmentValues(MetaDataCategory compCategory)
          addInfoToCategory(compCategory, new Observer().WithName(CoreConstants.Observer.TISSUE));
          addInfoToCategory(compCategory, new Observer().WithName(CoreConstants.Observer.INTERSTITIAL_UNBOUND));
          addInfoToCategory(compCategory, new Observer().WithName(CoreConstants.Observer.INTRACELLULAR_UNBOUND));
+         addInfoToCategory(compCategory, new Observer().WithName(CoreConstants.Observer.WHOLE_BLOOD));
```
msevestre: Also fixes that other issue at the same time since I was already in there

src/PKSim.Assets/PKSimConstants.cs
Yuri05: Good idea

PR #1002, opened 2018-11-14
Fixes #995 seed can be negative

### LINKED ISSUE #995: Seed can be negative in PKSim.
https://docs.microsoft.com/en-us/dotnet/api/system.environment.tickcount?view=netframework-4.7.2

Not sure if this is an issue or not. We can definitely use the trick described in this document to ensure it's always positive

msevestre:


msevestre:
The changes in the dump file are just line endings. Something in our setings is not quite right

tests/PKSim.Tests/Core/IndividualFactorySpecs.cs
```diff
@@ -132,7 +132,7 @@ public void should_set_the_individual_as_loaded()
       [Observation]
       public void individual_should_have_a_random_seed()
       {
-         _individual.Seed.ShouldBeGreaterThan(0);
+         _individual.Seed.ShouldNotBeEqualTo(0);
```
msevestre: can be negative after all

PR #1003, opened 2018-11-16
Fixes #854 Value origins for pregnancy pop not correct

### LINKED ISSUE #854: Value origins for pregnancy pop not correct
The value origin for the organ volumes and blood flows of the pregnancy population is not correct.
The current reference is https://rd.springer.com/article/10.1007%2Fs40262-017-0539-z.
However, this reference gives the unprocessed values as reported in the literature. Prior to inclusion in the PK-Sim database, almost all of these values have been scaled to have them summing up 60 kg at the very onset of pregnancy (fetal age week 0) and the PK-Sim implemented cardiac output at that stage. 
The scaling to the PK-Sim level is described in the following paper:
https://rd.springer.com/article/10.1007%2Fs40262-017-0538-0

Finally, the reference given for the specific GFR and the maternal (but not fetal) hematocrit should also be the previously mentioned paper.

Version 7.3.0 Build 42

Yuri05:


src/Db/db.rb
```diff
@@ -2,6 +2,7 @@
 require 'openssl'
 require 'open3'
 PKSIM_DB_FILE_NAME = 'PKSimDB.sqlite'
+OpenSSL::SSL::VERIFY_PEER = OpenSSL::SSL::VERIFY_NONE
```
Yuri05: After I've added environment variable **HTTPS_PROXY** and set it to my default windows proxy AND added this line, I am able to download the database via the rake script.

msevestre:
The dump is already in that case hard to read.... but better than nothing

src/Db/db.rb
Yuri05: >What does that mean
HTTPS_PROXY and set it to my default windows proxy

Is set to the http address of my windows proxy. Not including userid or password - so this will not be changed every month. If rake would give the possibility for something like 
`set HTTPS_PROXY=${DEFAULT_WINDOWS_PROXY}` there would be no need for definition of extra environment variables.

>I had to add this line for the suite setup at some point. Maybe it's enough with it without the HTTPS_PROXY?

Nope, download still fails if the proxy variable is not set.

PR #1012, opened 2018-11-21
Fixes #1811 wrong value origin

msevestre:


src/PKSim.Presentation/DTO/Mappers/CompoundToCompoundTypeDTOMapper.cs
```diff
@@ -21,15 +21,15 @@ public CompoundToCompoundTypeDTOMapper(IParameterToParameterDTOInContainerMapper
          _parameterMapper = parameterMapper;
       }
 
-      public CompoundTypeDTO MapFrom(IEnumerable<IParameter> compoundParameters)
```
msevestre: This type was doing nothing. Simply easier to return a list of sub types

src/PKSim.Presentation/Presenters/Compounds/CompoundTypeGroupPresenter.cs
```diff
@@ -85,10 +85,11 @@ public IEnumerable<CompoundType> AllCompoundTypes()
...
       public void EditCompoundParameters(IEnumerable<IParameter> compoundParameters)
       {
-         _compoundTypeParameters = compoundParameters.Where(x => string.Equals(x.GroupName, CoreConstants.Groups.COMPOUND_PKA)).ToList();
-         _parameterCache = new PathCache<IParameter>(_entityPathResolver).For(_compoundTypeParameters);
-         _compoundTypeDTO = _dtoMapper.MapFrom(_compoundTypeParameters);
-         _view.BindTo(_compoundTypeDTO.AllTypePKas);
+         var allCompoundPKaParameters = compoundParameters.Where(x => string.Equals(x.GroupName, CoreConstants.Groups.COMPOUND_PKA)).ToList();
+         _parameterCache = new PathCache<IParameter>(_entityPathResolver).For(allCompoundPKaParameters);
+         _compoundTypeDTOs = _dtoMapper.MapFrom(allCompoundPKaParameters);
+         _compoundTypeParameters = _compoundTypeDTOs.SelectMany(x => x.Parameters).ToList();
```
msevestre: These are all the parameters actually editable. we need to keep the list handy so that we can update all ValueOrigin at once

Yuri05: I don't see why we have now only EDITABLE parameters here (as opposite to the previous state, where ALL parameters were passed)

msevestre:
Because we have a lot of parameters in the compound type group such as
IsAcid IsNeutral etc... The values origin of those parameters should not
change

PR #1015, opened 2018-11-22
Fixes #1013 Update references to all microsoft packages

### LINKED ISSUE #1013: Issue with PKSim.Matlab... referenced assembly not foud

msevestre:


msevestre:
@Yuri05 Can you try to use this variation if PKSim.Matlab and see if the problem is solved?
The loging abstraction was 2.1 but the login implementation was 2.0...so this was probably why

src/PKSim.Matlab/app.config
```diff
@@ -34,6 +34,22 @@
         <assemblyIdentity name="Microsoft.Extensions.Logging.Abstractions" publicKeyToken="adb9793829ddae60" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-2.1.1.0" newVersion="2.1.1.0" />
```
Yuri05: Not sure why we need reference to `Microsoft.Extensions.Logging` in `PKSim.Matlab` if it's not used.

msevestre:
It's because it's a down reference of ospsuite.core
This was added automatically as peer dependencies

PR #1021, opened 2018-11-23
Fixes #1020 Return molecule ontogenies in MatlabPopulationFactory

### LINKED ISSUE #1020: Return molecule ontogenies in MatlabPopulationFactory
Return molecule ontogenies in MatlabPopulationFactory

Yuri05:


src/PKSim.Matlab/MatlabPopulationFactory.cs
```diff
@@ -1,41 +1,73 @@
...
          var populationSettings = _populationSettingsMapper.MapFrom(matlabPopulationSettings);
          var population = _randomPopulationFactory.CreateFor(populationSettings, new CancellationToken()).Result;
+
+         foreach (var moleculeOntogeny in moleculeOntogenies)
+         {
+            var allOntogeniesForSpecies = _ontogenyRepository.AllFor(matlabPopulationSettings.Individual.OriginData.Species).ToList();
+            if (!allOntogeniesForSpecies.Any())
+               continue;
+
+            var ontogeny = allOntogeniesForSpecies.FindByName(moleculeOntogeny.Ontogeny);
```
msevestre: Shouldn't that be an error here? Somehow someones asks for an ontogeny that does not exist?

msevestre: looking good otherwise. If this is ok, you can merge

PR #1029, opened 2018-11-27
973 problem with relative expression

msevestre:


src/Db/Diff/PKSimDB_diff.txt
```diff
@@ -1,12 +1,4 @@
...
-UPDATE tab_calculation_method_rates SET formula='min(TM * kcat * max(min(Solubility/MW; DrugLiquid/Liquid);0)^alpha / (Km^alpha+max(min(Solubility/MW; DrugLiquid/Liquid);0)^alpha); P_int_trans*Aeff*P_int_trans_lum_cell_factor * min(Solubility/MW; DrugLiquid/Liquid)) + TM * kcat * (K_water_cell * max(C_cell;0))^alpha / (Km^alpha+(K_water_cell * max(C_cell;0))^alpha)' WHERE calculation_method='LinksCommon' AND formula_rate='MucosaPgPToLumen_Hill';
-UPDATE tab_calculation_method_rates SET formula='Min(CP * f_cell * V * kcat * (K_water_int * max(C_int;0))^alpha / ( Km^alpha + (K_water_int * max(C_int;0))^alpha) ; PA_int_cell * K_water_int * C_int) + CP * f_cell * V * kcat * (K_water_cell * max(C_cell;0))^alpha / (Km^alpha + (K_water_cell * max(C_cell;0))^alpha)' WHERE calculation_method='LinksCommon' AND formula_rate='PgpSpecific_Hill';
+UPDATE tab_container_parameters SET sequence=2 WHERE container_id=5015 AND container_type='GENERAL' AND container_name='PROTEIN' AND parameter_name='Relative expression in plasma';
```
msevestre: Adding those guys here to ensure that we have the same display as in indivudal. This set to 1 for all of them thus the order was not always the same. Doof

src/PKSim.Core/Commands/SetRelativeExpressionAndNormalizeCommand.cs
```diff
@@ -1,18 +1,18 @@
...
       private readonly double _newValue;
       private IndividualMolecule _molecule;
-      private IParameter _relExpParameter;
+      private IParameter _relativeExpressionParameter;
 
-      public SetRelativeExpressionAndNormalizeCommand(IndividualMolecule molecule, string proteinContainerName, double value)
+      public SetRelativeExpressionAndNormalizeCommand(IndividualMolecule molecule, IParameter relativeExpressionParameter, double value)
       {
          _molecule = molecule;
-         _relExpParameter = molecule.GetRelativeExpressionParameterFor(proteinContainerName);
```
msevestre: Not sure why we were passing a string to retrieve the parmeter from the container. What about we pass the parameter instead ?

src/PKSim.Core/Services/ParameterTask.cs
```diff
@@ -330,10 +330,18 @@ public ICommand SetParameterValue(IParameter parameter, double value, bool shoul
 
       private IPKSimCommand commandForRelativeExpressionParameter(IParameter parameter, double value)
       {
-         if (!parameter.IsExpressionNorm())
-            return new SetRelativeExpressionInSimulationAndNormalizedCommand(parameter, value);
+         //Normalized expression parameters are readonly and will be updated when setting relative expression
+         if (parameter.IsExpressionNorm())
+            return new PKSimEmptyCommand();
```
msevestre: This is the big change here. If a normalized parameter value is set (which can happen when trying to update dyanmically one container from another, for example when update or commit is performed), then we do not do anything

msevestre: the idea is that the rel exp will also be changed and will update the rel exp norm anyways. Previously, it was  updating both ways resulting in very weird outcome

src/PKSim.Core/Services/ParameterTask.cs
```diff
@@ -330,10 +330,18 @@ public ICommand SetParameterValue(IParameter parameter, double value, bool shoul
...
 
-         return new SetRelativeExpressionFromNormalizedCommand(parameter, value);
+         var expressionContainer = parameter.ParentContainer;
+         switch (expressionContainer.ParentContainer)
```
msevestre: This was also wrong somehow. Not crucial at that stage but the individual in the simulation was not update as an invididual/ As such the value of the internal building block (the normalized value in fact) could be out of sync. Probably not an issue anyways but well might as well be right

src/PKSim.Presentation/DTO/Mappers/ExpressionParametersToSimulationExpressionsDTOMapper.cs
```diff
@@ -90,6 +90,7 @@ private ExpressionContainerDTO expressionContainerFor(IParameter relativeExpress
 
             group = _groupRepository.GroupByName(CoreConstants.Groups.VASCULAR_SYSTEM);
             expressionContainerDTO.ContainerPathDTO = _representationInfoRepository.InfoFor(RepresentationObjectType.CONTAINER, containerNameForGlobalExpression(relativeExpression.Name)).ToPathElement();
+            expressionContainerDTO.Sequence = relativeExpression.Sequence;
```
msevestre: In this particular case, I am, setting the sequence from the parameter. And won't update it after that.

src/PKSim.Core/Services/ParameterTask.cs
msevestre: the command `SetRelativeExpressionFromNormalizedCommand` is therefore not use anywhere. We might as well delete it at some point  (setting relexp from norm values)

PR #1038, opened 2018-12-03
Fixes #861 multiple metabolite

### LINKED ISSUE #861: Feature wish: Multiple metabolites arising from a single enzyme
Currently, one enzyme can only produce one metabolite. Frequently, multiple metabolites are formed from one enzyme, e.g. 1-hydroxylation and 4-hydroxylation of midazolam from CYP3A4.

It would be very helpful to add the possibility to model multiple metabolites arising from a single enzyme (and NOT make use of the work-around with doubling (or even more) the enzyme).

msevestre:


src/PKSim.Core/Services/MoleculesAndReactionsCreator.cs
```diff
@@ -290,16 +290,14 @@ private void addMetabolismProcess(IMoleculeBuilder compoundBuilder, IReactionMap
 
       private IMoleculeBuilder getOrCreateMetaboliteFor(IReactionMapping compoundReactionMapping, CompoundProperties compoundProperties)
       {
-         IMoleculeBuilder metabolite = null;
-         var enzymaticProcess = compoundReactionMapping as EnzymaticProcessSelection;
-         if (enzymaticProcess != null)
-            metabolite = _moleculeBuildingBlock[enzymaticProcess.MetaboliteName];
+         var enzymaticProcess = compoundReactionMapping.DowncastTo<EnzymaticProcessSelection>();
```
msevestre: This is always an EnzymaticPRocessSeletion here. Let's assert it!

src/PKSim.UI/Views/Simulations/SimulationCompoundEnzymaticProcessView.cs
```diff
@@ -79,7 +78,6 @@ private RepositoryItem repositoryItemForMetabolites(SimulationEnzymaticProcessSe
       public void HideMultipleCompoundsColumns()
       {
          _colMetaboliteName.UpdateVisibility(false);
-         _colAddPartialProcess.UpdateVisibility(false);
```
msevestre: Always visible

src/PKSim.Core/Services/MoleculesAndReactionsCreator.cs
msevestre: This is actually not accurate. Systemic process are also here!!

PR #1094, opened 2019-02-12
Fixes #1090 json color

### LINKED ISSUE #1090: Json Snapshot: Export color using RGB?
Some colors are exported using the system string. This might be a problem when using the snapshot plot from the reporting engine. Let's export using rgb

msevestre:
@Yuri05 It seems that loading from HEX Numbers works out of the box

### DIFF
--- a/src/PKSim.Infrastructure/Serialization/Json/ColorConverter.cs
+++ b/src/PKSim.Infrastructure/Serialization/Json/ColorConverter.cs
@@ -1,0 +1,33 @@
+﻿using System;
+using System.Drawing;
+using Newtonsoft.Json;
+using OSPSuite.Core.Extensions;
+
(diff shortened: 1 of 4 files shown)


src/PKSim.Infrastructure/Serialization/Json/ColorConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+         writer.WriteValue(hexString);
+      }
+
+      public override object ReadJson(JsonReader reader, Type objectType, object existingValue, Newtonsoft.Json.JsonSerializer serializer)
+      {
+         //this will never be called as default color reader will take precedence
+         return existingValue;
+      }
+
+      public override bool CanRead => false;
```
msevestre: because of that, the old snapshot are compatible. yep yep

PR #1112, opened 2019-02-22
Fixes #1111 rename ref

### LINKED ISSUE #1111: Rename refproject to project and refsimulation to simulation

msevestre:


msevestre:
@Yuri05 Most changes are in the same file as the runner (those files will move to core eventually)

PR #1126, opened 2019-03-07
1123 markdown export

msevestre:


src/PKSim.Infrastructure/InfrastructureRegister.cs
```diff
@@ -251,6 +256,19 @@ private static void registerReportBuilders(IContainer container)
          container.Register<IReportBuilderRepository, ReportBuilderRepository>(LifeStyle.Singleton);
       }
 
+
+      private static void registerMarkdownBuilders(IContainer container)
```
msevestre: similar to our tex builder, we need to implement markdown builder. It should be much simple and I believe could potentially replace Tex in the future

src/PKSim.Infrastructure/Reporting/Markdown/Builders/CompoundMarkdownBuilder.cs
```diff
@@ -0,0 +1,26 @@
...
+   public class CompoundMarkdownBuilder : MarkdownBuilder<Compound>
+   {
+      private readonly IMarkdownBuilderRepository _markdownBuilderRepository;
+
+      public CompoundMarkdownBuilder(IMarkdownBuilderRepository markdownBuilderRepository)
+      {
+         _markdownBuilderRepository = markdownBuilderRepository;
+      }
+
+      public override void Report(Compound compound, MarkdownTracker tracker)
```
msevestre: Example of (too simple until specified) markdown export for compound

PR #1137, opened 2019-03-19
1136 update to new core structure

msevestre:


msevestre:
@Yuri05 This is the implememtation using the new OSPSuite.Engine. I needed to also update ContextSpecifrication in BDDHElper so that I could inject a config step. Not the best but I am not sure how to do it otherwise. Hopefully setup will work :)

msevestre:
@Yuri05 FIxes the tests. Merging

PR #1142, opened 2019-03-22
Fixes #1138 observer building block management

### LINKED ISSUE #1138: Add Observer as new building block in PKSim
This includes
Create ObserverBuildingBlock
Deleting ObserverBuidlingBlock
Adding observer to building block
Removing observer from building block

msevestre:


msevestre:
@Yuri05 This is covering the creation of a building block as well as adding and removing single observers. Going to work know on ObserverBuildigBlock management itself (saving to project, which will require a DB Project conversion, saving or loading from template etc..)
This will be part of the next issue

msevestre:
@Yuri05 Need to merge this to move with rest of implementation. PLease review

src/PKSim.Core/CoreConstants.cs
```diff
@@ -1330,6 +1330,8 @@ public static class UI
          public const int EXPRESSION_QUERY_VIEW_WIDTH = 1100;
          public const int FORMULATION_VIEW_WITDH = 650;
          public const int FORMULATION_VIEW_HEIGHT = 760;
+         public const int OBSERVER_VIEW_HEIGHT = 760;
+         public const int OBSERVER_VIEW_WITDH = 650;
```
Yuri05: why are some consts here defined as `static readonly` and some as `const`?

src/PKSim.Core/Services/IObserverLoader.cs
```diff
@@ -0,0 +1,106 @@
...
+         _containerObserverSerializer = modellingXmlSerializerRepository.SerializerFor<ContainerObserverBuilder>();
+         _amountObserverSerializer = modellingXmlSerializerRepository.SerializerFor<AmountObserverBuilder>();
+      }
+
+      private void convertXml(XElement sourceElement, int version)
+      {
+         if (sourceElement == null) return;
+         //set version to avoid double conversion in the case of multiple load
+         convert(sourceElement, version, x => x.ConvertXml);
+         sourceElement.SetAttributeValue(Constants.Serialization.Attribute.VERSION, Constants.PKML_VERSION);
```
Yuri05: What is this about? Observers imported from MoBi do not store pkml version?

src/PKSim.Presentation/DTO/Observers/ObserverDTO.cs
```diff
@@ -0,0 +1,60 @@
...
+            details.Add(InBold("Formula"));
+            details.Add(AddNewLine(Observer.Formula.ToString()));
+
+            details.Add(InBold("References"));
+            Observer.Formula.ObjectPaths.Each(x => { details.Add($"{x.Alias}:  {x.PathAsString}"); });
+
+            details.Add(StartWithNewLine(InBold("Container Criteria")));
+            details.Add(Observer.ContainerCriteria.ToString());
+            
+            return details;
```
Yuri05: Some details are missing here:
* Is imported observer Amount-Observer or Container-Observer
* Is defined for all molecules or not
* Molecules-IncludeList (if defined not for all molecules)
* Molecules-ExcludeList (if defined for all molecules)
* Dimension

src/PKSim.Presentation/Presenters/Observers/ImportObserversPresenter.cs
```diff
@@ -0,0 +1,128 @@
...
+         var newFile = getNewFile();
+         if (string.IsNullOrEmpty(newFile))
+            return;
+
+         var observer = _observerTask.LoadObserverFrom(newFile);
+         if (observer == null)
+            return;
+
+         if (_observerDTOs.ExistsByName(observer.Name))
+            throw new OSPSuiteException(PKSimConstants.Error.NameAlreadyExistsInContainerType(observer.Name, ObjectTypes.ObserverBuildingBlock));
```
Yuri05: This is not consistent with Load-Behaviour of other objects (templates, MoBi-Simulations from pkml etc.)
In other cases a Rename-Dialog is shown and user can rename loaded object

src/PKSim/Program.cs
```diff
@@ -23,7 +23,7 @@ private static void Main(string[] args)
          Application.EnableVisualStyles();
          Application.SetCompatibleTextRenderingDefault(false);
 
-         WindowsFormsSettings.SetDPIAware();
+      //   WindowsFormsSettings.SetDPIAware();
```
Yuri05: What is this change about?

src/PKSim.Core/CoreConstants.cs
msevestre: Const: They can become default value in function and their value replaces the constant at build time. Static readonly, they cannot be changed but value is not replaced by constant at build time=>cannot be used as default value

src/PKSim.Presentation/DTO/Observers/ObserverDTO.cs
msevestre: created an issue here https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1147

src/PKSim.Presentation/Presenters/Observers/ImportObserversPresenter.cs
msevestre: Yeah but this is not the same level however. We are loading an element IN a building block which we never do in PKSim. 
If yo u feel strongly about it, please create a specific issue. I think this is overkill in this case, especially since the observer are readonly in pksim

src/PKSim/Program.cs
msevestre: SetDPIAware redners the app unsuable in High Res screens. I was wondering why PKsim looked like shit and MoBi not on my 4k monitor (laptop). That's the reason.  I believe this flag was used for the wrong reason and I can't remember why we added it (not in MoBi for instance)

Yuri05: Issue created: #1148
I think this could easily happen when users create observers with commonly used names e.g. _TotalAmount_, _AverageConcentration_ etc.

msevestre: But they would have to want to load the SAME observer in the same buildign block. Why would they want to do that

Yuri05: E.g. user has defined 2 observers in 2 different Mobi projects; same name but different functionality E.g.
* project1: _TotalAmount_ (which sums up everything in _Kidney_)
* project2: _TotalAmount_ (which sums up everything in _Liver_)
Now user wants to have both in PK-Sim

Yuri05: Not the show stopper if it's not done, so let ask the users.

msevestre: IT;'s easy. It needs to be in different building blocks! And if they want to use it in the same simulation, well they should name stuff propertly

PR #1152, opened 2019-03-26
1150 comparison observer

msevestre:


src/PKSim.Core/Comparison/ObserverSetDiffBuilder.cs
```diff
@@ -0,0 +1,23 @@
...
+      public ObserverSetDiffBuilder(EntityDiffBuilder entityDiffBuilder, EnumerableComparer enumerableComparer)
+      {
+         _entityDiffBuilder = entityDiffBuilder;
+         _enumerableComparer = enumerableComparer;
+      }
+
+      public override void Compare(IComparison<ObserverSet> comparison)
+      {
+         _entityDiffBuilder.Compare(comparison);
+         _enumerableComparer.CompareEnumerables(comparison, x => x.Observers, x => x.Name);
```
msevestre: This is the most basic comparison really. Making sure observers are the same

src/PKSim.Core/Services/SimulationBuildingBlockUpdater.cs
```diff
@@ -59,6 +59,11 @@ public interface ISimulationBuildingBlockUpdater
       void UpdateFormulationsInSimulation(Simulation simulation);
 
       bool BuildingBlockSupportsQuickUpdate(IPKSimBuildingBlock templateBuildingBlock);
+
+      /// <summary>
+      /// Returns whether a building block comparison is available for the building block
+      /// </summary>
+      bool BuildingBlockSupportComparison(IPKSimBuildingBlock templateBuildingBlock);
```
msevestre: In our case, we can compare BB even though not quick update is possible. The assumption is not valid anymore

PR #1157, opened 2019-03-27
Fixes #1156 update generated output

### LINKED ISSUE #1156: Update generated output for qualification runner to be less verbose

msevestre:


src/PKSim.CLI.Core/Services/QualificationRunner.cs
```diff
@@ -112,12 +112,12 @@ public async Task RunBatchAsync(QualificationRunOptions runOptions)
          await _jsonSerializer.Serialize(mapping, config.MappingFile);
          _logger.AddDebug($"Project mapping for '{project.Name}' exported to '{config.MappingFile}'", project.Name);
 
-         var projectFile = Path.Combine(projectOutputFolder, $"{project.Name}{CoreConstants.Filter.PROJECT_EXTENSION}");
+         var projectFile = Path.Combine(config.TempFolder, $"{project.Name}{CoreConstants.Filter.PROJECT_EXTENSION}");
```
msevestre: @Yuri05 Still dumping project and snapshot, however into temp folder so that we can access them if we need for some reason

src/PKSim.CLI.Core/Services/SimulationExporter.cs
```diff
@@ -206,12 +205,5 @@ private async Task exportIndividualSimulationResultsToExcelAsync(IndividualSimul
          await _simulationExportTask.ExportResultsToExcelAsync(simulation, simulationResultFileFullPath, launchExcel:false);
          _logger.AddDebug($"Exporting simulation results to '{simulationResultFileFullPath}'", simulationExportOptions.LogCategory);
       }
-
-      private void exportParameters(IndividualSimulation individualSimulation, SimulationExportOptions simulationExportOptions)
```
msevestre: We never used that. It's left from early on and does notb ring anuthing. Removing

PR #1161, opened 2019-04-01
Fixes #1146 add observer to build configuration

### LINKED ISSUE #1146: Add selected observers to build configuration

msevestre:
I think it's that eas...

### DIFF
--- a/src/PKSim.Infrastructure/ORM/Queries/ModelObserverQuery.cs
+++ b/src/PKSim.Infrastructure/ORM/Queries/ModelObserverQuery.cs
@@ -1,6 +1,10 @@
 using System;
-using System.Collections.Generic;
 using System.Linq;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Core.Domain.Descriptors;
+using OSPSuite.Core.Domain.Formulas;
+using OSPSuite.Core.Domain.Services;
 using OSPSuite.Utility.Extensions;
 using PKSim.Core;
 using PKSim.Core.Extensions;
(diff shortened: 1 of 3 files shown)


src/PKSim.Infrastructure/ORM/Queries/ModelObserverQuery.cs
```diff
@@ -67,6 +68,13 @@ public IObserverBuildingBlock AllObserversFor(IMoleculeBuildingBlock moleculeBui
          return observerBuildingBlock;
       }
 
+      private void addSimulationObservers(Simulation simulation, IObserverBuildingBlock observerBuildingBlock)
+      {
+         simulation.AllBuildingBlocks<ObserverSet>()
```
msevestre: Don't care at that stage about duplicated name etc.. DEpending on condition (For all moelcule etc..) it could still va valid

msevestre: Also I don't think I should clone the observer here. They are already specific to the simulation AND They won't be persisted anywhere  as core building blocks are discarded

tests/PKSim.Tests/IntegrationTests/ModelObserverQuerySpecs.cs
```diff
@@ -41,6 +42,12 @@ public override void GlobalContext()
          _moleculeBuildingBlock.Add(new MoleculeBuilder {Name = _complexProductName, QuantityType = QuantityType.Complex});
          compoundProperties.Processes.SpecificBindingSelection.AddPartialProcessSelection(specificBindingSelection);
          compoundProperties.Processes.MetabolizationSelection.AddPartialProcessSelection(metabolizationSelection);
+
+         var observerSet = new ObserverSet();
+         var observerSetBuildingBlock = new UsedBuildingBlock("OBS_ID", PKSimBuildingBlockType.Observers) {BuildingBlock = observerSet};
+         _observer = new AmountObserverBuilder().WithName("Test");
```
msevestre: Adding a new observer to the configuration and checkign that it's here

src/PKSim.Infrastructure/ORM/Queries/ModelObserverQuery.cs
Yuri05: So in case we have 2 observers with the same name (and Container-Criteria will allow creation of both in the same parent container), build process will crash with "not unique" exception?

msevestre: yes. Name "MyObserer" is not unique in parent container "XXx"

I would love if the error would use a type instead of parent container but this is a good as it gets for now

PR #1164, opened 2019-04-02
Fixes #1163 MatlabIndividualFactory: "CreateIndividual" method without ontogenies 

### LINKED ISSUE #1163: MatlabIndividualFactory: "CreateIndividual" method without ontogenies
`CreateIndividual(OriginData matlabOriginData, IEnumerable<MoleculeOntogeny> moleculeOntogenies)`

requires an enumerable of molecule ontogenies. If no custom ontogenies are available, Matlab creates an empty vector []<MoleculeOntogeny> and passes it to `CreateIndividual`. The R-to-.NET interface cannot create empty arrays of a certain type, so `CreateIndividual(OriginData matlabOriginData, IEnumerable<MoleculeOntogeny> moleculeOntogenies` cannot be called.

Solution: create a method `CreateIndividual(OriginData matlabOriginData)` that would be called by R.

PavelBal:


PKSim.sln
```diff
@@ -106,8 +106,8 @@ Global
 		HideSolutionNode = FALSE
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
-		SolutionGuid = {13620E23-1993-42C6-A4AB-468E29DD24DE}
 		VisualSVNWorkingCopyRoot = 
+		SolutionGuid = {13620E23-1993-42C6-A4AB-468E29DD24DE}
```
msevestre: undo this change here. This is some Local SVN stuff that we do not use

src/PKSim.Matlab/MatlabIndividualFactory.cs
```diff
@@ -14,6 +14,7 @@ namespace PKSim.Matlab
    public interface IMatlabIndividualFactory
    {
       ParameterValue[] CreateIndividual(OriginData matlabOriginData, IEnumerable<MoleculeOntogeny> moleculeOntogenies);
```
msevestre: Would that work if you were to pass a parameter as null instead? 
e.g. the first method would be
```
ParameterValue[] CreateIndividual(OriginData matlabOriginData, IEnumerable<MoleculeOntogeny> moleculeOntogenies = null);
```

src/PKSim.Matlab/MatlabIndividualFactory.cs
```diff
@@ -65,7 +66,13 @@ public ParameterValue[] CreateIndividual(OriginData matlabOriginData, IEnumerabl
          return allIndividualParameters.ToArray();
       }
 
-      public DistributedParameterValue[] DistributionsFor(OriginData matlabOriginData, IEnumerable<MoleculeOntogeny> moleculeOntogenies)
+      public ParameterValue[] CreateIndividual(OriginData matlabOriginData)
+      {
+         IEnumerable<MoleculeOntogeny> moleculeOnogenies = new MoleculeOntogeny[0];
+         return CreateIndividual(matlabOriginData, moleculeOnogenies);
```
msevestre: If my suggestion from above does not work, simply pass null as an argument
```
return CreateIndividual(matlabOriginData, null);
```

msevestre: and change the implementation of `https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/102d271e84823151de2ae9449d5779b51f1cb809/src/PKSim.Matlab/OntogenyFactorsRetriever.cs#L47`
to deal with null as expected.

e.g. if the list of ontogeny is null return

```
      if (!allOntogeniesForSpecies.Any() || moleculeOntogenies==null)
            return allOntogenyFactors;
```

PKSim.sln
PavelBal: The only change here ist that the lines 109 and 110 were interchanged (for whatever reason), nothing new was added.

msevestre: ok. I thought the line with SVN was added. Cool then

src/PKSim.Matlab/MatlabIndividualFactory.cs
msevestre: @PavelBal 
Forget about that. We need to merge something to see if it works with R and Matlab while we still support both

Suggestion
REmplace 
```
IEnumerable<MoleculeOntogeny> moleculeOnogenies = new MoleculeOntogeny[0];
return CreateIndividual(matlabOriginData, moleculeOnogenies);
```

with 
```
return CreateIndividual(matlabOriginData, Enumerable.Empty<MoleculeOntogeny>());
```

and I'll merge

PavelBal: Updated.

PR #1168, opened 2019-04-04
Fixes #1167 markdown for categorial parameters

### LINKED ISSUE #1167: Markdown export: Export categorial parameter using their category instead of value

msevestre:


src/PKSim.Core/CoreConstants.cs
```diff
@@ -1150,7 +1150,7 @@ public static bool IsVolumeFraction(string parameterName)
             USE_AS_SUSPENSION
          };
 
-         public static readonly IReadOnlyCollection<string> AllWithListOfValues = new List<string>(Halogens.Union(AllBooleanParameters))
+         public static readonly HashSet<string> AllWithListOfValues = new HashSet<string>(Halogens.Union(AllBooleanParameters))
```
msevestre: This will allow for faster check using contains

src/PKSim.Core/Mappers/ParameterListOfValuesRetriever.cs
```diff
@@ -0,0 +1,88 @@
+﻿using System.Linq;
+using OSPSuite.Core.Domain;
+using OSPSuite.Utility.Collections;
+using PKSim.Assets;
+using PKSim.Core.Model;
+
+namespace PKSim.Core.Mappers
+{
+   public interface IParameterListOfValuesRetriever
```
msevestre: Move this out of the DTO mapper into its own file so that I can use it here as well

src/PKSim.Infrastructure/Reporting/Markdown/Elements/ParameterElement.cs
```diff
@@ -1,16 +1,18 @@
...
-      string ValueOrigin { get; set; }
+      ValueOrigin ValueOrigin { get; set; }
    }
 
    public class ParameterElement : IParameterElement
    {
       public string Name { get; set; }
       public string Value { get; set; }
-      public string ValueOrigin { get; set; }
+      public ValueOrigin ValueOrigin { get; set; }
```
msevestre: The markdown exporter calls to string on object so we are good here

src/PKSim.Core/Mappers/ParameterListOfValuesRetriever.cs
```diff
@@ -0,0 +1,88 @@
...
+         else if (CoreConstants.Parameters.AllBooleanParameters.Contains(parameter.Name))
+         {
+            listOfValues.Add(1, PKSimConstants.UI.Yes);
+            listOfValues.Add(0, PKSimConstants.UI.No);
+         }
+         else if (parameter.NameIsOneOf(CoreConstants.Parameters.PARA_ABSORBTION_SINK, CoreConstants.Parameters.TRANS_ABSORBTION_SINK))
+         {
+            listOfValues.Add(CoreConstants.Parameters.SINK_CONDITION, PKSimConstants.UI.SinkCondition);
+            listOfValues.Add(CoreConstants.Parameters.NO_SINK_CONDITION, PKSimConstants.UI.NoSinkCondition);
+         }
```
Yuri05: Missing exception at the end
(Parameter has list of values but nothing was added to the listOfValues-Cache)

msevestre: Yeah. Don't know why would throw an exception to be honest. I think in that case the parameter behaves as a normal parameter like in MoBi.

msevestre: But I am ok to put it back. No strong opinion here for once. This would be more of a dev error really (which We never do;-))

Yuri05: exactly :)

msevestre: ok updated. And even wrote tests :

msevestre:
@Yuri05 MErging that so that I can push the next PR

PR #1190, opened 2019-04-17
Fixes #1189 Nhibernate update

### LINKED ISSUE #1189: Until we update NHibernate, we should add log4net.dll in distribution folder
or we update NHibernate. I can try and see if our tests are still passing ??

msevestre:


msevestre:
@Yuri05 I just want to run all the test using this branch to see if anything breaks. I did not have to change any code except one method that was removed in the test

msevestre:
IF this is successful, we can decide to merge or not. This should solve the issue with dependency on log4net that is not even used

msevestre:
@Yuri05 I ran all of the tests locally as well (nightly tests)
Looks good
image

src/PKSim.Matlab/ApplicationStartup.cs
```diff
@@ -47,29 +47,27 @@ private void initializeForMatlab()
 
       private static void redirectNHibernateAssembly()
       {
-         redirectAssembly("NHibernate", new Version(4, 1, 0, 4000), "aa95f207798dfdb4");
+         redirectAssembly("NHibernate", new Version(5, 2, 0, 0), "aa95f207798dfdb4");
```
Yuri05: what is this aa95f207798dfdb4 ?

msevestre: This is the nhiberbate package id (see app.config). This code ensures that the right dll is loaded. We had issues when loading from Matlab

Yuri05:
@msevestre I cannot say if there is something wrong or not. I didn't see anything suspicious, so if you think that me can merge it - do it.

msevestre:
For me it also looks legit. All tests are passing so I think we should do it

PR #1192, opened 2019-04-17
Fixes #1073 open file share drive

### LINKED ISSUE #1073: ExpressionDB on share drive: Sometimes, user can't open the file
1. Can we reproduce it somehow? (@Yuri05 This should be possible for you considering the complexity of your IT infrastructure :-))
1. Can we do something about it?
1. Can we somehow catch this generic error and be more specific about it?

msevestre:


src/PKSim.Infrastructure/ORM/Core/Database.cs
```diff
@@ -39,7 +40,7 @@ public void Connect(string databasePath)
          if (!FileHelper.FileExists(databasePath))
             throw new OSPSuiteException(PKSimConstants.Error.FileDoesNotExist(databasePath));
 
-         DatabaseObject.Connect(databasePath, _userName, _password, GetProvider());
+         DatabaseObject.Connect(databasePath.ToUNCPath(), _userName, _password, GetProvider());
```
msevestre: This is the only place where we connect to freaking DAS

Issue #1202, opened 2019-04-23
Supersaturation: Prevent usage if NOT Particale Dissolution


msevestre:
![image](https://user-images.githubusercontent.com/1041237/56617450-eaa75480-65ed-11e9-9b0c-d3485cabc594.png)
This blurb is very, very unclear in my opinion.

I would remove this text altogether (e.g. leave it as it was before)
When creating a simulation, we had an extra step that prevents exactly that from happening (we have a validator that we run to test stuff that does not make sense). 
We can show the message displayed here "E.g. Enabling supersaturation ... solutes." as error if the user creates this scenario

Thoughts? @Yuri05 @sfrechen 

Also I would group input parameters together (e..g move the parameter under Treat precipitated drug as solublit)

Yuri05:
>When creating a simulation, we had an extra step that prevents exactly that from happening (we have a validator that we run to test stuff that does not make sense).

Might be a better approach compared to the warning in text.
Thus the suggestion would be:
* when creating a simulation, check for every compound involved whether `Enable supersaturation` is set to true
* if YES: check if compound is administered with an administration protocol
* if YES: check if administration protocol for this compound contains one of the following
  * at least one **oral** administration with `formulation ≠ particles dissolution` or
  * at least one **user defined** administration with `Target organ = Lumen` 
* if so: show error message and don't create simulation

Correct? @sfrechen @KatrinCoboeken @AndreDlm 

>Also I would group input parameters together

I have no opinion on this and can live with both parameter orders (current and proposed)

PR #1203, opened 2019-04-23
1197 increase info area

msevestre:


msevestre:
IMHO this is not the right way to solve the problem (see #1202 )

PR #1207, opened 2019-04-24
Fixes #1202 prevent usage

### LINKED ISSUE #1202: Supersaturation: Prevent usage if NOT Particale Dissolution

msevestre:


src/PKSim.Core/Services/SimulationConfigurationValidator.cs
```diff
@@ -47,9 +49,41 @@ public void ValidateConfigurationFor(Simulation simulation)
...
+         {
+            x.Compound,
+            x.ProtocolProperties.Protocol,
+            x.ProtocolProperties.FormulationMappings
+         });
+
+         administeredCompoundWithSuperSaturationEnabled.Each(x => validateSupersaturationUsageFor(x.Compound, x.Protocol, x.FormulationMappings));
+      }
+
+      private void validateSupersaturationUsageFor(Compound compound, Protocol protocol, IReadOnlyList<FormulationMapping> formulationMappings)
```
msevestre: This is the code called for a supersaturated compound, the protocol and the formulations used for that protocol and compound

src/PKSim.Core/Services/SimulationConfigurationValidator.cs
```diff
@@ -47,9 +49,41 @@ public void ValidateConfigurationFor(Simulation simulation)
...
+      {
+         var protocolSchemaItems = _schemaItemsMapper.MapFrom(protocol);
+
+         var hasOralApplicationsNotUsingParticles = protocolSchemaItems
```
msevestre: so clear I find. No ifs etc.. All query. Beautiful

tests/PKSim.Tests/Core/SimulationConfigurationValidatorSpecs.cs
```diff
@@ -211,4 +224,129 @@ public void should_not_throw_an_exception()
          sut.ValidateConfigurationFor(_simulation);
       }
    }
+
+   public class When_validating_the_configuration_of_a_simulation_using_a_compound_with_an_oral_application_super_saturation_and_not_particle_dissolution : concern_for_SimulationConfigurationValidator
```
msevestre: All tests for all use cases I could think of

PR #1214, opened 2019-04-26
Fixes #1213 crash exporting

### LINKED ISSUE #1213: Crash when exporting qualification runner
see https://github.com/Open-Systems-Pharmacology/QualificationRunner/issues/30

msevestre:


src/PKSim.CLI.Core/Services/QualificationRunner.cs
```diff
@@ -89,6 +89,8 @@ public async Task RunBatchAsync(QualificationRunOptions runOptions)
 
          var begin = DateTime.UtcNow;
          var project = await _snapshotTask.LoadProjectFromSnapshot(snapshot);
+         _workspace.Project = project;
```
msevestre: This is used implicitely when exporting. But the project was not set.
I made sure that it is set but also ensures that it does not crash if it is not set

PR #1233, opened 2019-05-07
Fixes #1232 snapshot load

### LINKED ISSUE #1232: Issue with snapshot not loading as expected

msevestre:


src/PKSim.Infrastructure/Serialization/WorkspacePersistor.cs
```diff
@@ -127,13 +128,14 @@ public void LoadSession(ICoreWorkspace workspace, string fileFullPath)
 
                   progress.IncrementProgress(PKSimConstants.UI.LoadingLayout);
                   var workspaceLayout = _workspaceLayoutPersistor.Load(session);
+                  // The workspace layout may be null if the workspace was created via CLI. In that case, we simply initialize the workspace layout
                   if (workspace is IWithWorkspaceLayout withWorkspaceLayout)
-                     withWorkspaceLayout.WorkspaceLayout = workspaceLayout;
+                     withWorkspaceLayout.WorkspaceLayout = workspaceLayout ?? new WorkspaceLayout();
```
msevestre: Project created from CLI are using the CLIWorkspace that does not have a workspace layout (presentation stuff)
So the layout was null and it was impossible to open the simulation in a project created via CLI. This is now fixed :) (only in 7.5)

PR #1249, opened 2019-05-15
Fixes #1248 bump to version /

### LINKED ISSUE #1248: Bump to version 8.0

msevestre:


rakefile.rb
```diff
@@ -88,7 +88,7 @@
 	packages_dir =  File.join(solution_dir, 'packages')
 
 	all_users_dir = ENV['ALLUSERSPROFILE']
-	all_users_application_dir = File.join(all_users_dir, manufacturer, product_name, '7.5')
+	all_users_application_dir = File.join(all_users_dir, manufacturer, product_name, '8.0')
```
msevestre: Still using the X.Y for release so that we can support something like a patch 8.1 without the need to worry about folder structure etc..
8.0 means Version 8
8.1 means Version 8 Update 1( Update, Patch, SP whatever)

src/PKSim.Presentation/Presenters/AboutPresenter.cs
```diff
@@ -31,7 +31,7 @@ public AboutPresenter(IAboutView view, IPKSimConfiguration configuration, IVersi
       public override void Initialize()
       {
          _view.Product = _configuration.ProductDisplayName;
-         _view.VersionInformation = $"Version {_configuration.FullVersionDisplay}";
+         _view.VersionInformation = $"Version {_configuration.FullVersion}";
```
msevestre: This will show the following
image

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
```diff
@@ -154,7 +154,7 @@ public void Handle(ApplicationInitializedEvent eventToHandle)
       {
          updateProjectInfo(Captions.None, Captions.None, false);
          update(StatusBarElements.Version)
-            .WithCaption(_applicationConfiguration.FullVersionDisplay);
+            .WithCaption($"Version {_applicationConfiguration.FullVersionDisplay}");
```
msevestre: This will show the following
image

src/PKSim.Presentation/Presenters/AboutPresenter.cs
Yuri05: @msevestre what about the build number?

msevestre: Build number is the last digit. 
8.5.4 => Version 8, Update 5 build 4

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
Yuri05: please change it back to the full version

msevestre: This is the full version. Version 8 or sthg. 
IF we are in EAP, or beta, it will say so here as well

PR #1257, opened 2019-05-21
Fixes #1254 load from csv

### LINKED ISSUE #1254: Cannot load population from CSV anymore
[09_ReadFromPop.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/3201692/09_ReadFromPop.zip)

When combining a-Mobi simulation with a population previously exported to CSV: population cannot be imporrted anymore due to the units in parameter column headers

grafik

msevestre:


src/PKSim.Core/Model/ImportPopulationFactory.cs
```diff
@@ -135,39 +132,11 @@ private Task<ImportResult> importFiles(string file, CancellationToken cancellati
          }, cancellationToken);
       }
 
-      private void updatePathsContaininUnits(IndividualPropertiesCache individualValues)
```
msevestre: THis code can be used as is. It's just in the wrong place. It was done in the `ImportPopulationFactory` instead of the `ImportFile`

src/PKSim.Infrastructure/Services/IndividualPropertiesCacheImporter.cs
```diff
@@ -39,7 +39,7 @@ public IndividualPropertiesCache ImportFrom(string fileFullPath, IImportLogger l
                var individualPropertyCache = individualPropertiesCacheFrom(fileFullPath, delimiter);
                //we found at least one individual, this is a valid file for the delimiter and we can exit
                if (individualPropertyCache?.Count > 0)
-                  return individualPropertyCache;
+                  return withPathsContainingUnitsUpdated(individualPropertyCache, allParameters);
```
msevestre: making sure that we update the path of parameter containing units before returning int That way, anyone using this object will have clean values

msevestre:
@Yuri05 Ready to go as far as I am concerned

Yuri05:
We assume that (CSV-)export is currently in **internal** units only, correct?
(thus no need for untis conversion)

Yuri05:
>We assume that (CSV-)export is currently in internal units only, correct? (thus no need for untis conversion)

@msevestre is this assumption correct?

msevestre:
yes. CSV is purely for exchange. If we start to allow user defined units here... OUCH

PR #1267, opened 2019-05-26
Fixes #1240 Enable supersaturation should be IsInput = false

### LINKED ISSUE #1240: Enable supersaturation should be IsInput = false

Yuri05:


Yuri05:
@msevestre I wonder why [PKSimDB_diff.txt](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/1267/files#diff-2c75f16e0dc82bb44b9d134bf0cd579b) is shown as binary in the files view.

PR #1269, opened 2019-05-29
Fixes #1268 cli message

### LINKED ISSUE #1268: Irritating info message
Info message "_Qualification run finished_" comes every time **_before_** the next project is loaded from snapshot and exported. Thus it looks like below for several minutes and the user might think that the qualification runner is finished (or crashed).
</img>

The info message should either come **_after_** the project is loaded from snapshot or should not come at all. 
Actually instead of "_Qualification run finished_" in the screen above I would expect something like "_Reloading project from snapshot. This might take some time_"

msevestre:


src/PKSim.CLI.Core/Services/QualificationRunner.cs
```diff
@@ -59,7 +59,7 @@ ILogger logger
       public async Task RunBatchAsync(QualificationRunOptions runOptions)
       {
          _snapshotProjectCache.Clear();
-         _logger.AddDebug("Starting qualification run");
+         _logger.AddInfo(runOptions.Validate ? "Starting validation run..." : "Starting qualification run...");
```
msevestre: Depending on validation flag, show a different message

src/PKSim.CLI/Commands/CLICommand.cs
```diff
@@ -8,6 +8,8 @@ public abstract class CLICommand
    {
       public abstract string Name { get; }
 
+      public virtual bool LogCommandName { get; } = true;
```
msevestre: true by default except for our qualificationrun

PR #1276, opened 2019-06-11
Fixes #1271 another strange behavior

### LINKED ISSUE #1271: Another Strange Behavior of the Chart Editor PK-Sim 7.4
steps to reproduce:
create Simulation with existing BB. click Run, let everything on default (only peripheral venous blood plasma is selcted) and click OK

image

results look like expected
image
than save project and close.
Open Project again, open Simulation (e.g. Test 3) . Now Venous Blood Plasma and Lume fraction of oral dose absorbed is visible (However it is not checked or selected)

image

Why does this happen apparently without any user input?

msevestre:


src/PKSim.Core/Services/DataRepositoryFromResultsCreator.cs
```diff
@@ -41,11 +54,32 @@ public DataRepository CreateResultsFor(IndividualSimulation simulation)
...
          var individualResults = simulation.Results.First();
          individualResults.Select(value => dataColumnFor(value, simulation, timeColumn))
-            .Where(col=> col!=null).Each(dataRepository.Add);
+            .Where(col => col != null).Each(dataRepository.Add);
 
+         UpdateColumnInternalUse(simulation, dataRepository);
          return dataRepository;
       }
 
+      public void UpdateColumnInternalUse(IndividualSimulation simulation, DataRepository dataRepository = null)
```
msevestre: @Yuri05 This function was private. I moved it and made it public. The issue was that when we load from project, we always recreate a DataRepository from scratch. There, the knowledge of intermal columns was lost

PR #1278, opened 2019-06-12
1277 chart not updated

msevestre:


src/PKSim.Presentation/Presenters/Charts/ChartPresenter.cs
```diff
@@ -92,9 +92,11 @@ protected virtual void UpdateAnalysisBasedOn(IndividualSimulation simulation, Da
       {
          if (_repositoryCache.Contains(dataRepository))
          {
-            ChartEditorPresenter.RemoveUnusedColumns();
-            AddDataRepositoriesToEditor(new []{dataRepository});
-            ChartDisplayPresenter.Refresh();
+            using (_chartUpdater.UpdateTransaction(Chart))
```
msevestre: @Yuri05 This is what needs to be done in order to refresh the stuff

Issue #1308, opened 2019-08-27
remove last '|'  from tables created by PK-Sim


Incei:
remove last '|'  from tables created by PK-Sim, causing an additional empty column in tables.

PR #1319, opened 2019-09-18
1308 remove last dash

msevestre:


src/PKSim.CLI.Core/Services/QualificationRunner.cs
```diff
@@ -228,7 +228,7 @@ private InputMapping exportInput(PKSimProject project, QualifcationConfiguration
          var fileFullPath = Path.Combine(targetFolder, $"{buildingBlockName}{CoreConstants.Filter.MARKDOWN_EXTENSION}");
 
          // Use wait for now until we can support // run of input
-         _markdownReporterTask.ExportToMarkdown(buildingBlock, fileFullPath).Wait();
+         _markdownReporterTask.ExportToMarkdown(buildingBlock, fileFullPath, input.SectionLevel).Wait();
```
msevestre: I have added the section level here which now comes from the QualificationPlan
For example here https://github.com/Open-Systems-Pharmacology/Raltegravir-Model/blob/master/evaluation/Input/Evaluation_plan.json#L444, the level would be 3 which was the default value that was supposed to be used.
If the element is defined somewhere else, its level will be automatically adjusted

src/PKSim.Core/Services/IMarkdownReporterTask.cs
```diff
@@ -4,7 +4,7 @@ namespace PKSim.Core.Services
 {
    public interface IMarkdownReporterTask
    {
-      Task ExportToMarkdown(object objectToExport, string file);
-      string ExportToMarkdownString(object objectToExport);
+      Task ExportToMarkdown(object objectToExport, string file, int? indentationLevel = null);
+      string ExportToMarkdownString(object objectToExport, int? indentationLevel = null);
```
msevestre: Default is null. The code underneath will set the  value to 1

src/PKSim.Infrastructure/Reporting/Markdown/MarkdownTracker.cs
```diff
@@ -30,10 +30,9 @@ public override string ToString()
       private string tableMarkdownFor(Table table)
       {
          return table.ToString()
-            .Remove(0, OFFSET.Length)                             // to remove weird offset at the beginning
-            .Replace(TABLE_OFFSET, TABLE_OFFSET_NO_INDENT)        // to remove offset for each row
-            .Replace($"- {NEW_LINE}", $"-{NEW_LINE}")             // to remove the extra space for the separator for the last column
-            .Replace(NEW_LINE, $" |{NEW_LINE}");                  // to add a separator at the end of each row
```
msevestre: This was added by "hand" and well, this was wrong. This fixes #1308

src/PKSim.Infrastructure/Services/MarkdownReporterTask.cs
```diff
@@ -17,19 +17,19 @@ public MarkdownReporterTask(IReportGenerator reportGenerator, IMarkdownBuilderRe
...
-      public string ExportToMarkdownString(object objectToExport)
+      public string ExportToMarkdownString(object objectToExport, int? indentationLevel = null)
       {
          var builder = _markdownBuilderRepository.BuilderFor(objectToExport);
          if (builder == null)
             return _reportGenerator.StringReportFor(objectToExport);
 
          var tracker = new MarkdownTracker();
-         builder.Report(objectToExport, tracker);
+         builder.Report(objectToExport, tracker, indentationLevel ?? 1);
```
msevestre: If it's not set, use the value of 1. otherwise, it will all be relative to this level

PR #1334, opened 2019-10-29
1333 update to version9

msevestre:


msevestre:
@Yuri05 
WIP. Do not merge

msevestre:
@Yuri05 Going to merge even though there is still one test failing due to either a test not configured propertly and or bug in Simmodel. I want to be able to run all itnegrations test to see how it behaves( and also create a zip file to see if the setup still works)

PR #1354, opened 2019-11-16
Fixes #1644 icon for individual

msevestre:


src/PKSim.Core/Model/IndividualFactory.cs
```diff
@@ -73,8 +73,8 @@ private Individual createStandardIndividual(OriginData originData, Func<IIndivid
          var individual = _objectBaseFactory.Create<Individual>();
          individual.OriginData = originData;
 
-         //default icon for an individual is the name of the species
-         individual.Icon = originData.Species.Name;
+         //default icon for an individual is the icon of the species
+         individual.Icon = originData.Species.Icon;
```
msevestre: @Yuri05 I am not sure why this was done like that. Assuming that we did not have an icon property at the beginning

src/PKSim.Presentation/Services/BuildingBlockIconRetriever.cs
```diff
@@ -71,10 +71,10 @@ public ApplicationIcon IconFor(ISimulationComparison simulationComparison)
 
       public ApplicationIcon IconFor(QualificationPlan qualificationPlan) => ApplicationIcons.Formula;
 
-      private string speciesNameFrom(UsedBuildingBlock usedBuildingBlock)
+      private string speciesIconFrom(UsedBuildingBlock usedBuildingBlock)
       {
          var individual = _buildingBlockRepository.ById(usedBuildingBlock.TemplateId) as Individual;
-         return individual == null ? usedBuildingBlock.BuildingBlockType.ToString() : individual.Species.Name;
+         return individual == null ? usedBuildingBlock.BuildingBlockType.ToString() : individual.Species.Icon;
```
Yuri05: why not just `individual.Icon` ?

msevestre: fair enough. I did not want to change too much. This should be exactly identical

msevestre:
@Yuri05 Going to merge and will update with your suggestion in next PT

PR #1358, opened 2019-11-18
Fixes #1348 remove tags

### LINKED ISSUE #1348: Cleanup tags tables
image

Seems to be a lot of useless stuff there (historically this was potentially required but not anymore I believe)

/cc @Yuri05 ?

msevestre:


src/Db/Dump/PKSimDB_dump.txt
```diff
@@ -90174,6 +89670,508 @@ INSERT INTO tab_calculation_method_rate_descriptor_conditions VALUES('DynamicSum
 INSERT INTO tab_calculation_method_rate_descriptor_conditions VALUES('DynamicSumFormulas','PARAM_Q_lng','Lung','NotInContainer');
 INSERT INTO tab_calculation_method_rate_descriptor_conditions VALUES('DynamicSumFormulas','PARAM_Q_lng','Pericentral','NotInContainer');
 INSERT INTO tab_calculation_method_rate_descriptor_conditions VALUES('DynamicSumFormulas','PARAM_Q_lng','Blood flow rate','Has');
+CREATE TABLE IF NOT EXISTS "tab_tags" (
```
Yuri05: "If not exists"? interesting...

src/Db/Dump/PKSimDB_dump.txt
```diff
@@ -90174,6 +89670,508 @@ INSERT INTO tab_calculation_method_rate_descriptor_conditions VALUES('DynamicSum
 INSERT INTO tab_calculation_method_rate_descriptor_conditions VALUES('DynamicSumFormulas','PARAM_Q_lng','Lung','NotInContainer');
 INSERT INTO tab_calculation_method_rate_descriptor_conditions VALUES('DynamicSumFormulas','PARAM_Q_lng','Pericentral','NotInContainer');
 INSERT INTO tab_calculation_method_rate_descriptor_conditions VALUES('DynamicSumFormulas','PARAM_Q_lng','Blood flow rate','Has');
+CREATE TABLE IF NOT EXISTS "tab_tags" (
+	"tag"	text NOT NULL,
+	PRIMARY KEY("tag")
+);
```
Yuri05: I would keep the description column, just no need for NOT NULL condition there

msevestre:
@Yuri05 DEscription was never used so far (always set to tag), I'd suggest to approve this PR and we can add a description column back that will be empty

Yuri05:
agree

PR #1365, opened 2019-11-29
1362 wrong aucr calculation

msevestre:


src/PKSim.Core/Services/ParameterUpdater.cs
```diff
@@ -153,5 +153,11 @@ private bool areValuesEqual(IParameter targetParameter, IParameter sourceParamet
       {
          return ValueComparer.AreValuesEqual(targetParameter, sourceParameter);
       }
+
+      private bool areDisplayUnitsEqual(IParameter targetParameter, IParameter sourceParameter)
+      {
+         //We compare name are some dose related units may be the representing the same unit but with different reference
+         return string.Equals(targetParameter.DisplayUnit?.Name, sourceParameter.DisplayUnit?.Name);
```
msevestre: @Yuri05 I updated this method here because I realzied that a bunch of commands were created for nothing. For some reason, we have different instances of the same unit (only for our home made units for dose). That way, we do not create stupid commands for nothing

src/PKSim.Core/Services/SimulationParametersUpdater.cs
```diff
@@ -26,10 +27,10 @@ public SimulationParametersUpdater(IParameterSetUpdater parameterSetUpdater, IEn
          _entityPathResolver = entityPathResolver;
       }
 
-      public ValidationResult ReconciliateSimulationParametersBetween(Simulation sourceSimulation, Simulation targetSimulation)
+      public ValidationResult ReconciliateSimulationParametersBetween(Simulation sourceSimulation, Simulation targetSimulation, PKSimBuildingBlockType buildingBlockType = PKSimBuildingBlockType.Simulation)
```
msevestre: This is really the only changes
We typically use this methods to copy ALL Simul;ation parameters from one simulation to another (after clone for example, where all buidlign blocks were in teh green state)
in this case however, we are using the same method but the BB do not have to be green, which leads to protocol being outdated

Protocol is actually the only bb that creates issue because of its 1 to n relation ship. All others are working fine.

src/PKSim.Core/Model/SimulationFactory.cs
```diff
@@ -208,12 +208,13 @@ public Simulation CreateForVSS(Protocol protocol, Individual individual, Compoun
...
-         var ddiRatioSimulation = createModelLessSimulationBasedOn(originalSimulation);
+         var ddiRatioSimulation =  createModelLessSimulationBasedOn(originalSimulation);
 
          _simulationModelCreator.CreateModelFor(ddiRatioSimulation);
 
-         //now update all parameters from the orginal simulation
-         _simulationParametersUpdater.ReconciliateSimulationParametersBetween(originalSimulation, ddiRatioSimulation);
+         // now update all parameters from the original simulation. We need to take protocol parameters as well as simulation parameters into account for the update
+         // as simulation building block are all kept in sync with the changes values EXCEPT Protocol (1 to n mapping)
+         _simulationParametersUpdater.ReconciliateSimulationParametersBetween(originalSimulation, ddiRatioSimulation, PKSimBuildingBlockType.Simulation | PKSimBuildingBlockType.Protocol) ;
```
msevestre: The real change is here in that case. We need to update Simulation and Protocol parameters

PR #1366, opened 2019-12-01
Fixes #1346 bloow flow

### LINKED ISSUE #1346: Blood flow rate => lung has generic parmaeter with Tags?
@Yuri05 Do you see a problem if I define a tag that would be implemented by all organs whose blood flow flows directly into lung?
That way we can use a sum formula, similar to body weight and have new organs (such as Thyroid) incorporated automatically

msevestre:


msevestre:
This file src/Db/Diff/PKSimDB_diff.txt  is really useless isn'it?

Yuri05:
I wouldn't say PKSimDB_diff.txt is useless.
Sometimes the diferences are easier to see here, sometimes in the full dump.
Also PKSimDB_diff.txt contains exactly the SQL statements which are needed to come from version N to version N+1 of the DB.

PR #1372, opened 2019-12-13
WIP #1371 R develop

msevestre:


src/PKSim.Infrastructure/InfrastructureRegister.cs
```diff
@@ -124,25 +129,50 @@ public static void RegisterSerializationDependencies()
          container.Register(typeof(IXmlReader<>), typeof(XmlReader<>));
          container.Register(typeof(IXmlWriter<>), typeof(XmlWriter<>));
 
-         //load repository to trigger initialization
+         //Core serialization to serialize to pkml compatible with core
+         container.AddRegister(x => x.FromType<CoreSerializerRegister>());
+      }
+
+      public static void LoadDefaultEntities(IContainer container)
```
msevestre: I had to reoranize all this stuff to make sure that I could change the path of the PKSim DB dynamically. SOmehow, the logic to load the db locally does not work because the path returned is the path from rClr. Anyways...it took me a whule to get everything working again but this is now ok

src/PKSim.Infrastructure/InfrastructureRegister.cs
```diff
@@ -124,25 +129,50 @@ public static void RegisterSerializationDependencies()
...
+      public static void RegisterWorkspace(IContainer container)
       {
-         var container = IoC.Container;
          container.Register<IWorkspace, IWithWorkspaceLayout, ICoreWorkspace, OSPSuite.Core.IWorkspace, Workspace>(LifeStyle.Singleton);
       }
 
-      private void registerORMDependencies()
+      private void registerORMDependencies(IContainer container)
       {
-         var container = IoC.Container;
```
msevestre: I cannot use IoC...at all. Using ospsuite-R already populates the global IoC and for some reason, PKSim is in the same context. Nothing was working when loading PKSim and OSPSuite because of the shared container (being even two dffieretn components as PKSim is using castle and OSPSuite.R is using autofac)

So now I am not using IoC at all in PKSim.R which allows me to create a new instance of the container

src/PKSim.Infrastructure/ORM/Mappers/FlatContainerIdToContainerMapperBase.cs
```diff
@@ -1,28 +1,18 @@
...
-      private readonly IObjectBaseFactory _objectBaseFactory;
-      private readonly IFlatContainerRepository _flatContainerRepository;
-      private readonly IFlatContainerTagRepository _flatContainerTagRepository;
-
+      protected readonly IObjectBaseFactory _objectBaseFactory;
+      protected readonly IFlatContainerRepository _flatContainerRepository;
+      protected readonly IFlatContainerTagRepository _flatContainerTagRepository;
       protected FlatContainer FlatContainer { get; private set; }
 
-      protected FlatContainerIdToContainerMapperBase()
```
msevestre: I had to delete the IoC usage. This was lazyness anyways

src/PKSim.R/Api.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace PKSim.R
+{
+   public static class Api
+   {
+      private static IContainer _container;
+
+      public static void InitializeOnce()
+      {
+         _container = ApplicationStartup.Initialize();
```
msevestre: SAving a static ref to the container which is NOT the IoC container

src/PKSim.R/Bootstrap/ApplicationStartup.cs
```diff
@@ -0,0 +1,87 @@
...
+            container.RegisterImplementationOf(new SynchronizationContext());
+            container.AddRegister(x => x.FromType<RRegister>());
+            container.AddRegister(x => x.FromType<CoreRegister>());
+            container.AddRegister(x => x.FromType<InfrastructureRegister>());
+            InfrastructureRegister.RegisterSerializationDependencies(container);
+            registerMinimalImplementations(container);
+         }
+
+         var configuration = container.Resolve<IPKSimConfiguration>();
+         configuration.PKSimDbPath = Path.Combine(new FileInfo(Assembly.GetAssembly(GetType()).Location).DirectoryName, CoreConstants.PK_SIM_DB_FILE);
```
msevestre: I have tried a milllion different ways but in short, starting PKSim.R from rClr returns the wrong local path so I have to overwrite it here

src/PKSim.R/Services/IndividualFactory.cs
```diff
@@ -0,0 +1,87 @@
...
+using OriginData = PKSim.Core.Snapshots.OriginData;
+
+namespace PKSim.R.Services
+{
+   public interface IIndividualFactory
+   {
+      /// <summary>
+      /// Returns the <seealso cref="ParameterValues"/> representing an individual with the characteristics defined in <paramref name="individualCharacteristics"/>.
+      /// </summary>
+      ParameterValue[] CreateIndividual(IndividualCharacteristics individualCharacteristics);
```
msevestre: @PavelBal This is where I need your input

@Yuri05 please comment
Any idea why we have two methods here? Shouldn/t create individual only return distributed parameter anyways?

Right now it returns everyparameter that is changed either directly or indirectly by create Individual .But then it returns also formula paraemter which we then try to filter out etc... Shouldn't we not do that at the first place and only return distributed parameters?

src/PKSim.R/Services/IndividualFactory.cs
```diff
@@ -0,0 +1,87 @@
...
+         var individualProperties = _individualValuesMapper.MapFrom(individual);
+         var allIndividualParameters = individualProperties.ParameterValues.ToList();
+         allIndividualParameters.AddRange(_ontogenyFactorsRetriever.FactorsFor(originData, moleculeOntogenies));
+         return allIndividualParameters.ToArray();
+      }
+
+      public DistributedParameterValue[] DistributionsFor(IndividualCharacteristics individualCharacteristics)
+      {
+         var originData = originDataFrom(individualCharacteristics.OriginData);
+         var moleculeOntogenies = individualCharacteristics.MoleculeOntogenies;
```
msevestre: This is not used at the moment

src/PKSim.R/Services/IndividualFactory.cs
PavelBal: > Any idea why we have two methods here? Shouldn/t create individual only return distributed parameter anyways?

I agree... I do not see a point in returning (explicit) formula parameters.

msevestre:
@Yuri05 Will merge. There is still the question of WHy do we return non distributed paramters from createIndividual at all but this can be clarified later on

PR #1383, opened 2020-01-20
1380 fix failing test

msevestre:


src/PKSim.Core/Services/RangeChartsDataCreator.cs
```diff
@@ -67,18 +67,23 @@ private void connectCurvesDefinedIn(ChartData<RangeXValue, RangeYValue> chart)
...
-            var firstCurve = orderedCurveByXvalue[i - 1];
-            var secondCurve = orderedCurveByXvalue[i];
-            firstCurve.Add(secondCurve.XValues[0], secondCurve.YValues[0]);
+            var firstCurve = orderedCurveByXValue[i - 1];
+            var secondCurve = orderedCurveByXValue[i];
+            // Only connect points if curves already belong to one and only that were split for some binning reasons. 
+            // e.g if the current X is after the next X, those curves clearly do not belong together
+            if (firstCurve.XValues.Last().X < secondCurve.XValues[0].X)
+            {
+              firstCurve.Add(secondCurve.XValues[0], secondCurve.YValues[0]);
```
msevestre: This is the change to fix the issue with the range completely messed up. In this case, if the last point of the first curve is before the first point of the second curve, we bring them together. I am not sure how to do it in a better way since there is no way to say whether curve are related or not (e.g. grouping by sim name or gender creates the same structure, but clearly only  one makes sense to connect)

src/PKSim.Infrastructure/ProjectConverter/v5_5/Converter551To552.cs
```diff
@@ -174,7 +174,10 @@ private void addSurfaceAreaVariabilityTo(Population population)
          //use a clone as this parameter will be used to create the advanced parameter
          var clone = _cloner.Clone(factor);
          var advancedParameter = new AdvancedParameter {DistributedParameter = clone, Seed = population.Seed};
-         var randomValues = advancedParameter.GenerateRandomValues(population.NumberOfItems);
+         //Because of changes in population serialization in v9, NumberOfItems in population is not set yet. Let's retrieve it from the internal structure
+         var numberOfIndividuals = population.IndividualValuesCache.AllParameterValues.FirstOrDefault()?.Count;
```
msevestre: I have fixed the test for now as it is possible to retrieve the number of individuals from the object structure

msevestre:
@Yuri05 Once CI is ok, I'll merge so that I can move with the next implementaiton

PR #1391, opened 2020-01-21
Fixes #1386 remove name from simulation individual chart

### LINKED ISSUE #1386: Remove simulation name from display path in individual simulation charts
At the moment default display path of a simulated curve always starts with a simulation name in individual simulations.
grafik
This was done to distinguish between curves in individual simulation **comparison** charts. Here, adding simulation name makes totally sense and must be kept as default . But not in a single simulation chart.

msevestre:


src/PKSim.Presentation/Presenters/Charts/ChartPresenter.cs
```diff
@@ -116,6 +118,11 @@ protected virtual void UpdateAnalysisBasedOn(IndividualSimulation simulation, Da
          }
       }
 
+      protected override string NameForColumn(DataColumn dataColumn)
+      {
+         return _chartPresenterContext.CurveNamer.CurveNameForColumn(SimulationFor(dataColumn), dataColumn, _useSimulationNameToCreateCurveName);
```
msevestre: it was not possible to override that behavior. Issue with some refactoring probably. Flag is true for comparison and false for individual simulation

PR #1399, opened 2020-02-04
1314 deactivated processes activated

msevestre:


msevestre:
So this was not a really big issue because the simulation itself was not affected. Just the configuration

src/PKSim.Core/Snapshots/CompoundProcessSelection.cs
```diff
@@ -1,14 +1,13 @@
...
 namespace PKSim.Core.Snapshots
 {
    public class CompoundProcessSelection : IWithName
    {
-      [Required]
       public string Name { get; set; }
       public string MoleculeName { get; set; }
       public string MetaboliteName { get; set; }
       public string CompoundName { get; set; }
+      public string SystemicProcessType { get; set; }
```
msevestre: Needed to add this property to be able to know what type of process was being saved. In the case of NO mapping, the name is also empty so I removed the required flag

src/PKSim.Core/Snapshots/Mappers/CompoundPropertiesMapper.cs
```diff
@@ -66,7 +66,7 @@ private ProtocolSelection protocolPropertiesFrom(ProtocolProperties protocolProp
 
       private Task<CompoundProcessSelection[]> snapshotProcessSelectionFrom(CompoundProcessesSelection compoundProcessesSelection)
       {
-         return _processMappingMapper.MapToSnapshots(compoundProcessesSelection.AllEnabledProcesses());
+         return _processMappingMapper.MapToSnapshots(compoundProcessesSelection.AllProcesses());
```
msevestre: Here we were only mapping the one activated. The SimulationConfiguration adds the missins enzymes by default so this was creating the mismatch on configure

src/PKSim.Core/Snapshots/Mappers/CompoundPropertiesMapper.cs
```diff
@@ -138,31 +139,61 @@ private void updateFormulationMapping(ProtocolProperties protocolProperties, For
...
+      private async Task<CompoundProcessesSelection> modelProcessSelectionFrom(CompoundProcessSelection[] snapshotProcesses, Model.Compound compound, ISimulationSubject simulationSubject)
       {
          var compoundProcessesSelection = new CompoundProcessesSelection();
          if (snapshotProcesses == null)
             return compoundProcessesSelection;
 
          foreach (var snapshotProcess in snapshotProcesses)
          {
-            var process = compound.ProcessByName(snapshotProcess.Name);
+            var process = compound.ProcessByName(snapshotProcess.Name) ?? notSelectedProcessFrom(snapshotProcess, simulationSubject);
```
msevestre: If the process does not exist, it's potentially a not mapped process.

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
```diff
@@ -248,36 +248,41 @@ public override async Task<ModelSimulation> MapToModel(SnapshotSimulation snapsh
...
             if (interaction != null)
                simulation.InteractionProperties.AddInteraction(interaction);
          }
       }
 
-      private async Task<InteractionSelection> interactionSelectionFrom(CompoundProcessSelection snapshotInteraction, PKSimProject project)
+      private async Task<InteractionSelection> interactionSelectionFrom(CompoundProcessSelection snapshotInteraction, ISimulationSubject simulationSubject, PKSimProject project)
       {
-         var process = findProcess(project, snapshotInteraction);
+         var process = findProcess(project, snapshotInteraction, simulationSubject);
```
msevestre: Same cause, same effect, same solution

PR #1401, opened 2020-02-05
Fixes #1320 Cannot confugure or clone simulation after renaming of Ad…

### LINKED ISSUE #1320: Cannot confugure or clone simulation after renaming of Administratio Protocol
If simulation has the same name as administration protocol used in it: after renaming of the administration protocol simulation cannot be reconfigured or cloned anymore.
grafik

To reproduce: in the project attached rename the administration protocol from `aa `to `bb `and try to reconfigure the simulation: error below pops up.
[P3.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/3639022/P3.zip)

grafik

somehow renaming was made in the wrong place (the correct path would be _**aa**|Applications|**bb**|..._)

msevestre:
…ministratio Protocol


msevestre:
@Yuri05 I got it after all. I had to rethink the concept completely instead of trying to rename part of the path

src/PKSim.Presentation/Services/RenameBuildingBlockTask.cs
```diff
@@ -205,14 +205,24 @@ private void renameFormulaPathReferencingContainerInSimulation(Simulation simula
...
+         {
+            //Reference is always found by construction
+            var referenceIndex = formula.ObjectReferences.Select(x => x.Object).IndexOf(reference);
+            var objectPath = formula.ObjectPaths[referenceIndex];
+
+            //The reference is used. The path needs to be updated only if it is a path referencing the old name
+            if(!objectPath.Contains(oldName))
+               continue;
+
+            objectPath.ReplaceWith(reference.EntityPath().ToPathArray());
```
msevestre: Here is the one change. I am swapping the path directly with a new one instead of messing around with rename

PR #1410, opened 2020-02-05
Fixes update for create pop in R

msevestre:


src/PKSim.R/Services/PopulationFactory.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            population.AddMolecule(molecule);
+         }
+
+         _ontogenyVariabilityUpdater.UpdateAllOntogenies(population);
+
+         return population.IndividualValuesCache;
+      }
+
+      private void validate(RandomPopulationSettings populationSettings)
```
msevestre: @Yuri05 I figured we should implement some kind of validation here. Otherwise ageMin > ageMax works :)

tests/PKSim.R.Tests/PopulationFactorySpecs.cs
```diff
@@ -0,0 +1,154 @@
...
+         var parameterPaths = _result.AllParameterPaths().ToArray();
+         foreach (var moleculeOntogeny in _populationCharacteristics.MoleculeOntogenies)
+         {
+            parameterPaths.ShouldContain($"{moleculeOntogeny.Molecule}|{CoreConstants.Parameters.ONTOGENY_FACTOR}");
+            parameterPaths.ShouldContain($"{moleculeOntogeny.Molecule}|{CoreConstants.Parameters.ONTOGENY_FACTOR_GI}");
+         }
+      }
+   }
+
+   public class When_creating_a_preterm_population_from_r : concern_for_PopulationFactory
```
msevestre: exactly same tests as for matlab, just with a slightly better interface

PR #1415, opened 2020-02-07
1407 export to cpp

msevestre:


src/PKSim.Core/Services/PopulationSimulationExportTask.cs
```diff
@@ -87,19 +82,14 @@ public void ExportForCluster(PopulationSimulation populationSimulation, string p
...
-         var outputDefinitionFileFullPath = Path.Combine(populationFolder, $"{fileName}{CoreConstants.Population.OUTPUT_DEFINITION_EXPORT}.csv");
 
          //Model
-         _simModelExporter.Export(_modelCoreSimulationMapper.MapFrom(populationSimulation, shouldCloneModel: false), modelFileFullPath);
-         // Outputs
-
-         var outputSelection = populationSimulation.OutputSelections;
-
-         exportOutputDefinition(outputSelection, outputDefinitionFileFullPath);
+         _mobiExportTask.ExportSimulationToPkmlFile(populationSimulation, modelFileFullPath);
```
msevestre: export pkml instead of xml for cluster

src/PKSim.Core/Services/PopulationSimulationExportTask.cs
```diff
@@ -109,17 +99,7 @@ public void ExportForCluster(PopulationSimulation populationSimulation, string p
          if (agingData.Rows.Count > 0)
             agingData.ExportToCSV(agingFileFullPath, comments: comments);
       }
-
-      private void exportOutputDefinition(OutputSelections outputSelections, string outputDefinitionFileFullPath)
```
msevestre: did not even remember we had something like this. Well this is gone now :)

src/PKSim.Infrastructure/InfrastructureRegister.cs
```diff
@@ -83,10 +83,11 @@ private static void registerRunOptionsIn(IContainer container)
          container.Register<StartOptions, IStartOptions, StartOptions>(LifeStyle.Singleton);
       }
 
-      private static IContainer initializeContainer()
+      private static IContainer initializeContainer(bool registerContainerAsStatic)
       {
          var container = new CastleWindsorContainer();
-         IoC.InitializeWith(container);
+         if (registerContainerAsStatic)
```
msevestre: static stuff registered for PKSim starrt (defualt) .But false for R so that we are not relying on any IoC stuff

src/PKSim.Infrastructure/Services/SimulationExportTask.cs
```diff
@@ -79,6 +79,18 @@ public Task ExportResultsToCSVAsync(Simulation simulation)
...
+      public Task ExportSimulationToCppAsync(Simulation simulation)
+      {
+         var outputFolder = _dialogCreator.AskForFolder(PKSimConstants.UI.ExportSimulationToCpp, Constants.DirectoryKey.SIM_MODEL_XML);
+         return string.IsNullOrEmpty(outputFolder) ? Task.CompletedTask : ExportSimulationToCppAsync(simulation, outputFolder);
+      }
+
+      public Task ExportSimulationToCppAsync(Simulation simulation, string outputFolder)
+      {
+         _lazyLoadTask.Load(simulation);
+         return Task.Run(() => _simModelExporter.ExportCppCode(_coreSimulationMapper.MapFrom(simulation, shouldCloneModel: false), outputFolder, FormulaExportMode.Formula));
```
msevestre: in crashes in core with the example, in PKsim, it does nothing.... I have no idea what's going on :)

Yuri05: @msevestre where is the test which crashes when exporting to c++?

msevestre: In core. It's commented out

msevestre: https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/tests/OSPSuite.Core.IntegrationTests/SimModelExporterSpecs.cs#L159

msevestre: in pksim, it does not do anything. no crash, no files nothing

PR #1417, opened 2020-02-10
Fixes #1290 snapshot parameters

### LINKED ISSUE #1290: Issue when exporting snapshot with changed building block parameters in simulations
If a project cotains simulations with changed building block parameters (red sign), thex export to json throws a warning.
However, actually it works properly. Also re-loading the snapshot works with the changed parameter. Only the "red sign" is missing (it shows the green sign.)

Please fix!

msevestre:


src/PKSim.Core/Snapshots/AlteredBuildingBlock.cs
```diff
@@ -0,0 +1,10 @@
+﻿using OSPSuite.Core.Domain;
+
+namespace PKSim.Core.Snapshots
+{
+   public class AlteredBuildingBlock
```
msevestre: I am saving a list of building block type and name that were altered in the simulation.

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
```diff
@@ -228,17 +242,28 @@ public override async Task<ModelSimulation> MapToModel(SnapshotSimulation snapsh
...
+         var registrationVisitor = new RegisterObjectVisitor(withIdRepository);
+
          var simulation = await createSimulationFrom(snapshot, project);
 
          simulation.Solver = await _solverSettingsMapper.MapToModel(snapshot.Solver);
          simulation.OutputSchema = await _outputSchemaMapper.MapToModel(snapshot.OutputSchema);
          simulation.OutputSelections = await _outputSelectionsMapper.MapToModel(snapshot.OutputSelections, simulation);
 
-         await updateParameters(simulation, snapshot.Parameters);
+         registrationVisitor.Register(simulation);
```
msevestre: I need to do this here so that I can fnd the corresponding buildign block parameter. (in the copy of the bb saved in the simulation)

src/PKSim.Core/Services/RegisterObjectVisitor.cs
```diff
@@ -45,12 +42,5 @@ public void Visit(IObjectBase objToVisit)
          _withIdRepository.Unregister(objToVisit.Id);
          register(objToVisit);
       }
-
-      public void Visit(Simulation simulation)
```
msevestre: @Yuri05 This is not necessary. The simulation is already performing the iterations. In essensce, all used building block where registered twice. Not a problem because register of object base did unregister but time consuming for nothing

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
```diff
@@ -107,9 +107,23 @@ public override async Task<SnapshotSimulation> MapToSnapshot(ModelSimulation sim
          snapshot.IndividualAnalyses = await _simulationTimeProfileChartMapper.MapToSnapshots(simulation.AnalysesOfType<SimulationTimeProfileChart>());
          snapshot.PopulationAnalyses = await _populationAnalysisChartMapper.MapToSnapshots(simulation.AnalysesOfType<ModelPopulationAnalysisChart>());
          snapshot.HasResults = simulation.HasResults;
+         snapshot.AlteredBuildingBlocks = alteredBuildingBlocksIn(simulation);
```
Yuri05: I hope when reading from old snapshot where `snapshot.AlteredBuildingBlocks` was not set - it will not crash?

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
```diff
@@ -378,9 +403,28 @@ private Task updateAdvancedParameters(ModelSimulation simulation, AdvancedParame
...
+         //The parameter mapper does not update the original building block in the simulation. We need to make sure we are keeping these values in sync so 
+         // that update and commit will work as expected (or building block difference )
+         var allParameters = _containerTask.CacheAllChildren<IParameter>(simulation.Model.Root);
+
+         snapshotParameters?.Each(x =>
+         {
+            var parameter = allParameters[x.Path];
+            //Somehow the localized parameter path was edited by hand or does not exist anymore. Warning was already shown during mapping. 
+            if (parameter == null)
+               return;
```
Yuri05: what is the use case when something like this can happen?

Yuri05: And shouldn't we produce a warning if this happens?

src/PKSim.Presentation/UICommands/ExportODEForRUICommand.cs
```diff
@@ -0,0 +1,23 @@
...
+      private readonly ISimulationExportTask _simulationExportTask;
+
+      public ExportODEForRUICommand(ISimulationExportTask simulationExportTask, IActiveSubjectRetriever activeSubjectRetriever) : base(activeSubjectRetriever)
+      {
+         _simulationExportTask = simulationExportTask;
+      }
+
+      protected override async void PerformExecute()
+      {
+         await _simulationExportTask.SecureAwait(x => x.ExportODEForR(Subject));
```
Yuri05: I would throw "not implemented" exception right here. At the moment user is asked for folder first, and after the folder selection the exception appears.

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
msevestre: Nope. But we'll find out for sure soon enough

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
msevestre: See the comment right above the selected line:
1/ happens when snapshot was modified by hand. It maybe the BB creation results in this parameter gone (e.g renamed etc)
2/ warning already shown earlier

For once .all questions answered the comment;-)

src/PKSim.Presentation/UICommands/ExportODEForRUICommand.cs
msevestre: That's ok like that.
 That's just for dev anyways.

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
Yuri05: ok

src/PKSim.Presentation/UICommands/ExportODEForRUICommand.cs
Yuri05: ok.
btw: just seen that MAtlab-ODE Export in Pk-Sim is also defined for devs only. Why?

msevestre: Beacuse Matlab is dead (and also noone ever missed in in PKSim). Good like that IMO until proven otherwise

PR #1418, opened 2020-02-10
Fixes #1330 export snapshot protein parameters

### LINKED ISSUE #1330: Export to snapshot / reimport in PK-Sim alters protein reference concentration
The attached PK-Sim file contains a European Standard individual with 2 CYP3A4 enzymes (`CYP3A4` and `CYP3A4 1`), both loaded from the gene expression database with a reference concentration of 4.32 µmol/L each.
When exporting the file to Snapshot (attached) and re-importing it in PK-Sim, the reference concentration of  `CYP3A4 1` is set to 1 µmol/L.

[Alprazolam.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/3766422/Alprazolam.zip)

msevestre:


src/PKSim.Core/Snapshots/Mappers/MoleculeMapper.cs
```diff
@@ -49,6 +49,13 @@ public override async Task<Molecule> MapToSnapshot(IndividualMolecule molecule)
          return snapshot;
       }
 
+      protected override bool ShouldExportParameterToSnapshot(IParameter parameter)
```
msevestre: @Yuri05 I always export the global parameters. This is very easy as we already had an extension method checking if a parameter was one of the global parameters

msevestre: This works ok but not perfect. Upon reload from snapshot, it looks like so for a molecule that was renamed

image

The default value is not saved in the snapshot. Thus when we reload from snapshot, default differs from current and it is shown as changed.  I am not sure if you have an idea how to circumvent this. 
I do not have anything that makes sense at the moment (Aside from checking if value origin is set but this seems like  a nasty hack)

Yuri05: Well, I can live with it. Since we have the possibility to map more than 1 process to an enzyme having something like `CYP3A4 `and `CYP3A4 1` at the same time doesn't make much sense anymore. Thus as long as the values itself are properly read from the snapshot, I wouldn't invest time trying to set "changed" correctly for such a case

PR #1420, opened 2020-02-11
Fixes #1291 Export/re-load of snapshot: transparency settings of colo…

### LINKED ISSUE #1291: Export/re-load of snapshot: transparency settings of colors is lost

msevestre:
…rs is lost

### DIFF
--- a/src/PKSim.Infrastructure/Serialization/Json/ColorConverter.cs
+++ b/src/PKSim.Infrastructure/Serialization/Json/ColorConverter.cs
@@ -13,7 +13,10 @@ namespace PKSim.Infrastructure.Serialization.Json
             return;
 
          var c = (Color) value;
-         var hexString = $"#{c.R:X2}{c.G:X2}{c.B:X2}";
+         var hexString = $"{c.R:X2}{c.G:X2}{c.B:X2}";
+         // Only add hex transparency if required as it does not seem to be widely supported by color readers.
+         var preFix = c.A == 255 ? "#" : $"#{c.A:X2}";
+         hexString = $"{preFix}{hexString}";
          writer.WriteValue(hexString);
       }
 
@@ -23,7 +26,7 @@ namespace PKSim.Infrastructure.Serialization.Json
          return existingValue;
       }
 
-      public override bool CanRead => false;
+      public override bool CanRead => true;
 
       public override bool CanConvert(Type objectType)
       {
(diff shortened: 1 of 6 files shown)


tests/PKSim.Tests/Infrastructure/SnapshotSerializerSpecs.cs
```diff
@@ -106,4 +106,32 @@ public void should_have_deserialized_the_color()
...
+      private CurveOptions _curveOptions;
+      private CurveOptions _deserializedCurveOptions;
+
+      protected override async Task Context()
+      {
+
+         await base.Context();
+         _curveOptions = new CurveOptions
+         {
+            Color = Color.FromArgb(20, Color.CadetBlue)
```
msevestre: This is the real test here. That works now :)

src/PKSim.Infrastructure/Serialization/Json/ColorConverter.cs
```diff
@@ -13,7 +13,10 @@ public override void WriteJson(JsonWriter writer, object value, Newtonsoft.Json.
             return;
 
          var c = (Color) value;
-         var hexString = $"#{c.R:X2}{c.G:X2}{c.B:X2}";
+         var hexString = $"{c.R:X2}{c.G:X2}{c.B:X2}";
+         // Only add hex transparency if required as it does not seem to be widely supported by color readers.
+         var preFix = c.A == 255 ? "#" : $"#{c.A:X2}";
+         hexString = $"{preFix}{hexString}";
```
Yuri05: well, no idea what this code is doing. So you save the color info as #AARRGGBB instead of #RRGGBB and AA is the transparency info?

msevestre: Yep. It creates an hex strings (it was a request that colors are hex strings if you remember)

But usually transparency is not taken into consideration. That's why I only add it if it's been set by the user so that the hex code is not always weird...

msevestre: Here also write in the comment;-)

Yuri05: ok. I assume that this works as expected (never seen transparency info in the rgb hex strings) :)

msevestre: Me too. I didn't try it but the test looks good

PR #1426, opened 2020-02-12
Fixes #1424 ensure qualification run

### LINKED ISSUE #1424: Ensure that qualification run works after redesign

msevestre:


src/PKSim.Infrastructure/ProjectConverter/v9_0/Converter80To90.cs
```diff
@@ -135,5 +137,16 @@ private void convertIndividualValueCache(IndividualValuesCache individualValuesC
          individualValuesCache.IndividualIds.AddRange(Enumerable.Range(0, firstParameterValue.Count));
          _converted = true;
       }
+
+      public void Visit(PopulationAnalysisChart populationAnalysisChart)
```
msevestre: All right. We have renamed the race into population. Analysis referencing race however are not happy. So here we go:)

tests/PKSim.Tests/ProjectConverter/v9_0/Converter80To90Specs.cs
```diff
@@ -52,4 +54,31 @@ public void should_have_converted_all_population_building_blocks_defined_in_simu
...
    }
+
+   public class When_converting_the_population_analysis_74_project_to_90 : ContextWithLoadedProject<Converter80To90>
+   {
+      private List<PopulationSimulation> _allSimulations;
+
+      public override void GlobalContext()
+      {
+         base.GlobalContext();
+         LoadProject("PopulationAnalyses_740");
```
msevestre: @Yuri05 This is the project from #1424. TDD implementation and then I also tried it so pretty sure it works

PR #1434, opened 2020-02-17
1430 observed data errir

msevestre:


msevestre:
@Yuri05 Pretty hard core bug here. Conversion to YAxis value was done using the merged dimension of the YAxis...which was created using the first output. No problem until we have multiple outputs with different molweight and we are displaying values in mg/l. In that case, the wrong molweight was used.

msevestre:
This is a WIP so that you can review my changes. I need to write more tests and fix failing tests

src/PKSim.Core/Chart/BoxWhiskerValue.cs
```diff
@@ -2,66 +2,68 @@
...
+            objectWithTargetUnit.DisplayValueWithUnit(LowerWhisker.Value, valueDimension),
             LowerWhisker.IndividualId,
-            unitConverter.DisplayValue(LowerBox.Value),
+            objectWithTargetUnit.DisplayValueWithUnit(LowerBox.Value, valueDimension),
             LowerBox.IndividualId,
-            unitConverter.DisplayValue(Median.Value),
+            objectWithTargetUnit.DisplayValueWithUnit(Median.Value, valueDimension),
             Median.IndividualId,
-            unitConverter.DisplayValue(UpperBox.Value),
+            objectWithTargetUnit.DisplayValueWithUnit(UpperBox.Value, valueDimension),
```
msevestre: I am now passing the valueDimension which is the unit in which the value is saved. In most case, this will be the same as the Axis dimension but in our case, it could be different

src/PKSim.Core/Chart/CurveData.cs
```diff
@@ -53,6 +54,11 @@ public class CurveData<TX, TY> : ICurveDataSettings
       /// </summary>
       public virtual string QuantityPath { get; set; }
 
+      /// <summary>
+      /// Dimension in which the x values are stored.
```
Yuri05: **y** values I guess ;)

src/PKSim.Core/Extensions/WithDisplayUnitExtensions.cs
```diff
@@ -1,15 +1,23 @@
...
       {
-         return _doubleFormatter.Format(withDisplayUnit.ConvertToDisplayUnit(valueInCoreUnit), withDisplayUnit.DisplayUnit);
+         return _doubleFormatter.Format(DisplayValue(withDisplayUnit, valueInBaseUnit, valueDimension), withDisplayUnit.DisplayUnit);
+      }
+
+      public static double DisplayValue(this IWithDisplayUnit withDisplayUnit, double valueInBaseUnit, IDimension valueDimension = null)
+      {
+         var dimensionToUse = valueDimension ?? withDisplayUnit.Dimension;
+         //This value now needs to be converted in the display value of the target unit
+         return dimensionToUse.BaseUnitValueToUnitValue(withDisplayUnit.DisplayUnit, valueInBaseUnit);
```
msevestre: This is the key here. I am using the dimension to convert to the axis display dimension as opposed to using the dmension of the axis

src/PKSim.Core/Services/SimulationPKParametersImportTask.cs
```diff
@@ -14,42 +16,26 @@ public interface ISimulationPKParametersImportTask
...
       {
-         var importedPKAnalysis = await importPKAnalysesFromFile(fileFullPath, cancellationToken);
+         var importedPKAnalysis = await _coreSimulationPKParametersImportTask.ImportPKParameters(fileFullPath, populationSimulation, cancellationToken);
+         cancellationToken.ThrowIfCancellationRequested();
          validateConsistencyWithSimulation(populationSimulation, importedPKAnalysis);
-         addImportedPKToLogForSuccessfulImport(importedPKAnalysis);
          return importedPKAnalysis;
       }
 
-      private void addImportedPKToLogForSuccessfulImport(SimulationPKParametersImport pkParameterImport)
```
msevestre: This was also put to core and was consolidated when updated to latest core version. Unrelated to the changes here

src/PKSim.UI/Binders/ChartDataBinder.cs
```diff
@@ -447,7 +448,7 @@ protected void AssignCurveSeriesList(IEnumerable<Series> seriesList, XYDiagramPa
             var xValue = xAxis.ConvertToDisplayUnit(curveData.XValues[i].X);
             var yValues = yProperty(curveData.YValues[i]);
 
-            double[] yConvertedValues = yValues.Select(y => yAxis.ConvertToDisplayUnit(y)).ToArray();
+            double[] yConvertedValues = yValues.Select(y => yAxis.DisplayValue(y, curveData.YDimension)).ToArray();
```
msevestre: Other key change. Instead of using the yAxis to convert, I am using the YDimension of the curve to convert in the displayh unit of the yAxis.

src/PKSim.Core/Chart/CurveData.cs
msevestre: indeed. Fixed

msevestre:
@Yuri05 This should be good to go. Please review when you can

PR #1441, opened 2020-02-20
Fixes #1436 height in aging simulation

### LINKED ISSUE #1436: Height in aging simulations
Height is not properly scaled in aging simulations

msevestre:


src/PKSim.Core/Services/DistributedParameterToTableParameterConverter.cs
```diff
@@ -155,26 +158,35 @@ private void createSpatialStructureTableParameters(IBuildConfiguration buildConf
...
+            createSpatialStructureTableParameter(structureParameter, individualParameter, allDistributionsForMaleParameter, allDistributionsForFemaleParameter, buildConfiguration);
 
-            createPopulationTableParameter(baseIndividualParameter, simulationPopulation, allDistributionsForMaleParameter, allDistributionsForFemaleParameter);
+            createPopulationTableParameter(individualParameterPath, individualParameter, populationSimulation, allDistributionsForMaleParameter, allDistributionsForFemaleParameter);
          }
+
+         //Height parameter is not a distributed parameter. However, we need to define this parameter as table to ensure that height dependent parameters are updated properly
+         var heightParameter = spatialStructure.TopContainers.FindByName(Constants.ORGANISM).Parameter(CoreConstants.Parameters.HEIGHT);
+         var individualMeanHeightParameter = _baseIndividual.Organism.EntityAt<IDistributedParameter>(CoreConstants.Parameters.MEAN_HEIGHT);
+         createSpatialStructureTableParameter(heightParameter, individualMeanHeightParameter, _allHeightDistributionMaleParameters, _allHeightDistributionMaleParameters, buildConfiguration);
```
msevestre: @Yuri05 I am adding now a table parameter for the height parameter as well as an entry in the aging table

src/PKSim.Core/Services/DistributedParameterToTableParameterConverter.cs
```diff
@@ -275,20 +290,61 @@ private bool parameterShouldBeDefinedAsTable(IDistributedParameter parameter)
...
+         addAgingDataToPopulationSimulation(populationSimulation, individualParameterPath, individualParameter,
             p =>
             {
-               var distributions = allDistributionsWithAgeStrictBiggerThanOriginData(allDistributionsForMaleParameter, allDistributionsForFemaleParameter, p.OriginData).ToList();
+               var distributions = allDistributionsWithAgeStrictBiggerThanOriginData(allDistributionsForMaleParameter, allDistributionsForFemaleParameter, p.OriginData);
                return createTableFormulaFrom(p, distributions);
             });
       }
 
+      private void createPopulationHeightTableParameter(
```
msevestre: I could not use the deafult method because the height parmaeter is not distributed. I could have messed around with the original function but its' complicated enough already so I separated the tasks

src/PKSim.Core/Services/DistributedParameterToTableParameterConverter.cs
```diff
@@ -275,20 +290,61 @@ private bool parameterShouldBeDefinedAsTable(IDistributedParameter parameter)
...
+         {
+            //create origin data for individual i
+            originData.Age = allAges[individualIndex];
+            originData.GestationalAge = allGAs[individualIndex];
+            originData.Height = allHeights[individualIndex];
+            originData.Gender = allGender[individualIndex];
+            tableFormulaParameter.Value = originData.Height.Value;
+
+            var heightDistributions = heightDistributionsFor(originData);
+            tableFormulaParameter.Percentile = heightDistributions.currentPercentile;
```
msevestre: REquired to get the percentile of the current HEIGHT so that we can ensure perfect distributions scaling

src/PKSim.Core/Services/DistributedParameterToTableParameterConverter.cs
Yuri05: createSpatialStructureTableParameter(heightParameter, individualMeanHeightParameter, _allHeightDistribution**Male**Parameters, _allHeightDistribution**Male**Parameters, buildConfiguration);

one of them should be female I guess

msevestre: good catch!

msevestre: fixed

src/PKSim.Core/Services/DistributedParameterToTableParameterConverter.cs
```diff
@@ -275,20 +290,61 @@ private bool parameterShouldBeDefinedAsTable(IDistributedParameter parameter)
...
       }
 
+      private void createPopulationHeightTableParameter(
+         string heightParameterPath,
+         IDistributedParameter meanHeightParameter,
+         PopulationSimulation populationSimulation
+      )
+      {
+         if (populationSimulation == null || !_createAgingDataInSimulation)
+            return;
```
Yuri05: As far as I understand `_createAgingDataInSimulation `is responsible for creation of Table parameters in **population** simulations. 
I would rename it then to something like `_createAgingDataInPopulationSimulation`. And also rename the corresponding parameter in all calling functions:
* here in `UpdateBuildConfigurationForAging`
* in `BuildConfigurationTask.CreateFor`

Otherwise it's highly confusing.

Yuri05: apart from that: looks good for me :)

msevestre: I had renamed the variable in the interface already
image

But not th elocal one. I'll do it and then merge

PR #1454, opened 2020-03-24
Update core and add pk conversion

msevestre:


src/PKSim.UI/Views/Core/BaseUserControlWithValueInGrid.cs
```diff
@@ -39,6 +49,30 @@ protected virtual void OnGridViewMouseDown(UxGridView gridView, MouseEventArgs e
          OnValueColumnMouseDown(gridView, col, rowHandle);
       }
 
+      protected virtual void RaiseButtonClick(UxGridView gridView, GridColumn column, int rowHandle, MouseEventArgs e)
```
msevestre: @Yuri05 THis is an issue I found as part of the Ranorex test. After the update to new devexpress, button clicked was not always raised.... oh well

PR #1464, opened 2020-04-22
Fixes #1462 delete comparisons

### LINKED ISSUE #1462: Context menu "Delete all subfolders and data" for "Comparisons" without function.
Right click on "Comparisons" in the "Simulations"-Explorer -> "Delete all subfolders and data" is without any function.

msevestre:


src/PKSim.Presentation/Presenters/Main/SimulationExplorerPresenter.cs
```diff
@@ -72,10 +85,17 @@ public override bool RemoveDataUnderClassification(ITreeNode<IClassification> cl
       {
          if (classificationNode.Tag.ClassificationType == ClassificationType.Simulation)
          {
-            IReadOnlyList<Simulation> allSimulations = classificationNode.AllNodes<SimulationNode>().Select(x => x.Tag.Simulation).ToList();
+            var allSimulations = classificationNode.AllNodes<SimulationNode>().Select(x => x.Tag.Simulation).ToList();
             return _buildingBlockTask.Delete(allSimulations);
          }
 
+         if (classificationNode.Tag.ClassificationType == ClassificationType.Comparison)
```
msevestre, line 92: @Yuri05 This was always missing. e.g. never worked

src/PKSim.Presentation/Services/SimulationComparisonTask.cs
```diff
@@ -92,5 +104,27 @@ public void ConfigurePopulationSimulationComparison(PopulationSimulationComparis
             _singleStartPresenterTask.StartForSubject(simulationComparison);
          }
       }
+
+      public bool Delete(IReadOnlyList<ISimulationComparison> simulationComparisons)
```
msevestre, line 108: The code was in the command. It's now in a task and the command calls the task. That way we can call the same method from the context menu as well

PR #1490, opened 2020-05-22
1359 observers snapshot

msevestre:


msevestre:
Observer snapshot support in PK-Sim... woot woot

src/PKSim.Core/Snapshots/Mappers/DescriptorConditionMapper.cs
```diff
@@ -0,0 +1,77 @@
...
+               return new InContainerCondition(tag);
+            case Constants.ALL_TAG:
+               return new MatchAllCondition();
+            case MATCH_TAG:
+               return new MatchTagCondition(tag);
+            case Constants.NOT_IN_CONTAINER:
+               return new NotInContainerCondition(tag);
+            case Constants.NOT:
+               return new NotMatchTagCondition(tag);
+         }
```
Yuri05, line 50: Missing warning for the "default: " path

src/PKSim.Core/Snapshots/Mappers/DescriptorConditionMapper.cs
```diff
@@ -0,0 +1,77 @@
...
+               return Constants.IN_CONTAINER;
+            case MatchAllCondition _:
+               return Constants.ALL_TAG;
+            case MatchTagCondition _:
+               return MATCH_TAG;
+            case NotInContainerCondition _:
+               return Constants.NOT_IN_CONTAINER;
+            case NotMatchTagCondition _:
+               return Constants.NOT;
+         }
```
Yuri05, line 71: missing exception for the "default:" path

src/PKSim.Core/Snapshots/Mappers/ObserverMapper.cs
```diff
@@ -0,0 +1,105 @@
...
+      }
+
+      public override async Task<SnapshotObserver> MapToSnapshot(ModelObserver observer)
+      {
+         var snapshot = await SnapshotFrom(observer);
+         snapshot.Dimension = SnapshotValueFor(observer.Dimension?.Name);
+         snapshot.ContainerCriteria = await _descriptorConditionMapper.MapToSnapshots(observer.ContainerCriteria);
+
+         //We do not support any other type of formula at the moment
+         snapshot.Formula = await _explicitFormulaMapper.MapToSnapshot(observer.Formula as OSPSuite.Core.Domain.Formulas.ExplicitFormula);
```
Yuri05, line 51: what will happen then in case of other formula types (table etc.)? MapToSnapshot will crash then?

src/PKSim.Core/Snapshots/Mappers/DescriptorConditionMapper.cs
msevestre: disagree. This is treated in the rest of the method . Default does not make sense if you return everywhere

src/PKSim.Core/Snapshots/Mappers/DescriptorConditionMapper.cs
msevestre: disagree. This is treated in the rest of the method . Default does not make sense if you return everywhere

src/PKSim.Core/Snapshots/Mappers/ObserverMapper.cs
msevestre: Yes and I hope it will crash since we only support ExplicitFormula. We need to catch it as a BIG fat exception

src/PKSim.Core/Snapshots/Mappers/DescriptorConditionMapper.cs
Yuri05: right. missed that part.

src/PKSim.Core/Snapshots/Mappers/DescriptorConditionMapper.cs
Yuri05: same.

PR #1507, opened 2020-06-18
1465 remove warning

abdelr:


src/PKSim.CLI.Core/PKSim.CLI.Core.csproj
```diff
@@ -29,6 +29,8 @@
   <ItemGroup>
     <PackageReference Include="OSPSuite.Utility" Version="4.0.0.4" />
     <PackageReference Include="OSPSuite.Core" Version="9.0.128" />
+    <PackageReference Include="Serilog.Extensions.Logging" Version="3.0.1" />
+    <PackageReference Include="Serilog.Extensions.Logging.File" Version="2.0.0" />
```
msevestre, line 33: @abdelr Wondering why the core requires a reference. Isn't is something the caller is referencing and adding to the factory?

src/PKSim.CLI/Program.cs
```diff
@@ -5,74 +5,82 @@
...
 
 namespace PKSim.CLI
 {
-   [Flags]
-   enum ExitCodes
-   {
-      Success = 0,
-      Error = 1 << 0,
-   }
+  [Flags]
```
msevestre, line 15: @abdelr Looks like you are using 2 indent instead of 3
https://github.com/Open-Systems-Pharmacology/Suite/blob/develop/CODING_STANDARDS.md

src/PKSim.CLI/Program.cs
```diff
@@ -5,74 +5,82 @@
...
 
-         if (!string.IsNullOrEmpty(runCommand.LogFileFullPath))
-            loggerFactory.AddFile(runCommand.LogFileFullPath, runCommand.LogLevel, runCommand.AppendToLog);
+    private static IOSPLogger initializeLogger(CLICommand runCommand)
+    {
 
-         return IoC.Resolve<ILogger>();
-      }
-   }
+      var loggerCreator = IoC.Resolve<ILoggerCreator>();
```
msevestre, line 70: This is werid. LoggerCreator is just created locally and then not used.  
Should we rename it to something like LoggerConfigurator or sthg like this?

src/PKSim.Infrastructure/Services/ILoggerCreator.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace PKSim.Infrastructure.Services
```
msevestre, line 8: That really looks like something that should be in core

src/PKSim.Infrastructure/Services/PKSimLoggerCreator.cs
```diff
@@ -0,0 +1,44 @@
+﻿using PKSim.Infrastructure.Services;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Console;
+using NHibernate.Mapping;
```
msevestre, line 4: not sure why this is using NHiberate, Console etc..

src/PKSim.Infrastructure/Services/PKSimLoggerCreator.cs
```diff
@@ -0,0 +1,44 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using PKSim.Infrastructure.Extensions;
+using System.Collections.Concurrent;
+
+public class PKSimLoggerCreator : ILoggerCreator
```
msevestre, line 12: This can be in Core

src/PKSim.Infrastructure/Services/PresenterLogger.cs
```diff
@@ -13,20 +13,15 @@ public class PresenterLogger : ILogger
    {
       private readonly IEventPublisher _eventPublisher;
       public string Name { get; }
-      private readonly LogLevel _logLevel;
 
-      public PresenterLogger(string name, LogLevel logLevel)
+      public PresenterLogger(string name)
       {
          _eventPublisher = IoC.Resolve<IEventPublisher>();
-         _logLevel = logLevel;
```
msevestre, line 21: Why was it removed?

src/PKSim.Infrastructure/Services/PresenterLogger.cs
```diff
@@ -45,12 +40,12 @@ private void writeMessage(LogLevel logLevel, string name, int eventIdId, string
 
       public bool IsEnabled(LogLevel logLevel)
       {
-         return logLevel >= _logLevel;
+         return true;
```
msevestre, line 43: That looks wrong here. This is a way to filter what is being logged to the view. Unless this was never used..

src/PKSim.CLI.Core/PKSim.CLI.Core.csproj
abdelr: @msevestre @georgeDaskalakis Indeed it is not required. Maybe we add this at some experimental step. I just removed it and will be updated in the next commit.

src/PKSim.CLI/Program.cs
abdelr: @msevestre Corrected already

src/PKSim.CLI/Program.cs
abdelr: @msevestre It is used. It will be injected to the PKSimLogger and it should be a singleton. We might have forgotten to register it as a singleton though for the contract used in this case. Maybe @georgeDaskalakis could take a look into this? Otherwise I'll do it. About naming, well we could rename it but it is not just configuring loggers, it is actually creating the loggers and keeping them in a dictionary based on the category. We are not keeping the LoggerFactory, nor the LoggingBuilder alive anymore since this is not the way to go in the new version. Instead we keep the loggers alive while the app is running. This loggerCreator has then a ConcurrentDictionary from category names to loggers and also knows how to create new loggers in case of first use.

src/PKSim.Infrastructure/Services/ILoggerCreator.cs
abdelr: @msevestre @georgeDaskalakis I agree. We should move it when we are able to create new versions of our nugets so we could use them in MoBi as well.

src/PKSim.CLI/Program.cs
abdelr: @msevestre @georgeDaskalakis About _"Maybe @georgeDaskalakis could take a look into this? Otherwise I'll do it"_ it is already registered as a singleton in PKSim.Infrastructure.InfrastructureRegister

src/PKSim.Infrastructure/Services/PKSimLoggerCreator.cs
abdelr: @msevestre @georgeDaskalakis I have just removed those. We do not need them.

src/PKSim.Infrastructure/Services/PKSimLoggerCreator.cs
abdelr: @msevestre @georgeDaskalakis I agree. We should move it when we are able to create new versions of our nugets so we could use them in MoBi as well.

src/PKSim.Infrastructure/Services/PresenterLogger.cs
abdelr: @msevestre @georgeDaskalakis We do not need to keep track of this anymore. By setting the minimum log level with builder.SetMinimumLevel when configuring the LoggingBuilder we ensure the log method will not be called at all. We could still use the logLevel for filtering when in combination to something more but that is not the case now so we remove it for now and the configuration step will also be smoother.

src/PKSim.Infrastructure/Services/PresenterLogger.cs
abdelr: @msevestre @georgeDaskalakis We do not need to keep track of this anymore. By setting the minimum log level with builder.SetMinimumLevel when configuring the LoggingBuilder we ensure the log method will not be called at all. We could still use the logLevel for filtering when in combination to something more but that is not the case now so we remove it for now and the configuration step will also be smoother.

msevestre:
@abdelr @georgeDaskalakis I have also updated develop in PKSim to use v10 (renamed version, build ec..) 
so you can create a PR here as well whenever you feel like it. Don't forget to rebase/merge on develop

tests/PKSim.Tests/Infrastructure/PopulationPKAnalysesTaskSpecs.cs
```diff
@@ -105,6 +105,8 @@ public void should_have_created_one_pk_analyses_for_each_output_defined_in_the_s
          _results.AllPKParametersFor(_quantityPath2).Length.ShouldBeEqualTo(2);
       }
 
+      // TODO this has to be corrected and reactivated
```
msevestre, line 108: Remove

src/PKSim.CLI/Services/LoggingBuilderExtensions.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Serilog;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.Logging
```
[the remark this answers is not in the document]
abdelr: Indeed... It was already moved to Core but we forgot to remove it from here... Done with the next commit

src/PKSim.UI/BootStrapping/ApplicationStartup.cs
```diff
@@ -82,10 +83,15 @@ public void InitializeForStartup(LogLevel logLevel, Action<IContainer> registrat
 
       private void configureLogger(IContainer container, LogLevel logLevel)
       {
-         var loggerFactory = container.Resolve<ILoggerFactory>();
-         loggerFactory
-            .AddPresenter(logLevel)
-            .AddDebug(logLevel);
+         var loggerCreator = IoC.Resolve<ILoggerCreator>();
```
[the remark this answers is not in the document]
abdelr: Agree

abdelr:
> The two pksin files should NOT be changed. THis happens because you ran the tests locally and they were updated
> tests/PKSim.Tests/Data/SimpleSimulationIV_710.pksim5
> tests/PKSim.Tests/Data/Theophylline_612.pksim5
> 
> Also do not use IoC unless you really have to.
> Last, some code should be moved to core. This can be done later but we should not forget

These are the files I told you about some days ago. I am not sure if we should have them in the gitignore list. Right now every time I run tests they change and if I am not cautious they will be replaced like with this commit. Should we look for a better solution here?

msevestre:
@abdelr IF you have a better solution yes. 
It's only an issue when one is running ALL the tests (E.g. even the project conversion tests automatically)

abdelr:
What about this:
https://github.com/Open-Systems-Pharmacology/build-scripts/pull/16

PR #1518, opened 2020-06-30
Update PopulationSettingsDTO.cs

abdelr:
Fixed

### DIFF
--- a/tests/PKSim.Tests/Presentation/PopulationSettingsDTOSpecs.cs
+++ b/tests/PKSim.Tests/Presentation/PopulationSettingsDTOSpecs.cs
@@ -1,0 +1,30 @@
+﻿using NUnit.Framework;
+using OSPSuite.BDDHelper;
+using PKSim.Presentation.DTO.Populations;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace PKSim.Presentation
+{
+   public class concern_for_PopulationSettingsDTO : ContextSpecification<PopulationSettingsDTO>
+   {
+      protected override void Context()
+      {
+         sut = new PopulationSettingsDTO();
+         sut.NumberOfIndividuals = 5;
+         sut.ProportionOfFemales = 50;
+      }
+
+      [TestCase(50, true)]
+      [TestCase(101, false)]
+      [TestCase(-1, false)]
+      public void FemaleRate_should_be_properly_bounded(int femaleRate, bool expectation)
+      {
+         sut.ProportionOfFemales = femaleRate;
+         Assert.AreEqual(expectation, sut.Rules.All().All(r => r.IsSatisfiedBy(sut)));
+      }
+   }
+}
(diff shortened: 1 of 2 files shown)


abdelr:
@msevestre I think it is very imperative to keep it this way. We add tests for every bug fix. I just added some very simple specs. Please let me know what you think.

msevestre:
@abdelr I agree. That was my passive-aggressive way to ask you to do so :)

tests/PKSim.Tests/Presentation/PopulationSettingsDTOSpecs.cs
```diff
@@ -0,0 +1,30 @@
...
+         sut.ProportionOfFemales = 50;
+      }
+
+      [TestCase(50, true)]
+      [TestCase(101, false)]
+      [TestCase(-1, false)]
+      public void FemaleRate_should_be_properly_bounded(int femaleRate, bool expectation)
+      {
+         sut.ProportionOfFemales = femaleRate;
+         Assert.AreEqual(expectation, sut.Rules.All().All(r => r.IsSatisfiedBy(sut)));
```
msevestre, line 27: Can you use BDD helper here?
Also you can probably simplify this a lot by using sut.IsValid() or sthg like this.

abdelr: > @abdelr I agree. That was my passive-aggressive way to ask you to do so :)

@msevestre Well, do not hesitate to be just aggressive ;). I mean seriously, I will never take a feedback the wrong way. We are learning a lot from your feedback so everything you think just say it. So far, I have felt like every single comment is in the direction to keep quality high and that is good, specially for us, I would not like have a Frankenstein we cannot manage anymore in a few month...

I have just change to using IsValid to both, simplify the call but also make sure to use the code as it is intended to. I do not know how you would have me using the BDD here. If I use because method then I have to write three different tests which is right now the case but in a very compact and readable way. Is this what you mean? I could also imagine it could be for substituting the TestCase decorator but I am not sure this is the intention of the comment.

tests/PKSim.Tests/Presentation/PopulationSettingsDTOSpecs.cs
```diff
@@ -0,0 +1,31 @@
...
+      }
+
+      [TestCase(50, true)]
+      [TestCase(101, false)]
+      [TestCase(-1, false)]
+      public void FemaleRate_should_be_properly_bounded(int femaleRate, bool expectation)
+      {
+         sut.ProportionOfFemales = femaleRate;
+         Assert.AreEqual(expectation, sut.IsValid());
+      }
```
msevestre, line 29: Sorry I meant sut.IsVali().shouldbbeequaldTO(expectation)

We just almost never used Assert.AreEqual. But this is fine for me as well

abdelr: @msevestre I see, no you are right... I am changing it ;)

Issue #1568, opened 2020-10-08
Project Conversion : Relative Expression Redesign


msevestre:
## Individual

- [x] Add f_endo parameter in all organ and liver zones
- [x] Convert molecules location to new localization

msevestre:
@Yuri05 @PavelBal 
How do we convert or old Localization 

![image](https://user-images.githubusercontent.com/1041237/97727076-56e9fa80-1aa6-11eb-82e7-58397978a5ff.png)

to our new one?

As a reminder
![image](https://user-images.githubusercontent.com/1041237/97727174-7bde6d80-1aa6-11eb-9b85-b3e87a9c5d8b.png)

![image](https://user-images.githubusercontent.com/1041237/97727186-813bb800-1aa6-11eb-92e0-404400dba38b.png)

![image](https://user-images.githubusercontent.com/1041237/97727209-88fb5c80-1aa6-11eb-9b66-c0c141fe7988.png)

I need this asap :)

Yuri05:
Tissue | VascEndo | f_expr_intracellular | f_expr_interstitial | f_expr_BC_intracellular | f_expr_BC_interstitial | f_expr_VascEndo_Endosome | f_expr_VascEndo_Plasma | f_expr_VascEndo_Tissue
-- | -- | -- | -- | -- | -- | -- | -- | --
Intracellular | Endosomal | 1 | 0 | 1 | 0 | 1 | 0 | 0
Intracellular | Interstitial | 1 | 0 | 1 | 0 | 0 | 0 | 1
Extracellular   Membrane | Apical | 0 | 1 | 0 | 1 | 0 | 1 | 0
Extracellular   Membrane | Basolateral | 0 | 1 | 0 | 1 | 0 | 0 | 1
Interstitial |   | 0 | 1 | 1 | 0 | 0 | 0 | 1

@PavelBal Please check

msevestre:
@Yuri05 I am more interested in the Flag combination in the UI.

Yuri05:
f_expr_XXX correspond 1:1 to the flags in the UI - or do I miss something?

msevestre:
Maybe.

here is what i have
```
  public static Localization ConvertToLocalization(TissueLocation tissueLocation, MembraneLocation membraneLocation,
         IntracellularVascularEndoLocation vascularEndoLocation)
      {
         switch (tissueLocation)
         {
            case TissueLocation.Interstitial:
[...]

PR #1585, opened 2020-11-03
Work on #1568

msevestre:


src/PKSim.Infrastructure/ProjectConverter/v10/Converter9To10.cs
```diff
@@ -36,12 +60,89 @@ private void convertIndividualProtein(XElement proteinNode)
...
+         allFractionEndosomalParameters.Each(x =>
+         {
+            //Find container in the individual to convert
+            var containerPath = new ObjectPath(x.ParentContainer.ConsolidatedPath().ToPathArray());
+            
+            //Path under individual start with ROOT. This should be removed
+            if(containerPath[0] == Constants.ROOT)
+               containerPath.RemoveFirst();
+
+            var container = individual.Root.EntityAt<IContainer>(containerPath.ToArray());
```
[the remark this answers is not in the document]
msevestre: this is better now. I updated core to not mess around with path starting with a rootcontainer

src/PKSim.Core/Services/IndividualProteinTask.cs
```diff
@@ -125,41 +123,35 @@ protected void AddMucosaExpression(ISimulationSubject simulationSubject, string
...
+            organMucosa.GetChildren<Compartment>().Each(x => AddTissueParameters(x, moleculeName));
          }
       }
 
       protected void AddLumenExpression(ISimulationSubject simulationSubject, string moleculeName)
       {
          var lumen = simulationSubject.Organism.Organ(CoreConstants.Organ.Lumen);
-         foreach (var segment in lumen.Compartments.Where(c => c.Visible))
+         //Only visible compartments as we do not want to create RelExp for example in Feces
+         foreach (var segment in lumen.Compartments.Where(x => x.Visible))
```
msevestre, line 135: @Yuri05 Visible is required indeed. Otherwise compartmetn such as Feces are showing up

src/PKSim.Core/Services/SimulationModelCreator.cs
```diff
@@ -102,50 +98,14 @@ private void updateSimulationAfterModelCreation(Simulation simulation)
...
-      // }
-      //
       private void updateFromIndividualParameter(IParameter parameterToUpdate, IParameter parameterInIndividual, Individual individual, bool visible = true)
       {
          _parameterIdUpdater.UpdateParameterId(parameterInIndividual, parameterToUpdate);
          parameterToUpdate.BuildingBlockType = PKSimBuildingBlockType.Individual;
          parameterToUpdate.Origin.BuilingBlockId = individual.Id;
-         parameterToUpdate.Visible = visible;
+         parameterToUpdate.Visible = parameterInIndividual.Visible;
+         parameterToUpdate.DefaultValue = parameterInIndividual.DefaultValue;
```
msevestre, line 109: This addition is required because default value are set to NaN which means that there is a default value for the parameter (NaN being our default int he DB). We want NULL

src/PKSim.Core/Commands/SetExpressionLocalizationCommand.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace PKSim.Core.Commands
 {
-   internal class SetExpressionLocalizationInProteinCommand : BuildingBlockChangeCommand
+   internal class SetExpressionLocalizationInProteinCommand : BuildingBlockStructureChangeCommand
```
msevestre, line 13: I am not sure about this one but I think this makes sense

src/PKSim.Presentation/Presenters/Individuals/ExpressionParametersPresenter.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace PKSim.Presentation.Presenters.Individuals
+{
+   public interface IExpressionParametersPresenter:IPresenter<IExpressionParametersView>, IEditParameterPresenter
+   {
+      void Edit(IReadOnlyList<ExpressionParameterDTO> expressionParameters);
+      void SetExpressionParameterValue(IParameterDTO expressionParameterDTO, double value);
+
+   }
+
+   public class ExpressionParametersPresenter : EditParameterPresenter<IExpressionParametersView, IExpressionParametersPresenter>, IExpressionParametersPresenter
```
msevestre, line 21: I Am reusing this presenter in the simulation. IT was not possible before because we had a different structure. But it's now the same :)

PR #1622, opened 2020-12-07
Feature/relative expression redesign

msevestre:


msevestre:
Going to merge this so that @abdelr and @georgeDaskalakis can incorporate their changes

PR #1648, opened 2021-03-02
Fixes #1601 information lost for advanced protocol

### LINKED ISSUE #1601: Information in advanced protocol is lost in snapshot
Steps to reproduce:
Create a project with an advanced protocol as the following one:
advancedProtocol
Then export the project to snapshot. The snapshot does not contain the information about the repetitions anymore, see here:
SnapshotAdvancedProtocol

msevestre:


src/PKSim.Core/Snapshots/Mappers/SchemaMapper.cs
```diff
@@ -25,6 +26,13 @@ public override async Task<SnapshotSchema> MapToSnapshot(ModelSchema modelSchema
          return snapshot;
       }
 
+      protected override bool ShouldExportParameterToSnapshot(IParameter parameter)
+      {
+         //we want to ensure that start time and end time are always exported
+         return parameter.NameIsOneOf(Constants.Parameters.START_TIME, CoreConstants.Parameters.NUMBER_OF_REPETITIONS,
```
msevestre, line 32: @Yuri05 Those parameters are set as IsDefault = true  when creating a protocol. By switching back and forth between simple and advanced, those parameters can remains as default true. In any case, for advascncved protocol, we should always expor them to snapshot because it is clearly an input parameter

Yuri05 (approved):

PR #1649, opened 2021-03-02
Fixes #1586 info mouse over

### LINKED ISSUE #1586: Feature wish: Display distribution models in simulation dialog box on mouseover
I think some of the information displayed in the dialog box on mouseover of a simulation is less relevant, e.g. `Model` and `Allow Aging` and could be removed. However, other important information is missing. 
For example, when testing multiple distribution models (e.g. partition coeffs) during model building, I often forget which distribution model is used in the different simulations. Opening the `Configure` window for each simulation is quite cumbersome and I would appreciate if this information could be displayed in the dialog box on mouseover like this:

Untitled

msevestre:


src/PKSim.Infrastructure/Reporting/Summary/Items/CompoundPropertiesCalculationMethods.cs
```diff
@@ -0,0 +1,22 @@
+﻿using PKSim.Core.Model;
+
+namespace PKSim.Infrastructure.Reporting.Summary.Items
+{
+   public class CompoundPropertiesCalculationMethods
```
msevestre, line 5: @Yuri05  We need to create a placeholder object so that we can report the info

src/PKSim.Infrastructure/Reporting/Summary/CompoundPropertiesCalculationMethodsReportBuilder.cs
```diff
@@ -0,0 +1,27 @@
...
+
+      public CompoundPropertiesCalculationMethodsReportBuilder(IReportGenerator reportGenerator)
+      {
+         _reportGenerator = reportGenerator;
+      }
+
+      protected override void FillUpReport(CompoundPropertiesCalculationMethods compoundPropertiesCalculationMethods, ReportPart reportPart)
+      {
+         var (compoundName, compoundProperties) = compoundPropertiesCalculationMethods;
+         var compoundNameToUse = compoundProperties.Compound?.Name ?? compoundName;
```
msevestre: The cimpound is null when the simulation has not been loaded yet

src/PKSim.Infrastructure/Reporting/Summary/IReportBuilder.cs
```diff
@@ -21,6 +21,10 @@ public virtual ReportPart Report(T objectToReport)
       {
          var reportPart = new ReportPart();
          FillUpReport(objectToReport, reportPart);
+         //Only one item returned in the report. Then return it instead of wrapping it in yet another report part
+         if (reportPart.SubParts.Count == 1)
```
msevestre, line 25: The way this construct is working, we are creating a part, adding the actual report part and then returning the top part.
This makes sense for multiple levels but in our case. not really. 
So when we only have one level, we return the level instead of returning a bunch of empty wrappers

Yuri05 (approved):

PR #1663, opened 2021-03-09
Fixes #1661 export aging data

### LINKED ISSUE #1661: AgingData export - make sure that parameter names are saved in " "
to avoid sthg like

0,Organism|Fat|Vf (neutral phospholipid, plasma)-WS,0,0.00131199927600357

@Yuri05 Where should we add the method to save aging data as csv? Right now this is very hidden (Export for cluster computation). Do we need it somewhere else?

msevestre:


src/PKSim.Core/Model/AgingData.cs
```diff
@@ -52,6 +48,8 @@ public virtual DataTable ToDataTable()
          return dataTable;
       }
 
+      private string inQuote(string text) => $"\"{text}\"";
```
msevestre, line 51: This method should be in core. I'll add it later

src/PKSim.Core/Services/FileSelection.cs
```diff
@@ -31,6 +33,22 @@ public virtual string Description
          set => SetProperty(ref _description, value);
       }
 
+      public FileSelection AddSuffixToFileName(string suffix)
```
msevestre, line 36: In the spirit of trying to not mutate data...I am returning here a NEW instance with suffixed path and desription

Yuri05 (approved):

PR #1726, opened 2021-04-28
Adding missing organs

abdelr:


msevestre (changes requested):
No. Please use the code that was in PK-Sim before
https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/master/src/PKSim.Infrastructure/Services/ImportObservedDataTask.cs#L272

abdelr:
> No. Please use the code that was in PK-Sim before
> https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/master/src/PKSim.Infrastructure/Services/ImportObservedDataTask.cs#L272

Thanks for the reference. Done already.

Yuri05 (approved):

PR #1733, opened 2021-04-30
Export pkml qualification

### LINKED ISSUE #1696: Add option to NOT calculate the outputs in PKSim (using the CLI)

msevestre:
Fixes #1696


msevestre:
@Yuri05 This ensures that we export pkml etc... even if no results are available. This is required for the new reporting engine that moves calculation out of the runner and to the R side

msevestre:
Merging.

PR #1744, opened 2021-05-04
Small fix and update dlls

abdelr:


src/PKSim.Infrastructure/Services/ImportObservedDataTask.cs
```diff
@@ -164,7 +164,10 @@ private void AddObservedDataFromConfiguration(ImporterConfiguration configuratio
 
       private DataRepository findDataRepositoryInList(IEnumerable<DataRepository> dataRepositoryList, DataRepository targetDataRepository)
       {
-         return (from dataRepo in dataRepositoryList let result = targetDataRepository.ExtendedProperties.KeyValues.All(keyValuePair => dataRepo.ExtendedProperties[keyValuePair.Key].ValueAsObject.ToString() == keyValuePair.Value.ValueAsObject.ToString()) where result select dataRepo).FirstOrDefault();
+         return (from dataRepo in dataRepositoryList let result = targetDataRepository.ExtendedProperties.KeyValues.All(keyValuePair =>
```
msevestre, line 167: That feels that some exact same code that we had in core. Why do we have it twice?

msevestre: or am I mistaken?

msevestre (changes requested):

PR #1745, opened 2021-05-04
Fixes #1724 expression profile lost after scaling

### LINKED ISSUE #1724: Expression profile lost after scaling individual
Using the current build of v10, large parts of the expression profile of an individual is lost when scaling to another individual. E.g. when scaling from ICRP to Asian, only the following expression value is kept - all other organs/compartments are kicked out:
image

msevestre:


msevestre:
@Yuri05 This should fix the scaling issue

PR #1760, opened 2021-05-18
Fixes #1757 error resolving path

### LINKED ISSUE #1757: RelExp V10: Error resolving path
* Open the project [10.0.135_Fabs_DDI_MultipleCombinations_EffluxOnly.pksim5.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/6500609/10.0.135_Fabs_DDI_MultipleCombinations_EffluxOnly.pksim5.zip)
* Open the individual-BB and remove CYP3A4 
  

* In the simulation: Update the individual from Building Block and click Next: the error message below comes
grafik

msevestre:


tests/PKSim.Tests/IntegrationTests/IndividualEnzymeFactorySpecs.cs
```diff
@@ -57,6 +59,18 @@ public void should_have_marked_the_reference_concentration_parameter_as_variable
...
+      [Observation]
+      public void should_have_ensure_that_all_formula_have_different_ids()
+      {
+         var allInitialConcentrationParameters = _individual.AllMoleculeParametersFor(_result)
+            .Where(x => x.IsNamed(CoreConstants.Parameters.INITIAL_CONCENTRATION))
+            .ToList();
+
+         var allFormulaIds = new HashSet<string>(allInitialConcentrationParameters.Select(x => x.Formula.Id));
+
+         allInitialConcentrationParameters.Count.ShouldBeEqualTo(allFormulaIds.Count);
```
msevestre, line 72: I am testing that I am indeed creating copy of the formula.

src/PKSim.Core/Services/IndividualMoleculeFactory.cs
```diff
@@ -128,7 +131,10 @@ protected TMolecule CreateMolecule(string moleculeName)
          if (!string.IsNullOrEmpty(groupName))
             parameter.GroupName = groupName;
 
+         //Because we update the formula of the parameter, we need to make sure we also reset the formulaId so that it will appear as being unique
          parameter.Formula.ReplaceKeywordsInObjectPaths(new[] {ObjectPathKeywords.MOLECULE}, new[] {moleculeName});
+         parameter.Formula.Id = _idGenerator.NewId();
```
msevestre, line 136: @Yuri05 This was the error here. We have a formula that we get from the DB but we modify it "by hand". I need to update the Id otherwise it will be serialized as ONE and on load, you get the wrong references...

msevestre (commented):
very interesting.

Yuri05 (approved):

PR #1766, opened 2021-05-20
Fixes #1763 clone configure issue

### LINKED ISSUE #1763: Cannot clone/configure simulation
PK-Sim 10.0.137
Load the snapshot [07_MM_Competitive_Competitive_Mechanismbased_Mechanismbased.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/6509630/07_MM_Competitive_Competitive_Mechanismbased_Mechanismbased.zip)
Try to clone the simulation.
grafik

Same error when trying to configure the simulation

msevestre:


src/PKSim.Core/Services/IndividualTransporterFactory.cs
```diff
@@ -84,6 +84,9 @@ public IndividualTransporter UndefinedLiverTransporterFor(Individual individual)
             );
          });
 
+         _individualPathWithRootExpander.AddRootToPathIn(individual, transporter.Name);
```
msevestre, line 87: @Yuri05 This was missing here as well as for Enzyme. Also I am  now adding the molecule directlyhere instead of letting the caller having to do it. IT was ok before, it is not now

PR #1777, opened 2021-05-26
Fixes #1662 rename building block can corrupt snapshots

### LINKED ISSUE #1662: Simulations cannot be loaded from snapshot if used Individual or Population Building block was renamed
If I rename an individual (or population) building block and this building block is used in a simulation: this simulation cannot be reloaded from snapshot anymore.

This applies **only to individuals/populations**; all other building block can be renamed without producing such an error.

Example: **P1** is the initial project; projects **P1_renamedXXX** are created from P1 via renaming one of the building blocks. and are exported to snapshot.
Snapshots **P1_renamedIndividual.json** and **P1_renamedPopulation.json** produce then errors while loading; all other **P1_renamed*** snapshots can be loaded without problems.

[ErrorLoadingSnapshot.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/6109175/ErrorLoadingSnapshot.zip)

```
Information: Loading Project from snapshot file 'C:\2del\P1_renamedIndividual.json'
Warning: Snapshot is outdated and cannot be loaded for the following reason: 

Individual 'I1 not found in project.
Error: Cannot load Simulation 'S1'
Information: Project loaded from snapshot
```

```
Information: Loading Project from snapshot file 'C:\2del\P1_renamedPopulation.json'
Warning: Snapshot is outdated and cannot be loaded for the following reason: 

Population 'Pop1 not found in project.
Error: Cannot load Simulation 'S2'
Information: Project loaded from snapshot
```

msevestre:


src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
```diff
@@ -236,14 +235,23 @@ private string[] usedObservedDataFrom(ModelSimulation simulation, PKSimProject p
...
+      private string usedSimulationSubject<T>(ModelSimulation simulation, PKSimProject project) where T : class, ISimulationSubject
       {
-         var name = simulation.BuildingBlock<T>()?.Name;
+         var usedBuildingBlock = simulation.UsedBuildingBlockInSimulation<T>();
+
+         //simulation is not using this building block. 
+         if (usedBuildingBlock == null)
+            return null;
+
+         var templateBuildingBlock = project.BuildingBlockById<T>(usedBuildingBlock.TemplateId);
```
msevestre, line 246: @Yuri05 I am now retrieving the templatebuildign block by ID so that it does not matter if sthg was renamed or not. We can always use it. This is what is happening for all other building blocks (Events, Formulation etc... )That's why it was working

PR #1778, opened 2021-05-26
Fixes #1768 rename problem

### LINKED ISSUE #1768: Refresh Problem when renaming BBs or Simulations
Happens both in V9 and V10.
When I rename a building block or a simulation and give it a long name: it's truncated in the BB-/Sim-Explorer
 

The long name is only displayed properly after I reopen the project or add another building block/simulation

grafik

msevestre:


msevestre:
@Yuri05 This should do it

PR #1795, opened 2021-06-17
Do not trigger when cancelling import

### LINKED ISSUE #1794: App crashes when cancelling import
@abdelr commented on [Thu Jun 17 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1098)

abdelr:
Fixes #1794 (App crashes when cancelling import)

### DIFF
--- a/src/PKSim.Infrastructure/Services/ImportObservedDataTask.cs
+++ b/src/PKSim.Infrastructure/Services/ImportObservedDataTask.cs
@@ -227,7 +227,7 @@ namespace PKSim.Infrastructure.Services
 
          var importedObservedData = _dataImporter.ImportDataSets((IReadOnlyList<MetaDataCategory>)metaDataCategories, importConfiguration(), dataImporterSettings);
 
-         if (importedObservedData.DataRepositories == null || importedObservedData.Configuration == null) return;
+         if (importedObservedData.DataRepositories == null || importedObservedData.Configuration == null || string.IsNullOrEmpty(importedObservedData.Configuration.Id)) return;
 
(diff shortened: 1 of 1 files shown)


src/PKSim.Infrastructure/Services/ImportObservedDataTask.cs
```diff
@@ -227,7 +227,7 @@ private void addObservedData(Func<IReadOnlyList<ColumnInfo>> importConfiguration
 
          var importedObservedData = _dataImporter.ImportDataSets((IReadOnlyList<MetaDataCategory>)metaDataCategories, importConfiguration(), dataImporterSettings);
 
-         if (importedObservedData.DataRepositories == null || importedObservedData.Configuration == null) return;
+         if (importedObservedData.DataRepositories == null || importedObservedData.Configuration == null || string.IsNullOrEmpty(importedObservedData.Configuration.Id)) return;
```
msevestre, line 230: May I ask how we can have  a configuration that has no id?

abdelr: Yep, there was a wrong initialization of the configuration on the Core side. I fixed it now (since it was quite obvious once found) and I am planning to write a test for this later today after I finish some other more pressing issues. I would say we should actually close this PR even without merging this change. I am not sure we should ignore a bad configuration here since it is better to crash if unintended use rather than silently ignore.

msevestre: agreed. closing

PR #1801, opened 2021-06-21
1798 influx efflux

msevestre:


src/PKSim.Core/Services/TransportContainerUpdater.cs
```diff
@@ -58,9 +59,27 @@ public void SetDefaultSettingsForTransporter(ISimulationSubject simulationSubjec
          foreach (var transporterContainer in simulationSubject.AllMoleculeContainersFor<TransporterExpressionContainer>(transporter))
          {
             updateTransporterContainerFromTemplate(transporterContainer, null, transportType);
+            updateFractionExpressedEpithelial(transporterContainer, transportType);
          }
       }
 
+      private void updateFractionExpressedEpithelial(TransporterExpressionContainer transporterContainer, TransportType transportType)
+      {
+         if (!transportType.IsOneOf(TransportType.Efflux, TransportType.Influx))
```
msevestre, line 68: @Yuri05 Only udating the behavior for Influx and Efflux. Not sure what BiDirectional or pgp should do

Yuri05: PgP: same as Efflux I would say.
Bidirectional: no idea. Maybe @PavelBal can answer.

src/PKSim.Core/Services/TransportContainerUpdater.cs
```diff
@@ -58,9 +59,27 @@ public void SetDefaultSettingsForTransporter(ISimulationSubject simulationSubjec
...
+         var fractionExpressedEpithelial = transporterContainer.Parameter(CoreConstants.Parameters.FRACTION_EXPRESSED_EPITHELIAL);
+         if (fractionExpressedEpithelial == null)
+            return;
+
+         //value was set by the user.
+         if (fractionExpressedEpithelial.Value != 0 && fractionExpressedEpithelial.Value != 1)
+            return;
+         
+         //Set the value according to the new transport type
+         fractionExpressedEpithelial.Value = transportType == TransportType.Efflux ? 1 : 0;
```
Yuri05, line 80: is this function called only for "Organs with Lumen" (Liver, Kidney, GI) or for all organs? Because for other organs
fraction_expressed_Epithelial should be always const=0

msevestre: fractionExpressedEpithelial only exists for organ with lumen by construction

msevestre: here
https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/develop/src/PKSim.Core/Services/IndividualTransporterFactory.cs#L145

Yuri05: ok, all good then

Yuri05 (approved):

src/PKSim.Core/Services/TransportContainerUpdater.cs
msevestre: ok updating for pGP and leaving as is for BiDirectional

src/PKSim.Core/Services/TransportContainerUpdater.cs
```diff
@@ -58,9 +59,27 @@ public void SetDefaultSettingsForTransporter(ISimulationSubject simulationSubjec
...
+         var fractionExpressedEpithelial = transporterContainer.Parameter(CoreConstants.Parameters.FRACTION_EXPRESSED_EPITHELIAL);
+         if (fractionExpressedEpithelial == null)
+            return;
+
+         //value was set by the user.
+         if (fractionExpressedEpithelial.Value != 0 && fractionExpressedEpithelial.Value != 1)
+            return;
+         
+         //Set the value according to the new transport type (Efflux one, all other 0)
+         fractionExpressedEpithelial.Value = transportType == TransportType.Efflux ? 1 : 0;
```
Yuri05, line 80:
```
         //Set the value according to the new transport type (Efflux one, all other 0)
         fractionExpressedEpithelial.Value = transportType == TransportType.Efflux ? 1 : 0;
```
I think it's not correct for PgP. I would expect that fraction_expressed_endothelial = 1 for PgP (same as for Efflux)

msevestre: oh sorry I read same as Influx. Will correct in next PR

PR #1813, opened 2021-06-25
1781 ckd populations

msevestre:


msevestre:
don't forget to run postclean so that the db is copied at the right place...

PR #1818, opened 2021-06-29
Update core and fire event

### LINKED ISSUE #1782: Observed data import redesign: Refresh Problem
If observed data is plotted somewhere (simulation, PI): after updating the observed data via "Reload under the same settings": simulation/PI has to be closed and reopened in order to display the new observed data values.

abdelr:
Fixes #1782 (Fires event on DataRepository reloaded)

### DIFF
--- a/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
+++ b/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
@@ -26,8 +26,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="10.0.259" />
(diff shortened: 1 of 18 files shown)


abdelr:
I am not sure if PI and SimulationResults will be updated since I find them to implement IListener<ObservedDataValueChangedEvent>. I would say for the Importer scope this change is enough but we need to extend every view that will have to update the data on this event. This change actually solves the fact that if you have the DataRepositoryEditor opened and reload, now the view (DataRepositoryChartPresenter and DataRepositoryDataPresenter do implement IListener<ObservedDataValueChangedEvent>) is refreshed. I would like to solve the issue on PI and SimulationResults but I have not yet gone that far. We might discuss whether we do this now, or later.

msevestre:
Good stuff. Great use of application wide events. Well done

PR #1827, opened 2021-07-26
Fixes #1825 bug comparisoon chart

### LINKED ISSUE #1825: Bug comparison chart population
Hi,

I found a bug in the comparison chart for population simulation in PK-Sim. I was able to reproduce the bug on a dummy file,
it is attached. This is a description of the bug:
If you have two population simulation in a project file, create a comparison chart with the two populations., by adding a grouping by simulation name (in my case this was selected by default).  Once that the comparison chart is created, right click on the graph and chose edit (in my case I wanted to change units on the plot). What happened, is that the grouping by simulation name is duplicated, and this produce an error when the user tries to close the chart with the changes. The user needs to delete the duplicated selection.

I found this on PK-Sim 9 build 170, so not really the very last version.
Just let me know if you have issues in reproducing.

Ciao!

Donato

 
[Bug comparison chart population.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/6853717/Bug.comparison.chart.population.zip)

msevestre:


src/PKSim.Presentation/Presenters/PopulationAnalyses/CreatePopulationAnalysisPresenter.cs
```diff
@@ -125,15 +125,18 @@ protected virtual bool Edit(IPopulationDataCollector populationDataCollector, Po
...
       protected virtual void AddDefaultAnalysisField()
       {
-        //Add grouping by simulation name when comparing simulations as it is a step that is always required in order to display something meaningful
-        if (!_populationDataCollector.IsAnImplementationOf<PopulationSimulationComparison>())
-           return;
+         //Add grouping by simulation name when comparing simulations as it is a step that is always required in order to display something meaningful
+         if (!_populationDataCollector.IsAnImplementationOf<PopulationSimulationComparison>())
+            return;
+
+         if (PopulationAnalysis.Has(CoreConstants.Covariates.SIMULATION_NAME))
```
msevestre, line 132: @Yuri05 Only adding the field if not already available.

PR #1850, opened 2021-09-03
1844 multiple click

msevestre:


src/PKSim.UI/Views/Core/BaseUserControlWithValueInGrid.cs
```diff
@@ -37,7 +33,7 @@ protected virtual void OnGridViewMouseDown(UxGridView gridView, MouseEventArgs e
 
          if (ColumnIsButton(col))
          {
-            RaiseButtonClick(gridView, col, rowHandle, e);
+            gridView.RaiseButtonClick(e);
```
msevestre, line 36: this was moved to core

src/PKSim.UI/Views/Applications/ApplicationParametersView.cs
```diff
@@ -49,10 +46,10 @@ public ApplicationParametersView(IImageListRetriever imageListRetriever)
          mainView.ShowColumnHeaders = false;
 
          gridViewParameters.OptionsView.ShowGroupPanel = false;
+         gridViewParameters.MultiSelect = false;
```
msevestre, line 49: This was enough to make it work. Using Multiselect true created some issue with background color. Good enough for now

PR #1852, opened 2021-09-03
Fixes #1710 wrong defualt for mucosa influx

### LINKED ISSUE #1710: Apply stored configuration to a set of files
Importer should allow for applying a saved configuration to a folder. This means that:

1. User selects a folder where the data files are.
    - We might think about filtering the files by extension or by a pattern in the name.
2. All information except source file will be applied to the existing data coming from the set of files.
    - Information to use:
        - Mappings
        - Filters
        - Naming pattern
        - NaN-settings (indicator, row settings)
        - Sheets
    - If some mapping columns are missing in the files, the user should be informed but the process should continue.
3. Sheet information from the configuration should be used to select the data to import.
    - If some sheets are missing in the files, the user should be informed but the process should continue.
    - If no sheet information is given in the configuration we should either inform the user that the configuration is not complete or load all the sheets from all files.
4. If it ends up in an inconsistent state at some files, those files will not be imported and the user should be informed but the remaining files should be used.
    - As inconsistent state, consider for example that the time column is missing, or that the units of measurement and error have different dimensions. In this case the view should clearly show the error but allow the user to fix such an error.
5. The user should be prompted with the full list of data to import at the end of the process where he/she could either accept or cancel the import.

msevestre:


src/PKSim.Core/Services/IndividualTransporterFactory.cs
```diff
@@ -140,9 +140,13 @@ private void addOrganWithLumenParameters(IContainer organ, IndividualTransporter
             InitialConcentrationParam(CoreConstants.Rate.INITIAL_CONCENTRATION_INTERSTITIAL_TRANSPORTER)
          );
 
+         //By default, active transport in mucosa are always apical
+         var defaultExpressionApical = organ.IsInMucosa() ? CoreConstants.Rate.ONE_RATE :
```
msevestre, line 144: I think this is correct...

src/PKSim.Core/Services/TransportContainerUpdater.cs
```diff
@@ -68,16 +69,16 @@ private void updateFractionExpressedEpithelial(TransporterExpressionContainer tr
...
 
          //value was set by the user.
-         if (fractionExpressedEpithelial.Value != 0 && fractionExpressedEpithelial.Value != 1)
+         if (fractionExpressedApical.Value != 0 && fractionExpressedApical.Value != 1)
             return;
-         
-         //Set the value according to the new transport type (Efflux + pgp one, Influx 0)
-         fractionExpressedEpithelial.Value = transportType == TransportType.Influx ? 0 : 1;
+
+         //Set the value according to the new transport type (Efflux + pgp one, Influx 0) except in mucosa where it is always apical
```
msevestre, line 80: Comment was from before. Clearly I read it somewhere  :)

tests/PKSim.Tests/Core/TransportContainerUpdaterSpecs.cs
```diff
@@ -146,19 +153,26 @@ protected override void Because()
...
-      public void should_set_the_value_of_the_fraction_expressed_epithelial_to_zero()
+      public void should_set_the_value_of_the_fraction_expressed_epithelial_to_zero_in_all_non_mucosa_compartment()
       {
          _fractionExpressedApical.Value.ShouldBeEqualTo(0);
       }
+
+      [Observation]
+      public void should_set_the_value_of_the_fraction_expressed_epithelial_to_one_in_all_mucosa_compartment()
+      {
+         _fractionExpressedApicalMucosa.Value.ShouldBeEqualTo(1);
```
msevestre, line 164: and here we go :)

Yuri05 (approved):

PR #1859, opened 2021-09-08
Fixes #1858 return seed

### LINKED ISSUE #1858: Get seed of population/individual characteristics
@PavelBal commented on [Tue Aug 24 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/587)

As an extension of https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/472.

I think it should be also possible to get the seed that is used to create a certain individual/population if a random seed has been used. When is the seed generated (of none passed) - during creation of characteristics or individiaul/populatoin?

---

@msevestre commented on [Wed Sep 08 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/587#issuecomment-915257976)

yes that can be exposed

msevestre:


src/PKSim.R/Services/CreateIndividualResults.cs
```diff
@@ -6,11 +6,13 @@ public class CreateIndividualResults
    {
       public ParameterValueWithUnit[] DistributedParameters { get;  }
       public ParameterValueWithUnit[] DerivedParameters { get;  }
+      public int Seed { get;  }
```
msevestre, line 9: Adding this to the return value

src/PKSim.R/Services/CreatePopulationResults.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+   public class CreatePopulationResults
+   {
+      public IndividualValuesCache IndividualValuesCache { get; }
+      public int Seed { get; }
+
+      public CreatePopulationResults(IndividualValuesCache individualValuesCache, int seed)
+      {
+         IndividualValuesCache = individualValuesCache;
+         Seed = seed;
```
msevestre, line 13: New type for population. Code needs to be adjusted in R

src/PKSim.R/Services/IndividualFactory.cs
```diff
@@ -113,13 +113,7 @@ public DistributedParameterValueWithUnit[] DistributionsFor(IndividualCharacteri
 
       private OriginData originDataFrom(Core.Snapshots.OriginData originData)
       {
-         var coreOriginData = _originDataMapper.MapToModel(originData).Result;
```
msevestre, line 116: This was a bug here that needed to be fixed for R. Why do we required age but not height etc.. not consistent.

src/PKSim.R/Services/PopulationFactory.cs
```diff
@@ -41,11 +40,11 @@ public class PopulationFactory : IPopulationFactory
          _ontogenyVariabilityUpdater = ontogenyVariabilityUpdater;
       }
 
-      public IndividualValuesCache CreatePopulation(PopulationCharacteristics populationCharacteristics)
+      public CreatePopulationResults CreatePopulation(PopulationCharacteristics populationCharacteristics)
       {
          var populationSettings = _populationSettingsMapper.MapToModel(populationCharacteristics).Result;
          validate(populationSettings);
-         var population = _randomPopulationFactory.CreateFor(populationSettings, new CancellationToken()).Result;
+         var population = _randomPopulationFactory.CreateFor(populationSettings, new CancellationToken(), seed: populationCharacteristics.Seed).Result;
```
msevestre, line 47: This was a bug here also. The seed was not used....impossible to reproduce the same population

msevestre:
@Yuri05 I will merge as soon as the code compiles so that I can move with the R side

PR #1865, opened 2021-09-20
fixes probkem with export

msevestre:


src/PKSim.CLI.Core/Services/QualificationRunner.cs
```diff
@@ -101,7 +101,8 @@ public async Task RunBatchAsync(QualificationRunOptions runOptions)
          var exportRunOptions = new ExportRunOptions
          {
             OutputFolder = projectOutputFolder,
-            ExportMode = SimulationExportMode.Xml | SimulationExportMode.Csv | SimulationExportMode.Pkml
+            //We run the output, this is for the old matlab implementation where we need xml. Otherwise, we only need pkml export
+            ExportMode = runOptions.Run ?   SimulationExportMode.Xml | SimulationExportMode.Csv  :  SimulationExportMode.Pkml
```
msevestre, line 105: @Yuri05 As discussed. If we run the sim, we export the csv and XML. Otherwise we export only pkml

PR #1868, opened 2021-09-21
Fixes #1867 use json

### LINKED ISSUE #1867: Update versions.xml to use versions.json
Looks like it would be more approprotate since we are using json for everything else

msevestre:


msevestre:
@Yuri05 Reusing stuff that we have already from snapshot. Autoamtic cast etc.. Much cleaner

msevestre:
@Yuri05 Merging as I need to update the code again. Please review when you have time

PR #1876, opened 2021-10-01
Ensure that object type resolver is thread safe

msevestre:


msevestre:
Apparently, this is the reason why the PI in // is crachigng.. Are we thread safe now?

PR #1889, opened 2021-10-06
Fixes #1888 nopop for csv

### LINKED ISSUE #1888: population files not exported
Not sure if this is the right place.
When qualification runner is launched with the option `no-run`, the population files are not exported

msevestre:


src/PKSim.CLI.Core/Services/SimulationExporter.cs
```diff
@@ -127,13 +127,17 @@ public Task Export(Simulation simulation, SimulationExportOptions simulationExpo
          var exportMode = simulationExportOptions.ExportMode;
 
          if (exportMode.HasFlag(SimulationExportMode.Csv) && individualSimulation != null)
-            tasks.Add(exportIndividualSimulationToCsvAsync(individualSimulation, simulationExportOptions));
+            tasks.Add(exportSimulationResultsToCsv(individualSimulation, simulationExportOptions));
+
+         // Population always exported no matter what
+         if(populationSimulation != null)
+            tasks.Add(exportPopulationToCsvAsync(populationSimulation, simulationExportOptions));
```
msevestre, line 134: @Yuri05 This is new here. Always exporting

Yuri05: this will also export the aging data, right?

Yuri05: when exporting a pop simulation with aging for cluster: aging parameters are exported...
Can you add it here?

msevestre: yep (never exported as part as the qualification export I meant)

msevestre: done. Also removing the append method for log that is not supported by serilog anymore

PR #1941, opened 2021-11-02
Fixes #1929 hill transporter issue

### LINKED ISSUE #1929: Active influx transports Lumen=>Mucosa with the Hill kinetics: cannot create simulations
E.g. try to create a simulation in the attached project
[MucosaInflux_Hill.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/7441001/MucosaInflux_Hill.zip)
grafik

msevestre:


src/Db/Dump/PKSimDB_dump.txt
```diff
@@ -82970,7 +82970,7 @@ INSERT INTO tab_rate_generic_molecules VALUES('LinksCommon','MetabolizationSpeci
 INSERT INTO tab_rate_generic_molecules VALUES('LinksCommon','MetabolizationSpecific_Hill',372,'DRUG','C',0);
 INSERT INTO tab_rate_generic_molecules VALUES('LinksCommon','MetabolizationSpecific_MM',370,'DRUG','CP',0);
 INSERT INTO tab_rate_generic_molecules VALUES('LinksCommon','MetabolizationSpecific_MM',372,'DRUG','C',0);
-INSERT INTO tab_rate_generic_molecules VALUES('LinksCommon','MucosaActiveInfluxFromLumen_Hill',368,'DRUG','TM',1);
+INSERT INTO tab_rate_generic_molecules VALUES('LinksCommon','MucosaActiveInfluxFromLumen_Hill',366,'DRUG','TM',1);
```
msevestre, line 82973: This was Target|Transporter. Now Source|Transporter

Yuri05: no, that's not right. Is probably a relic from V9 (where all GI processes were defined with source in Mucosa and Influx-transports had the negative rate to compensate this)
Must be defined Lumen=>Mucosa (same as e.g. MucosaActiveInfluxFromLumen_MM)

msevestre:
MucosaActiveInfluxFromLumen_MM is defined like this as well

msevestre:
And setting path to be SOURCE/ doesn't work. So clearly the transporter is not defined here...we might have a little problem to check

msevestre:
image

msevestre:
@PavelBal Is it by design that the Lumen => Mucosa transporter observed the Transporter conc. in Mucosa?
It is clearly not defined in Lumen

image

msevestre:
@Yuri05 That's the same for  `ActiveEffluxSpecificIntracellularToInterstitial_MM`
We are observing the concentration in Interstitial (target) 

image

Yuri05:
ok, after some thinking: that's correct. We place now the basolateral transporter (`Mucosa|XXX|Interstitial<=>Mucosa|XXX|Intracellular`) always into **Mucosa|XXX|Interstitial** and the apical transporter (`Lumen|XXX<=>Mucosa|XXX|Intracellular`) always into **Mucosa|XXX|Intracellular**.
Thus for the influx transporters Lumen=>Mucosa the Source is Lumen, but the transporter amount is taken from the target (=Mucosa)

Yuri05 (approved):

PavelBal:
> @PavelBal Is it by design that the Lumen => Mucosa transporter observed the Transporter conc. in Mucosa?
> It is clearly not defined in Lumen

Yes, it makes sense from the physiological point of view.

PR #1945, opened 2021-11-03
Fixes #1943 add species to create expression profile

### LINKED ISSUE #1943: Expression Profile: How do we deal with species dependent action?
@PavelBal @Yuri05 @StephanSchaller 
At the moment, the species is known before we create an expression profile which allows us to select the specific "Expression Database" (human, or something else for those lucky enough to have another version working for another species)

Now with the new Expression profile concept, it is somewhat different because we do not have a species 
Suggestion #1
We create an Expression Profile for a given Species (e.g. when we create one, you also need to specify for which species it applies)
Suggestion #2
When loading from database, we need to select the expressions database to pick from (only if there is more than one defined)

Thoughts?

msevestre:


src/PKSim.Core/Model/ExpressionProfile.cs
```diff
@@ -4,7 +4,9 @@ namespace PKSim.Core.Model
 {
    public class ExpressionProfile : PKSimBuildingBlock
    {
-      public ExpressionProfile():base(PKSimBuildingBlockType.ExpressionProfile)
+      public virtual Species Species { get; set; }
```
msevestre, line 7: Species is saved as member of ExpresssionProfile

msevestre (commented):
Trivial for now.

PR #1947, opened 2021-11-04
color grouping by dropping to chart implemented

### LINKED ISSUE #1946: Color group observed data when drag and dropping folders to charts
When adding data sets to a chart via drag-and-drop of a folder (or multiple folders), there should be the possibility to get them all assigned the same color in the chart. The whole feature specification can be found in the corresponding issue in Core:

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1337

georgeDaskalakis:
close #1946

### DIFF
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -1565,7 +1565,6 @@ namespace PKSim.Assets
          public static readonly string ScaleButton = "Scale";
          public static readonly string AllowsScientificNotation = "Allows scientific notation";
          public static readonly string ShouldRestoreWorkspaceLayout = "Restore opened view when loading project";
-         public static readonly string ShouldColorGroupObservedData = "Color group observed data from same folder when dropping to chart";
          public static readonly string ShowUpdateNotification = "Show software update notification if available";
          public static readonly string ActiveSkin = "Active skin";
          public static readonly string DecimalPlace = "Decimal place";
@@ -1896,7 +1895,6 @@ namespace PKSim.Assets
          public static readonly string ComparisonViewDescription = "Show or hide the comparison";
          public static readonly string SimulationExplorerViewDescription = "Show or hide the simulation explorer";
          public static readonly string BuildingBlockExplorerViewDescription = "Show or hide the building block explorer";
-         public static readonly string ColorGroupObservedDataContextMenu = "Color group when adding to chart";
          public static readonly string UndoDescription = "Undo the last action";
          public static readonly string AddObservedDataDescription = "Add observed data (concentration, amount, fraction) to the project...";
          public static readonly string ExportHistoryToExcelDescription = "Export history to Excel file...";
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -1565,6 +1565,7 @@ namespace PKSim.Assets
          public static readonly string ScaleButton = "Scale";
          public static readonly string AllowsScientificNotation = "Allows scientific notation";
          public static readonly string ShouldRestoreWorkspaceLayout = "Restore opened view when loading project";
+         public static readonly string ShouldColorGroupObservedData = "Color group observed data from same folder when dropping to chart";
          public static readonly string ShowUpdateNotification = "Show software update notification if available";
          public static readonly string ActiveSkin = "Active skin";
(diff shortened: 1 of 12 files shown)


src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -1565,6 +1565,7 @@ public static class UI
          public static readonly string ScaleButton = "Scale";
          public static readonly string AllowsScientificNotation = "Allows scientific notation";
          public static readonly string ShouldRestoreWorkspaceLayout = "Restore opened view when loading project";
+         public static readonly string ShouldColorGroupObservedData = "Color group observed data from same folder when dropping to chart";
```
msevestre, line 1568: You can define this in core. It should be the same for PKSim and MoBi right?

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -1895,6 +1896,7 @@ public static class UI
          public static readonly string ComparisonViewDescription = "Show or hide the comparison";
          public static readonly string SimulationExplorerViewDescription = "Show or hide the simulation explorer";
          public static readonly string BuildingBlockExplorerViewDescription = "Show or hide the building block explorer";
+         public static readonly string ColorGroupObservedDataContextMenu = "Color group when adding to chart";
```
msevestre, line 1899: same with this

src/PKSim.Presentation/Presenters/Charts/ChartPresenter.cs
```diff
@@ -51,7 +53,8 @@ public abstract class ChartPresenter<TChart, TView, TPresenter> : OSPSuite.Prese
          IIndividualPKAnalysisPresenter pkAnalysisPresenter,
          IChartTask chartTask, 
          IObservedDataTask observedDataTask, 
-         IChartUpdater chartUpdater, bool useSimulationNameToCreateCurveName)
+         IChartUpdater chartUpdater, bool useSimulationNameToCreateCurveName,
```
msevestre, line 56: align this parameter to the next line (old code but still, why not)

src/PKSim.UI/UserSettings.cs
```diff
@@ -63,6 +63,7 @@ public class UserSettings : ValidatableDTO, IUserSettings
       public JournalPageEditorSettings JournalPageEditorSettings { get; set; }
       public string LastIgnoredVersion { get; set; }
       public OutputSelections OutputSelections { get; set; }
+      public bool ColorGroupObservedDataFromSameFolder { get; set; }
```
msevestre, line 66: This needs to be added to serialization

msevestre (changes requested):

src/PKSim.Presentation/Presenters/Charts/ChartPresenter.cs
georgeDaskalakis: not sure how you mean... should "IUserSettings userSettings)" go to the line above?

src/PKSim.UI/UserSettings.cs
georgeDaskalakis: done

src/PKSim.Assets/PKSimConstants.cs
georgeDaskalakis: done

src/PKSim.Assets/PKSimConstants.cs
georgeDaskalakis: done

src/PKSim.Presentation/Presenters/Charts/ChartPresenter.cs
msevestre: Lol no
The useSimulationNameToCreateCurve

But that's not a big deal.

msevestre:
I can't merge this until you make a PR in core correct?

georgeDaskalakis:
You can merge, it will just be merged in the feature branch here in PK-Sim. I have commited the changes in core (the constants move from here) directly to the feature branch. I mean since we have one, I think it is OK to do like that without a PR for jus the 2 lines of code for the definition right?

msevestre:
>I think it is OK to do like that without a PR for jus the 2 lines of code for the definition right?
Oh yes for sure. I just did not see those changes that's why. Let's make sure we only do that for obvious trivial stuff. Adding constants etc.. is clearly the case.

PR #1948, opened 2021-11-04
Fixes #1917 explorer bb structure

### LINKED ISSUE #1917: Expression Profile: Tree structure in bb explorer
Flow: Under the Exrepssion Node in the building block explorer, we could have sthg like this
image

Right click on each top node would show the usual Add Enzyme etc..

msevestre:
This PR implements the basic structure required to get the Expression Profile going in PKSim. 

- Expression Profile Building Block management (creating, deleting, etc.. as well as serialization)
- UI to create and Edit
-  and all the other goodies that we typically have for building blocks


src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -398,6 +398,7 @@ public static class Error
          public static string DistributionUnknown(string distribution) => $"Distribution '{distribution}' is unknown.";
          public const string NameIsRequired = "Name is required.";
          public const string MoleculeIsRequired = "Molecule is required.";
+         public const string CategoryIsRequired = "Category is required.";
```
msevestre, line 401: @Yuri05 @PavelBal I am not sure if Category is the right name. For instance CYP3A4-Healthy, healthy is the category

image


Also not sure that Settings is the right name :)

src/PKSim.Core/Model/ExpressionProfile.cs
```diff
@@ -1,13 +1,40 @@
...
+      {
+         get => _category;
+         set
+         {
+            _category = value;
+            RefreshName();
+         }
+      }
+
+      public Individual Individual { get; set; }
```
msevestre, line 29: @Yuri05 I am saving a full Individual for each profile. However the individual has no parameters (only container structure)

src/PKSim.Core/Model/ExpressionProfileFactory.cs
```diff
@@ -0,0 +1,60 @@
...
+            return;
+
+         updateSpecies(expressionProfile, species, currentMolecule.GetType());
+      }
+
+      private void updateSpecies(ExpressionProfile expressionProfile, Species species, Type individualMoleculeType)
+      {
+         var moleculeFactory = _individualMoleculeFactoryResolver.FactoryFor(individualMoleculeType);
+         //TODO use species?
+         expressionProfile.Individual = _individualFactory.CreateParameterLessIndividual();
```
msevestre, line 55: It always create a human for now

src/PKSim.Presentation/Core/MenuBarItemIds.cs
```diff
@@ -47,6 +47,9 @@ public static class MenuBarItemIds
       public static MenuBarItemId NewObserverSet = createMenuBarItemId("NewObserverSet");
       public static MenuBarItemId LoadObserverSet = createMenuBarItemId("LoadObserverSet");
       public static MenuBarItemId NewExpressionProfile = createMenuBarItemId("NewExpressionProfile");
+      public static MenuBarItemId NewExpressionProfileEnzyme = createMenuBarItemId("NewExpressionProfileEnzyme");
+      public static MenuBarItemId NewExpressionProfileTransporter = createMenuBarItemId("NewExpressionProfileTransporter");
+      public static MenuBarItemId NewExpressionProfileSpecificBindingPartner = createMenuBarItemId("NewExpressionProfileSpecificBindingPartner");
```
msevestre, line 52: This does then something like this

image


or like this

image

src/PKSim.Presentation/Presenters/Main/BuildingBlockExplorerPresenter.cs
```diff
@@ -109,8 +109,8 @@ protected override void AddProjectToTree(IProject project)
       private ITreeNode addProtocolToTree(Protocol protocol) => addBuildingBlockToTree(protocol, PKSimRootNodeTypes.ProtocolFolder, ApplicationIcons.Protocol);
 
       private ITreeNode addObserverSetToTree(ObserverSet observerSet) => addBuildingBlockToTree(observerSet, PKSimRootNodeTypes.ObserverSetFolder, ApplicationIcons.Observer);
-      
-      private ITreeNode addExpressionProfileToTree(ExpressionProfile expressionProfile) => addBuildingBlockToTree(expressionProfile, PKSimRootNodeTypes.ExpressionProfileFolder, ApplicationIcons.ProteinExpression);
+
+      private ITreeNode addExpressionProfileToTree(ExpressionProfile expressionProfile) => addBuildingBlockToTree(expressionProfile, PKSimRootNodeTypes.ExpressionProfileFolder, ApplicationIcons.IconByName(expressionProfile.Icon));
```
msevestre, line 113: Maybe it's enough to have them like this?
image

msevestre (commented):
@all: I am not sure how to bring the Load from Expression DB workflow. The species is required first. Maybe a button in the view? Feedback welcome

src/PKSim.Assets/PKSimConstants.cs
StephanSchaller: Category --> Phenotype? Sub-Population?
Settings --> Reference Population (--> move "Enzyme" to "Properties")
Properties --> Global Enzyme Properties

msevestre: Enzyme can't be move (at least not easily) to Properties as the "views" below will also be used in the Individual

msevestre:
What about something like this?

image

msevestre:
Database query would start the Protein Expression Query for the species selected or throw an error if none is defined for the given species

Yuri05:
>What about something like this?

I like this proposal. Later, when opening a creted profile for editing: the field "Species" would be read-only, correct?

msevestre:
Yes. In fact even Enzyme name and phenotype are readonly (can be changed via rename method). This is to ensure unicity

msevestre:
image

msevestre:
and 
image

msevestre:
Querying prefills with the name of the molecule

image

msevestre:
Next week, I will write tests and I think a good chunk of this feature is implemented :)

msevestre:
(Expression Profile onky that is...then we need to link it to the Individual)

src/PKSim.Presentation/Presenters/Main/BuildingBlockExplorerPresenter.cs
Yuri05: I think for the most proteins we will have only 1 phenotype within a project, thus for me it seems to be enough

msevestre: Agreed. I tried with a nested structure and it was less "usable" because one always had to open multiple folders

msevestre:
@Yuri05 Merging this PR for now. It will be part of the big feature branch commit then.
Next step. Linking Indiviudal to expression profile

PR #1950, opened 2021-11-09
Add expression profile reference to individual and organize serializa…

msevestre:
…tion


abdelr (approved):

src/PKSim.Infrastructure/Serialization/ORM/Mappers/ProjectToProjectMetaDataMapper.cs
```diff
@@ -164,11 +165,39 @@ public void Visit(Compound compound)
...
+         //no need to compress population data. This is mostly only double arrays
+         serializeContentFor(randomPopulation, compress: false);
+         _metaData.Properties.Data = _serializationManager.Serialize(randomPopulation.Settings);
+      }
+
+      public void Visit(ImportPopulation importPopulation)
+      {
+         _metaData = new ImportPopulationMetaData
+         {
+            ExpressionProfileIds = expressionProfileIdsFor(importPopulation)
```
msevestre, line 191: @Yuri05 I am saving the ids used in the DB as discussed. That way, I can load the bb on demand when required

src/PKSim.Infrastructure/Serialization/ORM/Mapping/RandomPopulationMetaDataMapping.cs
```diff
@@ -9,6 +9,7 @@ public RandomPopulationMetaDataMapping()
       {
          Table("RANDOM_POPULATIONS");
          KeyColumn("PopulationId");
+         Map(x => x.ExpressionProfileIds);
```
msevestre, line 12: THis needs to be done for Individual, RandomPopulation and Improt Population

src/PKSim.Infrastructure/Serialization/ORM/MetaData/BuildingBlockMetaData.cs
```diff
@@ -11,23 +11,58 @@ public abstract class BuildingBlockMetaData : ObjectBaseMetaDataWithProperties<B
    {
       public virtual int Version { get; set; }
       public virtual int StructureVersion { get; set; }
+
       public virtual string Icon { get; set; }
 
+      //High number by default to signify that loading does not matter
+      public virtual int LoadOrder { get; set; } = 1000;
```
msevestre, line 18: This was introduced so that I can load ExpressionProfile first

src/PKSim.Infrastructure/Serialization/Xml/Serializers/PopulationXmlSerializer.cs
```diff
@@ -15,20 +15,10 @@ public override void PerformMapping()
 
    public class RandomPopulationXmlSerializer : PopulationXmlSerializer<RandomPopulation>
    {
-      public override void PerformMapping()
-      {
-         base.PerformMapping();
-         Map(x => x.Settings);
```
msevestre, line 21: This is not required! I just realized it was saved twice all that time . It is saved as a separate column in the DB (see mapping)

src/PKSim.Infrastructure/Serialization/Xml/Serializers/PopulationXmlSerializer.cs
```diff
@@ -15,20 +15,10 @@ public override void PerformMapping()
...
-      }
    }
 
    public class ImportPopulationXmlSerializer : PopulationXmlSerializer<ImportPopulation>
    {
-      public override void PerformMapping()
-      {
-         base.PerformMapping();
-         Map(x => x.Settings);
```
msevestre, line 30: same here

src/PKSim.Presentation/Presenters/Individuals/MoleculesPresenter.cs
```diff
@@ -26,26 +26,6 @@ namespace PKSim.Presentation.Presenters.Individuals
    public interface IMoleculesPresenter :
       IPresenterWithContextMenu<ITreeNode>
    {
-      /// <summary>
-      ///    rename the given protein
-      /// </summary>
-      void RenameMolecule(IndividualMolecule molecule);
```
msevestre, line 32: There is probably even more code that will be removed once the feature is done

src/PKSim.Presentation/Services/EditMoleculeTask.cs
```diff
@@ -168,25 +149,26 @@ public ICommand EditMolecule(IndividualMolecule molecule, TSimulationSubject sim
...
-               simulationSubject.RemoveMolecule(newMolecule);
-               return new PKSimEmptyCommand();
-            }
-
-            var queryResults = presenter.GetQueryResults();
-            return _moleculeExpressionTask.AddMoleculeTo(simulationSubject, newMolecule, queryResults);
-         }
-      }
+      //
+      // private ICommand proteinFromQuery<TMolecule>(TSimulationSubject simulationSubject) where TMolecule : IndividualMolecule
```
msevestre, line 153: I kept it for now but we don't have this use case anymore. We always EDIT a profile

src/PKSim.Presentation/Services/RenameBuildingBlockTask.cs
```diff
@@ -105,46 +115,21 @@ private void renameIndividualSimulation(IndividualSimulation individualSimulatio
...
-      {
-         if (!path.Any()) return false;
-         return path.LastIndexOf(oldCompoundName) > 0;
-      }
 
-      public void SynchronizeCompoundNameIn(Simulation targetSimulation, string oldCompoundName, string newCompoundName)
+      public void RenameBuildingBlock(IPKSimBuildingBlock templateBuildingBlock, string oldBuildingBlockName)
       {
-         //the cache will be referencing quantity by path using the newCompoundName
-         var quantityCache = _containerTask.CacheAllChildren<IQuantity>(targetSimulation.Model.Root);
```
msevestre, line 117: This was never used. Removed

msevestre:
@Yuri05 REady for review. When this is merged, I'll tackle Individual Scaling, filtering of expression profile to only allow the right species and then project conversion

src/PKSim.Core/Model/PKSimBuildingBlock.cs
```diff
@@ -20,6 +20,13 @@ public interface IPKSimBuildingBlock : ILazyLoadable, IVersionable, IContainer,
       ///    Indicates if a loaded building block was changed and hence needs to be saved
       /// </summary>
       bool HasChanged { get; set; }
+
+      /// <summary>
+      /// A building block might use building block internally to hold structure
+      /// (ExpressionProfile or Population have an internal individual instance).
+      /// In this case, this would be the reference to the parent building block
+      /// </summary>
+      PKSimBuildingBlock OwnedBy { get; set; }
```
Yuri05, line 29: so here e.g. an Expression profile within an Individual-BB would be OwnedBy this Individual-BB?

Yuri05:
Looks good. The only question is that about renaming.

src/PKSim.Core/Model/PKSimBuildingBlock.cs
msevestre: No. The problem is the following
ExpressionProfile owns an Individual (the organism structure) which olds the enzyme. When changing an expression parameter, it believes that the parameter changed is in the Individual but in fact it's in the ExpressionPRofile whose version should change etc.. So this was required. It is only used when creating the Individual for the expression profile

src/PKSim.Core/Model/ExpressionProfile.cs
```diff
@@ -45,22 +59,21 @@ public override string Name
                return;
 
             var names = CoreConstants.NamesFromCompositeName(value);
-            if (names.Count != 2)
+            if (names.Count != 3)
                return;
 
             _moleculeName = names[0];
-            _category = names[1];
+            _category = names[2];
```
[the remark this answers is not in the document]
msevestre: yes. There is a specifc presneter for this
image

msevestre:
ok then I am going to merge

src/PKSim.Core/Model/ExpressionProfile.cs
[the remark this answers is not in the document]
Yuri05: That's good. However I still don't understand why we have the Name setter.
`Name.get` should just return {Protein}-{Species}-{Phenotype}.
Especially now if I e.g. set Protein name = "P-gp" or Phenotype="Disease - stage 1" it will not work, because the value is splitted by "-".

PR #1964, opened 2021-11-17
1921 do not delete uses expression profile

msevestre:


msevestre (commented):
Making sure that we cannot delete an expression if it's being used in an individual, a population or a simulation.
BTW: We need to add the expression profile to the list of used building block in a simulation right?

PR #1972, opened 2021-11-23
Fixes #1963 multi select expression profile

### LINKED ISSUE #1963: Implement Multi Select action for Expression Profile

msevestre:


src/PKSim.Core/Services/IndividualModelTask.cs
```diff
@@ -53,6 +64,11 @@ public void CreateModelFor(Individual individual)
          _buildingBlockFinalizer.Finalize(individual);
       }
 
+      public void CreateOrganStructureFor(Individual individual)
```
msevestre, line 67: @Yuri05 In fact for ExpressionProfile, I can even make it better by not using neighbourhood. I just need the container structure. The "internal" individual will be more smaller. No parameters, no neighborhood, just contianers

msevestre:
@Yuri05 Trivial changes. Merging

PR #1975, opened 2021-11-24
Fixes #1974 set name automatically

### LINKED ISSUE #1974: QueryProtein from DB. If name of protein is empty, set name (and ontogeny) automatically

msevestre:


src/PKSim.Core/Model/ExpressionProfile.cs
```diff
@@ -34,15 +33,12 @@ public ExpressionProfile() : base(PKSimBuildingBlockType.ExpressionProfile)
...
-         {
-            _category = value;
-            RefreshName();
-         }
+         set => _category = value;
       }
 
       public override string Name
       {
+         get => ExpressionProfileName(MoleculeName, Species, Category);
```
msevestre, line 41: All right, I have removed this also... I am not sure if this is going to be a problem with serialization but I don't think so. That way, we never have to worry about sync

src/PKSim.Core/Services/ExpressionProfileUpdater.cs
```diff
@@ -112,7 +128,7 @@ public ICommand UpdateMoleculeName(ExpressionProfile expressionProfile, string n
 
       private IOSPSuiteCommand renameMolecule(IndividualMolecule molecule, ISimulationSubject simulationSubject, string newMoleculeName)
       {
-         return _individualExpressionTask.RenameMolecule(molecule, newMoleculeName, simulationSubject.Individual);
+         return new RenameMoleculeInSimulationSubjectCommand(molecule, simulationSubject, newMoleculeName, _executionContext).Run(_executionContext);
```
msevestre, line 131: @Yuri05 I hae moved around some of the methods that were in molecule specific stuff but in fact it was really for expression profile

src/PKSim.Core/Services/MoleculeExpressionTask.cs
```diff
@@ -167,27 +87,5 @@ public ICommand SetTransporterTypeFor(IndividualTransporter transporter, Transpo
       {
          return new SetTransportTypeCommand(transporter, transportType, _executionContext).Run(_executionContext);
       }
-
-      private void setDefaultFor(IndividualMolecule molecule, TSimulationSubject simulationSubject, string moleculeName)
```
msevestre, line 171: All of this stuff has moved

src/PKSim.Presentation/Presenters/ExpressionProfiles/CreateExpressionProfilePresenter.cs
```diff
@@ -39,6 +39,8 @@ public class CreateExpressionProfilePresenter : AbstractSubPresenterContainerPre
          if (_view.Canceled)
             return new PKSimEmptyCommand();
 
+         //Make sure we save all information that was edited by the user
+         _subPresenterItemManager.AllSubPresenters.Each(x => x.Save());
```
msevestre, line 43: This needs to be triggered so that we do not have the problem from before

src/PKSim.Presentation/Presenters/ExpressionProfiles/EditExpressionProfilePresenter.cs
```diff
@@ -33,7 +33,7 @@ public override void Edit(ExpressionProfile expressionProfile)
       {
          _expressionProfile = expressionProfile;
          _subPresenterItemManager.AllSubPresenters.Each(x => x.Edit(_expressionProfile));
-         _subPresenterItemManager.PresenterAt(ExpressionProfileItems.Molecules).DisableSettings();
+         _subPresenterItemManager.PresenterAt(ExpressionProfileItems.Molecules).IsEditMode = true;
```
msevestre, line 36: I am only updating name and default based when we CREATE not when we EDIT

src/PKSim.Presentation/Presenters/ExpressionProfiles/ExpressionProfileMoleculesPresenter.cs
```diff
@@ -57,30 +63,49 @@ public void Edit(ExpressionProfile expressionProfile)
...
 
-         AddCommand(_editMoleculeTask.EditMolecule(_expressionProfile.Molecule, _expressionProfile.Individual, _expressionProfileDTO.MoleculeName));
+         var queryResults = _expressionProfileProteinDatabaseTask.QueryDatabase(_expressionProfile, _expressionProfileDTO.MoleculeName);
+         if (queryResults == null)
+            return;
+
+         if (!IsEditMode)
+         {
+            _expressionProfileDTO.MoleculeName = queryResults.ProteinName;
+            _moleculeParameterTask.SetDefaultFor(_expressionProfile, _expressionProfileDTO.MoleculeName);
```
msevestre, line 104: This is a bit of a breaking change compared to before. Entering the name CYP3A4 without using the DB will not set the default value...I am not sure how to do this

Yuri05: You mean that if you e.g. set the name = CYP3A4 **manually**, then ReferenceConcentration, HalfTimes etc. of CYP3A4 will not be used?

msevestre: Yes this is the problem. Before the input was a modal so we knew what to set

msevestre:
The question of default for manually input name still needs to be addressed

PR #1995, opened 2021-11-30
Fixes #1979 create expression profile

### LINKED ISSUE #1979: How to deal with default for Enzyme/Transporter when name is entered by hand (Without query)
see https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/1975#discussion_r756431305

Before: USer would enter the name in a modal view and from there, we could match the name to existing molecule, ontogeny etc and set some default, even if the data was not read from the protein expression

Now: Everything happens in the modal view

Option 1: We change the flow. ADding a new only show the view where you can select
Enzyme name, phenotype and Species (similar to clojne) . Then you are in the edit view

Option 2: we do not change anything

Option 3 ?

msevestre:


src/PKSim.Core/Repositories/UsedMoleculeRepository.cs
```diff
@@ -13,19 +13,28 @@ public interface IUsedMoleculeRepository : IRepository<string>
...
+         _moleculeParameterRepository = moleculeParameterRepository;
       }
 
       public IEnumerable<string> All()
       {
-         return allMoleculesDefinedInIndividuals()
-            .Union(allMoleculesDefinedInCompounds())
+         //First add User defined molecules
+         return allMoleculesDefinedInCompounds()
             .Union(allExpressionProfiles())
```
msevestre, line 33: Project molecule first and then predefined

PR #2002, opened 2021-12-01
Investigate differences batch

msevestre:


msevestre:
@Yuri05 REsults are now the same. v9 v10 and v11

PR #2005, opened 2021-12-03
2004: Update Create Individual Interface to support disease state

msevestre:


src/Db/Diff/PKSimDB_diff.txt
```diff
@@ -17,6 +17,7 @@ CREATE TABLE tab_disease_states
     implementation text
 );
 INSERT INTO tab_disease_states(disease_state,display_name,description,implementation) VALUES('CKD','CKD Display TODO','CKD Description TODO','CKDDiseaseState');
+INSERT INTO tab_disease_states(disease_state,display_name,description,implementation) VALUES('Healthy','Healthy Display TODO','Healthy Description TODO',NULL);
```
msevestre, line 20: Also added this disease state to show the Healthy state if required

src/PKSim.Core/Model/Individual.cs
```diff
@@ -53,16 +53,7 @@ public Individual() : base(PKSimBuildingBlockType.Individual)
...
       public virtual bool IsHuman => Species.IsHuman;
 
-      public virtual bool IsPreterm
-      {
-         get
-         {
-            if (!Population.IsPreterm)
-               return false;
-
-            return OriginData.GestationalAge.HasValue && OriginData.GestationalAge.Value <= CoreConstants.NOT_PRETERM_GESTATIONAL_AGE_IN_WEEKS;
```
msevestre, line 63: @Yuri05 This is interesting here. Look at the constraint. We have 
.GestationalAge.Value **<=** CoreConstants.NOT_PRETERM_GESTATIONAL_AGE_IN_WEEKS;

But in another place in the code, we use **<** (string inferior) and I am not sure why.

src/PKSim.Core/Model/OriginData.cs
```diff
@@ -1,58 +1,134 @@
+using System.Collections.Generic;
+using System.Security.Cryptography.X509Certificates;
+using NPOI.HSSF.Record;
 using OSPSuite.Core.Domain;
 using PKSim.Assets;
 
 namespace PKSim.Core.Model
 {
+   public class OriginDataParameter
```
msevestre, line 9: I have added this new class. Maybe better as InputParameter. It's always the value and unit. We will need a project conversion but it's much easier. Dynamic parameters coming from DiseaseState are also such parameters

src/PKSim.Core/Model/OriginData.cs
```diff
@@ -1,58 +1,134 @@
...
+      public virtual bool IsPreterm
+      {
+         get
+         {
+            var gaValue = GestationalAge?.Value;
+            return gaValue.GetValueOrDefault(CoreConstants.NOT_PRETERM_GESTATIONAL_AGE_IN_WEEKS) <= CoreConstants.NOT_PRETERM_GESTATIONAL_AGE_IN_WEEKS;
+         }
+      }
 
+      public virtual bool IsRealPreterm
```
msevestre, line 98: I have added this method with the < but I am not sure why we have this difference. @Yuri05 Can you remember?

src/PKSim.Core/Model/ParameterDistributionMetaDataExtensions.cs
```diff
@@ -12,11 +12,11 @@ public static IReadOnlyList<ParameterDistributionMetaData> DefinedFor(this IEnum
          {
             var allNonPreterm = metaDataForParameter.allForGA(CoreConstants.NOT_PRETERM_GESTATIONAL_AGE_IN_WEEKS).ToList();
 
-            if (!isRealPreterm(originData))
+            if (!originData.IsRealPreterm)
```
msevestre, line 15: here. using this weird method

src/PKSim.Infrastructure/Serialization/Xml/Serializers/OriginDataXmlSerializer.cs
```diff
@@ -2,6 +2,16 @@
...
+   public class OriginDataParameterXmlSerializer : BaseXmlSerializer<OriginDataParameter>
+   {
+      public override void PerformMapping()
+      {
+         Map(x => x.Value);
+         Map(x => x.Unit);
+      }
+   }
+
    public class OriginDataXmlSerializer : BaseXmlSerializer<OriginData>
```
msevestre, line 15: Xml conversion required here (issue created) #2006

src/PKSim.Presentation/DTO/Individuals/IndividualDefaultValuesUpdater.cs
```diff
@@ -0,0 +1,121 @@
...
+using OSPSuite.Utility.Extensions;
+using PKSim.Core.Model;
+using PKSim.Core.Repositories;
+using PKSim.Core.Services;
+using PKSim.Presentation.DTO.Mappers;
+using PKSim.Presentation.DTO.Parameters;
+
+namespace PKSim.Presentation.DTO.Individuals
+{
+   public interface IIndividualDefaultValueUpdater
```
msevestre, line 13: I simplified this guy a lot. Some definite duplicate code was removed

src/PKSim.Presentation/DTO/Individuals/IndividualSettingsDTO.cs
```diff
@@ -14,12 +14,18 @@ public class IndividualSettingsDTO : IWithValueOrigin
...
       public IEnumerable<CategoryCalculationMethodDTO> CalculationMethods { get; set; }
       public Species Species { get; set; }
       public Gender Gender { get; set; }
       public ValueOrigin ValueOrigin { get; }
+      public DiseaseState DiseaseState { get; set; }
+      
+      /// <summary>
+      /// For now, we assume only one disease state parameters. We can make the view more generic if this is ever required
+      /// </summary>
+      public IParameterDTO DiseaseStateParameter { get; set; }
```
msevestre, line 28: I am assuming for now that we only have ONE parameter as I did not want to spend too much time dealing with multiple entries in the view etc.. if we have a disease state with multiple inputs...then we can update the DTO (keep it simple). The model has multiples parameters. Just the view shows one

src/PKSim.Presentation/Views/Individuals/IIndividualSettingsView.cs
```diff
@@ -12,13 +12,15 @@ public interface IIndividualSettingsView : IView<IIndividualSettingsPresenter>,
...
       void BindToParameters(IndividualSettingsDTO individualSettingsDTO);
+      void BindToDiseaseState(IndividualSettingsDTO individualSettingsDTO);
       void BindToSubPopulation(IEnumerable<CategoryParameterValueVersionDTO> subPopulation);
       void RefreshAllIndividualList();
       bool AgeVisible { get; set; }
       bool HeightAndBMIVisible { get; set; }
       bool IsReadOnly { get; set; }
       bool SpeciesVisible { get; set; }
       bool GestationalAgeVisible { get; set; }
+      bool DiseaseStateVisible { get; set; }
```
msevestre, line 23: Following the logic. I would probably do that differntly nowdays but this is good enough

msevestre (commented):
This is how it looks

Human default
image

Then CKD
image

It's hidden for species/pop without any disease state

msevestre:
Tests are missing.

src/PKSim.Core/Model/Individual.cs
Yuri05: No idea. 
But actually there is a wish since a long time to restrict preterms creation not by GA<=40 , but by GA<=37 (or 38)

src/PKSim.Core/Model/OriginData.cs
```diff
@@ -1,69 +1,144 @@
...
+      }
 
-      public virtual string Comment { get; set; }
+      public bool IsRealPreterm
+      {
+         get
+         {
+            //WHY THE DIFFERENCE?
+            var gaValue = GestationalAge?.Value;
+            return gaValue.GetValueOrDefault(CoreConstants.NOT_PRETERM_GESTATIONAL_AGE_IN_WEEKS) < CoreConstants.NOT_PRETERM_GESTATIONAL_AGE_IN_WEEKS;
```
Yuri05, line 105: This is weird. Where is IsPreterm used and where "IsRealPreterm"?

src/PKSim.Core/Model/ParameterDistributionMetaDataExtensions.cs
Yuri05: ok, I think I know why. 
Some parameters do not differ between different GAs (or at least we just don't have enough information to inform the parameter) and we did not want to create N copies of the same parameter for all possible GAs.
And for those parameters DistributionMetaData is only available for GA=40 (="NOT_PRETERM_GESTATIONAL_AGE_IN_WEEKS")
Thus if we have a preterm of GA=40 (which in fact is not really a preterm :)) - we just take all parameter distributions.
If we have a preterm of GA<40: for each parameter we check first if the distribution data **for the given GA** is available. If yes - take it, if no - take the distribution for GA=40.

src/PKSim.Core/Model/IndividualFactory.cs
```diff
@@ -57,6 +60,10 @@ public Individual CreateAndOptimizeFor(OriginData originData, int? seed = null)
             individual.Seed = seed.Value;
 
          _createIndividualAlgorithm.Optimize(individual);
+
+         var diseaseStateImplementation = _diseaseStateImplementationFactory.CreateFor(individual);
+         diseaseStateImplementation.ApplyTo(individual);
```
Yuri05, line 65: For a healthy individual it will be a "Do nothing"-implementation, correct?

src/PKSim.Core/Repositories/IDiseaseStateRepository.cs
```diff
@@ -8,5 +8,7 @@ namespace PKSim.Core.Repositories
    public interface IDiseaseStateRepository : IStartableRepository<DiseaseState>
    {
       IReadOnlyList<DiseaseState> AllFor(SpeciesPopulation population);
+
+      DiseaseState HealthyState { get; }
```
Yuri05, line 12: Health**y**State or HealthState?

src/PKSim.Core/Model/Individual.cs
msevestre: ok. Can we create another issue for this?

src/PKSim.Core/Model/ParameterDistributionMetaDataExtensions.cs
msevestre: ok I see. Maybe we can make this a bit clearer in the code

src/PKSim.Core/Model/IndividualFactory.cs
msevestre: yes

src/PKSim.Core/Repositories/IDiseaseStateRepository.cs
msevestre: This is the ONE instance of HealthyState. Correct here with *Healthy*

msevestre:
Ok I am going to merge this and start with the Projet Conversion

msevestre:
@Yuri05 Merging into feature branch

src/PKSim.Core/Model/Individual.cs
Yuri05: Think we have an issue for this already.. but maybe not. Will check.

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
```diff
@@ -0,0 +1,154 @@
...
+         var fatVolume = fat.Parameter(VOLUME);
+         var stomach = organism.EntityAt<IContainer>(LUMEN, STOMACH);
+         var plasmaProteinScaleFactorParameter = organism.Parameter(PLASMA_PROTEIN_SCALE_FACTOR);
+         var smallIntestinalTransitTime = smallIntestine.Parameter(SMALL_INTESTINAL_TRANSIT_TIME);
+         var gastricEmptyingTime = stomach.Parameter(GASTRIC_EMPTYING_TIME);
+
+         var bsaInm2 = bsa.ConvertToUnit(bsa.Value, CoreConstants.Units.m2);
+         var GFR_0 = GFR_spec.Value * kidneyVolume.Value * 1.73 / bsaInm2;
+         if (GFR_0 < targetGFR.Value)
+            throw new OSPSuiteException("ERROR");
```
Yuri05, line 59: TODO: better error message

Yuri05: I wonder what should we do if `GFR_0 = TargetGFR`? Should we continue or exit?
The most scaling factors would become 1 in this case, however some of them (e.g. hematocrit) might be <>1.
But in fact `GFR_0 = TargetGFR` means healthy individual, so I would not expect any further parameter changes in this case. Maybe smthg we should check with Paul.

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
```diff
@@ -0,0 +1,154 @@
...
+         //Adjust kidney volume and update fat accordingly
+         var healthyKidneyVolume = kidneyVolume.Value;
+         kidneyVolume.Value = getKidneyVolumeFactor(targetGFR.Value) / getKidneyVolumeFactor(GFR_0) * healthyKidneyVolume;
+         fatVolume.Value = fatVolume.Value + healthyKidneyVolume - kidneyVolume.Value;
+
+         //Adjust rena blood flow
+         kidneySpecificBloodFlowRate.Value = getRenalBloodFlowFactor(targetGFR.Value) / getRenalBloodFlowFactor(GFR_0) * kidneySpecificBloodFlowRate.Value;
+
+         //Correct specific GFR
+         GFR_spec.Value = GFR_spec.Value * targetGFR.Value / GFR_0 * healthyKidneyVolume / kidneyVolume.Value;
```
Yuri05, line 70: if regarding the question above we decide still to apply changes in case `GFR_0 = TargetGFR`: I would put the lines 61-70 at the end of the routine and put the check before: (` if (GFR_0 == TargetGFR) return`).
Because those factors will be 1, so no need to adjust (and also no need to change the value origin)

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
```diff
@@ -0,0 +1,154 @@
...
+         if (value <= 20)
+            diseaseValue = isFemale ? 33.5 : 34.3;
+         else if (value <= 30)
+            diseaseValue = isFemale ? 37.4 : 41.7;
+         else if (value <= 40)
+            diseaseValue = isFemale ? 38.4 : 42.6;
+         else if (value <= 50)
+            diseaseValue = isFemale ? 39.7 : 43.3;
+         else
+            diseaseValue = isFemale ? 39.6 : 44.8;
```
Yuri05, line 102: targetGFR is <=60 per construction, correct?

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
```diff
@@ -0,0 +1,154 @@
...
+
+         return diseaseValue / healthyValue;
+      }
+
+      private void updateValueOriginsFor(params IParameter[] parameters)
+      {
+         parameters.Each(x => x.ValueOrigin.UpdateAllFrom(_valueOriginRepository.FindBy(CKD_VALUE_ORIGIN_ID)));
+      }
+
+      private double getKidneyVolumeFactor(double gfrValue) => getFactor(gfrValue, 6.3 * 10E-5, 0.0149, 4.13);
```
Yuri05, line 112: must be `6.3 * 1E-5` , not `6.3 * 10E-5`

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
```diff
@@ -0,0 +1,154 @@
...
+
+      private double getRenalBloodFlowFactor(double gfrValue) => getFactor(gfrValue, -3.1 * 10E-5, 0.0170, 4.09);
```
Yuri05, line 114: same

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
```diff
@@ -0,0 +1,154 @@
...
+         if (value < 30)
+            return CKDStage.Stage4;
+
+         return CKDStage.Stage3;
+      }
+
+      private double getTargetGFRValue(OriginDataParameter targetGFR)
+      {
+         var dimension = _dimensionRepository.DimensionForUnit(targetGFR.Unit);
+         return dimension.BaseUnitValueToUnitValue(dimension.Unit(targetGFR.Unit), targetGFR.Value);
```
Yuri05, line 136: this will return the value in `mL/min/1.73m²`, correct?

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
```diff
@@ -0,0 +1,154 @@
...
+         var kidneySpecificBloodFlowRate = kidney.Parameter(SPECIFIC_BLOOD_FLOW_RATE);
+         var fatVolume = fat.Parameter(VOLUME);
+         var stomach = organism.EntityAt<IContainer>(LUMEN, STOMACH);
+         var plasmaProteinScaleFactorParameter = organism.Parameter(PLASMA_PROTEIN_SCALE_FACTOR);
+         var smallIntestinalTransitTime = smallIntestine.Parameter(SMALL_INTESTINAL_TRANSIT_TIME);
+         var gastricEmptyingTime = stomach.Parameter(GASTRIC_EMPTYING_TIME);
+
+         var bsaInm2 = bsa.ConvertToUnit(bsa.Value, CoreConstants.Units.m2);
+         var GFR_0 = GFR_spec.Value * kidneyVolume.Value * 1.73 / bsaInm2;
+         if (GFR_0 < targetGFR.Value)
```
Yuri05, line 58: Not sure if we don't miss a unit conversion here. GFR_0 is in **L/min** but TargetGFR is in **L/min/1.73m²**

msevestre: Yeah could be. The actual implementation is a big Todo at the moment. I just wanted to get the infrastructure going. I need to get some test data

PR #2012, opened 2021-12-08
Fixes #2010

### LINKED ISSUE #2010: CKD Disease State Implementation for Population

msevestre:


src/PKSim.Core/CoreConstants.cs
```diff
@@ -1176,6 +1176,7 @@ public static bool IsVolumeFraction(string parameterName)
                BLOOD_FLOW,
                GFR_SPEC,
                VOLUME_PLASMA,
+               E_GFR
```
msevestre, line 1179: @Yuri05 I think we want to be ableto see the distribution of target gfr in our population.

src/PKSim.Core/Mappers/IndividualToPopulationSettingsMapper.cs
```diff
@@ -31,42 +32,36 @@ public RandomPopulationSettings MapFrom(Individual individual)
...
          if (population.IsHeightDependent)
+            populationSettings.AddParameterRange(parameterRangeFrom(organism.Parameter(BMI)));
+
+         individual.OriginData.DiseaseStateParameters.Each(x =>
          {
-            var bmiParameter = individual.Organism.Parameter(CoreConstants.Parameters.BMI);
-            populationSettings.AddParameterRange(parameterRangeFrom(bmiParameter));
-         }
+            var parameter = individual.OriginData.DiseaseState.Parameter(x.Name);
+            populationSettings.AddParameterRange(constrainedParameterRangeFrom(parameter));
```
msevestre, line 63: Adding all disease state paraemters as constrained parmaeters (min/max)

src/PKSim.Core/Model/RandomPopulationFactory.cs
```diff
@@ -75,11 +90,12 @@ public Task<RandomPopulation> CreateFor(RandomPopulationSettings populationSetti
                fllUpGenderQueueBasedOn(populationSettings);
                progressUpdater.Initialize(populationSettings.NumberOfIndividuals, PKSimConstants.UI.CreatingPopulation);
 
+               var diseaseStateImplementation = _diseaseStateImplementationFactory.CreateFor(populationSettings.BaseIndividual);
                //the base individual is used to retrieve the default values. 
-               var baseIndividual = populationSettings.BaseIndividual;
+               var baseIndividual = diseaseStateImplementation.CreateBaseIndividualForPopulation(populationSettings.BaseIndividual);
```
msevestre, line 95: First hook that we discussed. Return the base individual for the population. I am not sure how generic this will be

src/PKSim.Core/Model/RandomPopulationFactory.cs
```diff
@@ -105,21 +121,28 @@ public Task<RandomPopulation> CreateFor(RandomPopulationSettings populationSetti
...
-                  updateCurrentIndividualFromSettings(populationSettings, currentIndividual, allDistributedParameters, allBaseDistributedParameters, currentGender, randomPopulation.RandomGenerator);
+                  updateCurrentIndividualFromSettings(populationSettings, baseIndividual, currentIndividual, allDistributedParameters, allBaseDistributedParameters, currentGender, randomPopulation.RandomGenerator);
 
-                  bool success = tryRandomize(currentIndividual, populationSettings, allCurrentIndividualParameters, randomPopulation.RandomGenerator);
-                  if (!success) continue;
+                  var success = tryRandomize(currentIndividual, populationSettings, allCurrentIndividualParameters, randomPopulation.RandomGenerator);
+                  if (!success)
+                     continue;
+
+                  success = diseaseStateImplementation.ApplyForPopulationTo(currentIndividual);
```
msevestre, line 130: Second Hook. Instead of Apply that overwrites also the formula, I need to have a method that only overwrites the value in the indivudal so that I can reset them later

src/PKSim.Core/Model/RandomPopulationFactory.cs
```diff
@@ -105,21 +121,28 @@ public Task<RandomPopulation> CreateFor(RandomPopulationSettings populationSetti
...
+                  if (!success)
+                     continue;
 
                   randomPopulation.AddIndividualValues(_individualValuesMapper.MapFrom(currentIndividual, allChangedByCreatedIndividualParameters));
 
+                  diseaseStateImplementation.ResetParametersAfterPopulationIteration(currentIndividual);
```
msevestre, line 136: For instance GFR_Spec is changed but needs to be set back to a formula

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
```diff
@@ -60,33 +80,64 @@ public void ApplyTo(Individual individual)
...
+
+      private IReadOnlyList<IParameter> parametersChangedByCKDAlgorithmAsList(Individual individual)
+      {
+         var p = parametersChangedByCKDAlgorithm(individual);
 
-         var GFR_0 = getTargetGFRValue(GFR_spec.Value * kidneyVolume.Value / bsa.Value);
+         return new[] {p.hct, p.GFR_spec, p.kidneyVolume, p.kidneySpecificBloodFlowRate, p.fatVolume, p.plasmaProteinScaleFactorParameter, p.smallIntestinalTransitTime, p.gastricEmptyingTime};
+      }
+
+      public bool ApplyTo(Individual individual) => applyTo(individual, updateParameter);
```
msevestre, line 93: Same algorithm. Only different way to apply the values...to discuss..

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
```diff
@@ -111,6 +162,53 @@ private void updateParameter(IParameter parameter, double value)
          parameter.IsFixedValue = false;
       }
 
+      public Individual CreateBaseIndividualForPopulation(Individual originalIndividual)
```
msevestre, line 165: @Yuri05 I am not sure about this. This looks very complicated.
 I just want one healthy individual with all the parameter changed from the user. I do not update parameters changed by the CKD parameter becaus ethat does not really make sense...does it? Say I change the volume of kidney to be bigger in my base individual and then use the individual to create CKD patients... this is kind of crazy

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
```diff
@@ -111,6 +162,53 @@ private void updateParameter(IParameter parameter, double value)
...
+         //we need to create a new individual WITHOUT CKD and set all percentiles as in the original individuals. Other parameters, value wil be taken as is
+         var originData = originalIndividual.OriginData.Clone();
+
+         //remove the disease state to create a healthy Individual
+         originData.DiseaseState = null;
+         var healthyIndividual = _individualFactory.CreateAndOptimizeFor(originData, originalIndividual.Seed);
+
+         var allCKDParameters = parametersChangedByCKDAlgorithmAsList(healthyIndividual);
+
+         var allHealthyDistributedParameters = _containerTask.CacheAllChildrenSatisfying<IDistributedParameter>(healthyIndividual, x => !allCKDParameters.Contains(x));
```
msevestre, line 176: @Yuri05 I am not sure that we need to differentiate the percentile. This is what we discussed a few days ago but I can't see why anymore

msevestre:
I am creating population CKD all right. But no idea how to evaluate. Also the last point in the slide "Adapt Population Distributions" cannot really be done as we generate healthy individual and then apply the algorithm

msevestre:
image

src/PKSim.Core/Model/RandomPopulationFactory.cs
Yuri05: >For instance GFR_Spec is changed but needs to be set back to a formula 

Why? And what happens if in my baseIndividual GFR_spec was ovewritten with a value by the user?
At the moment (with healthy populations): if we overwrite a GFR_spec in the individual and create a population based on it: all individuals in the pop get this GFR_spec value entered by user.

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
Yuri05: to be discussed. while indeed it does not make much sense to change GFR_spec in the individual (s. the question above) - technically it still can be done. And for Volume_Kidney I could even imagine it might make sense.
So maybe in such a case we still additionaly multiply by the factor Value/DefaultValue? Not sure

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
```diff
@@ -111,6 +162,53 @@ private void updateParameter(IParameter parameter, double value)
...
+         var allOriginalDistributedParameters = _containerTask.CacheAllChildren<IDistributedParameter>(originalIndividual);
+
+         //update percentile for distributed parameters
+         foreach (var keyValue in allHealthyDistributedParameters.KeyValues)
+         {
+            var originalParameter = allOriginalDistributedParameters[keyValue.Key];
+            _parameterTask.SetParameterPercentile(keyValue.Value, originalParameter.Percentile);
+         }
+
+         //now update all other parameters that may have been changed by the user. Allowing the user to overwrite CKD Parameters is probably a recipe for disaster.
```
Yuri05, line 186: I think we should differentiate and maybe indeed set some parameters to readonly in this case?
* While the TargetGFR is an input, it probably does not make much sense to overwrite parameters which are used for the calculation of the GFR: GFR_spec, Volume_Kidney, etc.
* All other parameters which are changed by CKD-Algo but not used for the GFR-Calculation (Hematocrit, etc.) - don't see a problem if the user overwites them

But this is something I would really like to discuss with Paul, Jan, etc.

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
Yuri05: Because if we shift a distributed parameter in the baseIndividual, we want the same shif

src/PKSim.Core/Model/RandomPopulationFactory.cs
msevestre: It needs to be reset to a formula because otherwise there is no way to get the target eGFR. it simply does not make sense to change those parameters if we are using CKD.

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
msevestre: Technically it can still be done but at the end of the day, GFR_spec will be overwritten according to the selected algorithm. I think this is fair and it can be explained in the manual. And changing the volume of kidney does not make sense since it changes the eGFR which is an input of the user.

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
msevestre: >All other parameters which are changed by CKD-Algo but not used for the GFR-Calculation (Hematocrit, etc.) - don't see a problem if the user overwites them

I think it does not make sense. I talked briefly with @aedginto about this yesterday. The publication describes changes from Healthy individual. We should stick with what we can explain. Because otherwise the algorithm will multiply a factor to the value that was changed by the user.

Yuri05: To be discussed. But for me it makes sense. 
If you e.g. have a hematocrit measurement for your CKD individual - why not to allow to enter it and ovewrite the "CKD_Adjusted healthy hematocrit value"?

msevestre: For an individual yes.. But not for a population

src/PKSim.Core/Model/RandomPopulationFactory.cs
```diff
@@ -243,6 +269,15 @@ private void perturbate(Individual currentIndividual, RandomPopulationSettings p
...
             }
+
+            //now assign any random parameters due to disease state
+            foreach (var diseaseStateParameter in originData.DiseaseStateParameters)
+            {
+               diseaseStateParameter.Value = createDiseaseStateRandomParameterValueFor(originData, populationSettings, diseaseStateParameter, randomGenerator, out success);
+            }
+
+            if (!success) continue;
+            (success, _) = diseaseStateImplementation.IsValid(originData);
```
msevestre, line 280: Here I am also adding validation on specific case. for Instance age >=18 for CKD

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
msevestre: I simplified the implementation by relying on the Parameter Updater

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
```diff
@@ -111,6 +163,55 @@ private void updateParameter(IParameter parameter, double value)
...
+      public void Validate(OriginData originData)
+      {
+         var (valid, error) = IsValid(originData);
+         if (valid)
+            return;
+
+         throw new OSPSuiteException(error);
+      }
+
+      public (bool isValid, string error) IsValid(OriginData originData)
```
msevestre, line 206: Validation. Simple stuff

src/PKSim.Core/Services/CKDDiseaseStateImplementation.cs
Yuri05: >Unless we say that we do not apply changes for parmeter such as HCT changed by the user and they stay as is. Which also does not makle sense because HCT is gender dependent etc..

True. This is something I would not do for sure.

>specifically because you would apply a change twice.

Maybe this is exactly the expectation? Hard to judge for me...

msevestre:
@Yuri05 I am going to merge this beacuse I am about to make bigger changes in the DB for the  IsChangedBycreateIndividual stuff. Your questions are defined as separate issue so we can address them in other PR

PR #2043, opened 2022-01-12
Fixes #1939 template db auto update

### LINKED ISSUE #1939: Template Database. Automatic update from github
At the moment, the file is defined a submodule. Upon start, PK-Sim should check if there is a new version available and if yes, download them and copy it in the user profile folder

msevestre:


src/PKSim.Core/CoreConstants.cs
```diff
@@ -80,19 +80,21 @@ public static class Filter
       public static readonly string TEMPLATE_USER_DATABASE_TEMPLATE = "PKSimTemplateDBUser.template";
       public static readonly string REMOTE_TEMPLATE_SUMMARY = "templates.json";
       public const string PRODUCT_NAME = "PK-Sim";
+      public const string TEMPLATES_PRODUCT_NAME = "BuildingBlockTemplates";
       public static readonly string PRODUCT_NAME_WITH_TRADEMARK = "PK-Sim®";
       public static readonly string DEFAULT_SKIN = "Office 2013 Light Gray";
       public static readonly string VALUE_PROPERTY_NAME = "Value";
       public static readonly string PROJECT_UNDEFINED = "Undefined";
       public static readonly string VERSION_FILE_URL = "https://raw.githubusercontent.com/Open-Systems-Pharmacology/Suite/master/versions.json";
+      public static readonly string REMOTE_TEMPLATE_FILE_URL = "https://raw.githubusercontent.com/Open-Systems-Pharmacology/OSPSuite.BuildingBlockTemplates/main/templates.json";
```
msevestre, line 89: This is the path of the latest file released. When we release a new file, we need to remember to also update the version file in the suite that will be the trigger for the update

Yuri05: we should setup the **develop** branch as default in the templates repo then, to make sure that only released versions are merged into **main**.

Yuri05 (approved):

src/PKSim.Core/CoreConstants.cs
msevestre: Yep. Can you create an issue for this in the repo?

Yuri05: done

PR #2045, opened 2022-01-12
2041 remove example project

msevestre:


msevestre:
@Yuri05 Merging to see if setup creation is successful

PR #2047, opened 2022-01-13
Multiple PI in parallel

### LINKED ISSUE #1901: 1336 running multiple parameter identifications in parallel in pk sim
@abdelr commented on [Mon Oct 18 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/pull/1345)

Fixes #1336 (Support for multiple parameter identifications in parallel)

abdelr:
Fixes #1901 (Multiple PI in parallel)

### DIFF
--- a/src/PKSim.CLI.Core/PKSim.CLI.Core.csproj
+++ b/src/PKSim.CLI.Core/PKSim.CLI.Core.csproj
@@ -28,7 +28,7 @@
 
   <ItemGroup>
     <PackageReference Include="OSPSuite.Utility" Version="4.0.0.4" />
-    <PackageReference Include="OSPSuite.Core" Version="11.0.87" />
+    <PackageReference Include="OSPSuite.Core" Version="11.0.88" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
+++ b/src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
@@ -23,6 +23,7 @@ using OSPSuite.Presentation.Views;
 using OSPSuite.Assets;
 using OSPSuite.Presentation.Presenters.Events;
 using PKSim.Core;
+using System.Collections.Generic;
 
 namespace PKSim.Presentation.Presenters.Main
 {
@@ -64,6 +65,9 @@ namespace PKSim.Presentation.Presenters.Main
       private readonly IActiveSubjectRetriever _activeSubjectRetriever;
       private bool _enabled;
       private SimulationState _simulationState;
+      private List<string> _activePIs;
+      private string _currentlyActivePI;
+      private readonly IEventPublisher _eventPublisher;
 
       //cache containing the name of the ribbon category corresponding to a given type.Returns an empty string if not found
       private readonly ICache<Type, string> _dynamicRibbonPageCache = new Cache<Type, string>(t => string.Empty);
@@ -72,7 +76,7 @@ namespace PKSim.Presentation.Presenters.Main
 
       public MenuAndToolBarPresenter(IMenuAndToolBarView view, IMenuBarItemRepository menuBarItemRepository,
          IButtonGroupRepository buttonGroupRepository, IMRUProvider mruProvider,
-         ISkinManager skinManager, IStartOptions startOptions, ICoreWorkspace workspace, IActiveSubjectRetriever activeSubjectRetriever) : base(view, menuBarItemRepository, mruProvider)
+         ISkinManager skinManager, IStartOptions startOptions, ICoreWorkspace workspace, IActiveSubjectRetriever activeSubjectRetriever, IEventPublisher eventPublisher) : base(view, menuBarItemRepository, mruProvider)
       {
          _menuBarItemRepository = menuBarItemRepository;
          _buttonGroupRepository = buttonGroupRepository;
@@ -81,6 +85,8 @@ namespace PKSim.Presentation.Presenters.Main
          _workspace = workspace;
          _activeSubjectRetriever = activeSubjectRetriever;
          _enabled = true;
+         _activePIs = new List<string>();
+         _eventPublisher = eventPublisher;
       }
 
       protected override void AddRibbonPages()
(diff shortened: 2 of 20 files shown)


src/PKSim.Core/Model/PKSimParameter.cs
```diff
@@ -29,7 +29,8 @@ public override double Value
          }
       }
 
-      public override bool IsChangedByCreateIndividual
+      //ToDo: @michael -> No longer virtual so it cannot be overwriten
```
msevestre, line 32: Please revert this. I have merged CKD into develop so you are good to go

msevestre (changes requested):
Remove hack. I don't think we should do stuff like this for the sake of compiling because the code is clearly going to break

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -510,18 +516,24 @@ public void Visit(SensitivityAnalysis sensitivityAnalysis)
 
       public void Visit(ParameterIdentification parameterIdentification)
       {
+         _currentlyActivePI = parameterIdentification.Id;
+         _parameterIdentificationRunning = _activePIs.Contains(parameterIdentification.Id);
```
msevestre, line 520: Sorry i don't understand any of this code.
Why do you need these flags? 
`_parameterIdentificationRunning ` seems to only be set and never used?

msevestre: ok I see it's being used below. But you can start multiple PI at the same time. Why do you need to know that one is running?

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -510,18 +516,24 @@ public void Visit(SensitivityAnalysis sensitivityAnalysis)
...
          _parameterIdentificationRunning = true;
          updateParameterIdentificationItems(parameterIdentificationEvent.ParameterIdentification);
       }
 
       public void Handle(ParameterIdentificationTerminatedEvent parameterIdentificationEvent)
       {
-         _parameterIdentificationRunning = false;
+         _activePIs.Remove(parameterIdentificationEvent.ParameterIdentification.Id);
+         if (parameterIdentificationEvent.ParameterIdentification.Id == _currentlyActivePI)
+            _parameterIdentificationRunning = false;
```
msevestre, line 536: this is wrong here. One might still be running you have multiple ones. 
What are you trying to do ? Is this to only have a stop button? This is also weird that the menu and tool bar presenter needs to track active PIs were we already have a manager tracking active PIs...

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -510,18 +516,24 @@ public void Visit(SensitivityAnalysis sensitivityAnalysis)
...
+         _parameterIdentificationRunning = _activePIs.Contains(parameterIdentification.Id);
          updateParameterIdentificationItems(parameterIdentification);
+         _eventPublisher.PublishEvent(new ParameterIdentificationSelectedEvent(parameterIdentification));
       }
 
       public void Handle(ParameterIdentificationStartedEvent parameterIdentificationEvent)
       {
+         _activePIs.Add(parameterIdentificationEvent.ParameterIdentification.Id);
          _parameterIdentificationRunning = true;
          updateParameterIdentificationItems(parameterIdentificationEvent.ParameterIdentification);
```
msevestre, line 529: this flag is also `_parameterIdentificationRunning `  suprirsing. I think this could be a parameter of the updateParameterIdentificationItems no?
You do not need to tract activePI IMO here as well. Let's discuss

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
abdelr: `_parameterIdentificationRunning` is used to set the menu view accordingly. You can check the `updateParameterIdentificationItems` method where the buttons are enabled and disabled based on its value.

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
abdelr: Now we have multiple PIs. So we have the list and the active one. When the event `ParameterIdentificationTerminatedEvent` is handled, then its id is removed from the list and in case it was the current PI, the running flag is set to false (so we can show the run button enabled, and disabled the stop for example). Sorry if it is not clear. I am open to suggestions on how to improve it.

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
abdelr: Sure, we can discuss. Not sure if you are working now. Otherwise we talk tomorrow.

abdelr:
Only one addition. No need to raise the event anymore since the active parameter identification will be fetched in core from the ActiveSubjectRetriever. @msevestre Please, wait to merge since we need to update Core dlls first

src/PKSim.CLI.Core/PKSim.CLI.Core.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
Yuri05, line 1: Lol, where is the difference between those 2 lines?

abdelr: Can't see any but after I get some conflicts with different core component versions I always need to edit the project files manually so maybe github does not like this... ;)

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -543,21 +551,23 @@ private void updateParameterIdentificationItems(ParameterIdentification paramete
...
       {
-         _sensitivityRunning = true;
+         _runningSensitivityAnalysis = sensitivityAnalysisEvent.SensitivityAnalysis;
          updateSensitivityAnalysisItems(sensitivityAnalysisEvent.SensitivityAnalysis);
       }
 
       public void Handle(SensitivityAnalysisTerminatedEvent sensitivityAnalysisEvent)
       {
-         _sensitivityRunning = false;
+         _runningSensitivityAnalysis = null;
```
Yuri05, line 560: Sensitivity analyses cannot be run in parallel, only PIs, correct?

abdelr: There is no change on the sensitivityAnalysis logic so far

Yuri05 (approved):

abdelr:
@msevestre Core has been plugged in and seams to work fine.

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -23,6 +23,8 @@
 using OSPSuite.Assets;
 using OSPSuite.Presentation.Presenters.Events;
 using PKSim.Core;
+using System.Collections.Generic;
+using NPOI.OpenXmlFormats.Dml;
```
msevestre, line 27: Let's remove this

tests/PKSim.Tests/Presentation/MenuAndToolBarPresenterSpecs.cs
```diff
@@ -10,6 +10,7 @@
 using OSPSuite.Presentation.Services;
 using OSPSuite.Presentation.Views;
 using PKSim.Core;
+using OSPSuite.Utility.Events;
```
msevestre, line 13: this should go as well

msevestre (commented):
looks good. Just some very minor changes. Can you do that and I'll merge?

PR #2049, opened 2022-01-13
Fixes #1933 Add urine compartment

### LINKED ISSUE #1933: Add "Urine compartment emptying" as an event template to PK-Sim
Similar as the gallblader emptying event, an event is required to empty the urine compartment.
The event should be available as a selection in the "Create Event" menu. No parameters needed to be specified by the user. The user should simply remove all compounds from the urine compartment (set all amounts to zero).
The event should then be added to a simulation via the "Events" tab, potentially at multiple times for one simulation.

@Yuri05

msevestre:


src/Db/Dump/PKSimDB_dump.txt
```diff
@@ -74327,6 +74334,7 @@ INSERT INTO tab_parameters VALUES('Sum of active process rates lumen to mucosa',
 INSERT INTO tab_parameters VALUES('Target eGFR','Target eGFR','Target eGFR value','Flow per body surface area',NULL);
 INSERT INTO tab_parameters VALUES('eGFR','eGFR','Estimated GFR is the GFR normalized to body surface area (BSA). The eGFR is used clinically to determine staging of chronic kidney disease.','Flow per body surface area',NULL);
 INSERT INTO tab_parameters VALUES('Disease factor','Disease factor','Used for scaling reference concentration based on the selected disease state.','Dimensionless',NULL);
+INSERT INTO tab_parameters VALUES('Urine compartment emptying enabled','Urine compartment emptying enabled','TODO: Enable urine compartment emptying','Dimensionless',NULL);
```
msevestre, line 74337: @Yuri05 Description required here

src/Db/Dump/PKSimDB_dump.txt
```diff
@@ -54944,6 +54946,8 @@ INSERT INTO tab_container_parameter_rates VALUES(5954,'DISEASE_STATE','CKD','Tar
 INSERT INTO tab_container_parameter_rates VALUES(157,'ORGAN','Kidney','eGFR','GFR_Calculation','PARAM_eGFR',0.0,1,NULL,1,NULL);
 INSERT INTO tab_container_parameter_rates VALUES(5015,'GENERAL','PROTEIN','Disease factor','DiseaseStates','One_Rate',0.0,1,NULL,1,NULL);
 INSERT INTO tab_container_parameter_rates VALUES(5941,'GENERAL','TRANSPORTER','Disease factor','DiseaseStates','One_Rate',0.0,1,NULL,1,NULL);
+INSERT INTO tab_container_parameter_rates VALUES(5955,'EVENTGROUP','Urine compartment emptying','Urine compartment emptying enabled','Events','One_Rate',0.0,1,1.0,1,NULL);
```
Yuri05, line 54949: Why do we need this parameter `Urine compartment emptying enabled`?
Because the users want to be able to activate/deactivate the event in a simulation without reconfiguring?

msevestre: since urine is a compartment already. Why do we have the name compartment?

msevestre: Similar to Gallbladder emptying enabled we do not say Gallbladder compartment xx
Going to rename for consistency sake

Yuri05: Actually I think the event and the parameter should be called "Bladder emptying" / "Bladder emptying enabled" 
(or "Urinary bladder emptying" / "Urinary bladder emptying enabled").
Maybe somebody with medical background (@sfrechen ?) can comment on this?

msevestre: Makes sense to me..
Urinary bladder sounds more physiological than urine compartment

msevestre:
Ok let's agree on nomenclature before I continue
Suggetion of @Yuri05 makes sense to me

msevestre:
going to go with Urinary bladder and rename accordingly
Then we can always rename later. I need to merge the DB to keep workoing

Yuri05 (approved):

PR #2051, opened 2022-01-14
Suggestion for cleaner code

msevestre:


msevestre:
Here is my suggestion. I think this can be even simplified further if we can access the object holding a refernece on all running PIS ( i think we have this no?)

Also because this will ahve to be implemented in MoBi, this can probably be promoted to core.. but maybe not

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -65,14 +66,12 @@ public class MenuAndToolBarPresenter : AbstractMenuAndToolBarPresenter, IMenuAnd
       private readonly IActiveSubjectRetriever _activeSubjectRetriever;
       private bool _enabled;
       private SimulationState _simulationState;
-      private List<string> _activePIs;
-      private string _currentlyActivePI;
+      private readonly List<ParameterIdentification> _runningParameterIdentifications = new List<ParameterIdentification>();
```
msevestre, line 69: Let's not use id. LEt's use object and reference. Id is only for serialization 
ALso run resharper to see that the field could be readonly

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -65,14 +66,12 @@ public class MenuAndToolBarPresenter : AbstractMenuAndToolBarPresenter, IMenuAnd
...
       private readonly IEventPublisher _eventPublisher;
 
       //cache containing the name of the ribbon category corresponding to a given type.Returns an empty string if not found
       private readonly ICache<Type, string> _dynamicRibbonPageCache = new Cache<Type, string>(t => string.Empty);
-      private bool _parameterIdentificationRunning;
-      private bool _sensitivityRunning;
```
msevestre, line 75: I also rewrite the sensitivity stuff to be similar

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -516,32 +514,33 @@ public void Visit(SensitivityAnalysis sensitivityAnalysis)
...
       public void Handle(ParameterIdentificationTerminatedEvent parameterIdentificationEvent)
       {
-         _activePIs.Remove(parameterIdentificationEvent.ParameterIdentification.Id);
-         if (parameterIdentificationEvent.ParameterIdentification.Id == _currentlyActivePI)
-            _parameterIdentificationRunning = false;
-         updateParameterIdentificationItems(parameterIdentificationEvent.ParameterIdentification);
+         var parameterIdentification = parameterIdentificationEvent.ParameterIdentification;
+         _runningParameterIdentifications.Remove(parameterIdentification);
+         //only update if selected subject in the actual parameter identification terminated
+         var activeSubject = _activeSubjectRetriever.Active<ParameterIdentification>();
```
msevestre, line 533: no need to save the selected one as member. We can always get the selected subject

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -516,32 +514,33 @@ public void Visit(SensitivityAnalysis sensitivityAnalysis)
...
+         if(Equals(activeSubject, parameterIdentification))
```
msevestre, line 534: you were always updating the view. I think this was wrong, It should only be done when the active subject is the one also terminated no?

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -516,32 +514,33 @@ public void Visit(SensitivityAnalysis sensitivityAnalysis)
...
+            updateParameterIdentificationItems(parameterIdentification);
       }
 
       private void updateParameterIdentificationItems(ParameterIdentification parameterIdentification)
       {
-         var hasResult = !_parameterIdentificationRunning && parameterIdentification.HasResults;
-         _menuBarItemRepository[MenuBarItemIds.RunParameterIdentification].Enabled = !_parameterIdentificationRunning;
-         _menuBarItemRepository[MenuBarItemIds.StopParameterIdentification].Enabled = _parameterIdentificationRunning;
+         var parameterIdentificationRunning = _runningParameterIdentifications.Contains(parameterIdentification);
```
msevestre, line 540: this can be infer from the list I believe

msevestre (commented):
I think this is cleaner and easier to understand because we rely on object and do not use flags to model state that can be infer from properties

abdelr (approved):

PR #2056, opened 2022-01-17
Fixes #2052 renaming uses the wrong type

### LINKED ISSUE #2052: REnaming simulation in v11 asks for new Name for project
I think it's only in v11
image

msevestre:


src/PKSim.Presentation/Services/EntityTask.cs
```diff
@@ -34,7 +37,7 @@ public ICommand Rename(IEntity elementToRename)
       private ICommand rename(IEntity elementToRename, bool isStructuralChange)
       {
          var dto = _renameObjectDTOFactory.CreateFor(elementToRename);
-         var newName = NewNameFor(elementToRename, dto.UsedNames, dto.ContainerType);
+         var newName = NewNameFor(elementToRename, dto.UsedNames, TypeFor(elementToRename));
```
msevestre, line 40: Just using the wrong tupe after refactoring. ContainerType is always project for bb

Yuri05 (approved):

PR #2057, opened 2022-01-17
Get a graph again

msevestre:
@Yuri05 I tried at least 20 different enzymes between v10 and v11 and never get a difference in the graph
So until you find one that is different, I am considering it a **VERY GOOD AND SECURE** fix :)


msevestre:
This is the real fix. Changing the argument to QualitativeSummaryOptions instead of NumericSummaryOptions

Yuri05:
better :)

msevestre:
agreed

PR #2058, opened 2022-01-17
2039 red blood cells

msevestre:


src/PKSim.UI/Views/ProteinExpression/ExpressionDataView.cs
```diff
@@ -698,23 +698,23 @@ private void configChart(string titleText)
...
-         // Register the summary function in chart.
-         const string STR_CUSTOM_SUMMARY_FUNCTION_NAME = "WeightedAverage";
-         const string STR_CUSTOM_SUMMARY_FUNCTION_DISPLAY_NAME = "WeightedAverage";
-         chart.RegisterSummaryFunction(STR_CUSTOM_SUMMARY_FUNCTION_NAME, STR_CUSTOM_SUMMARY_FUNCTION_DISPLAY_NAME, 1,
-                                       new[] {argument1Description, argument2Description},
-                                       customSummary);
-
-         chart.SeriesTemplate.NumericSummaryOptions.SummaryFunction = 
-            $"{STR_CUSTOM_SUMMARY_FUNCTION_NAME}([{_fieldNormValue.FieldName + "_" + _fieldNormValue.SummaryType}],[{_fieldNormValue2.FieldName + "_" + _fieldNormValue2.SummaryType}])";
+         // // Create argument descriptions for the summary function.
```
msevestre, line 701: This is just required for now so that I can get some values. We can uncomment when the problem is solved with query

msevestre (commented):
@Yuri05 I think this fixes the issue. No changes of default value. will do it next

src/PKSim.Core/Commands/EditIndividualMoleculeExpressionInSimulationSubjectFromQueryCommand.cs
```diff
@@ -58,5 +63,17 @@ public override void Execute(IExecutionContext context)
          _queryExpressionResults = null;
          _simulationSubject = null;
       }
+
+      private void updateLocalizationForProtein(IndividualProtein protein, IExecutionContext context)
+      {
+         //special case for blood cells and vascular endothelium. if the value is set > 0. we need to turn on some locations programatically
+         var bloodCells = _queryExpressionResults.ExpressionResultFor(CoreConstants.Compartment.BLOOD_CELLS);
+         if (bloodCells?.RelativeExpression > 0 && !protein.IsBloodCell())
```
Yuri05, line 71: why do we need the part `&& !protein.IsBloodCell()` ?  
(same question for VascEndo below)

msevestre: The logic that I have in mind is the following: If Bloodcells was already activated (Membrane OR Intracellular is on), then the field was set by the user and the rel exp is visible. Therefore, I do not want to trigger a command that willl reset potentially some fractions etc... 

In short, I just want to ensure that the rel exp field is made visible only if it was not already visible.

Yuri05: The logic seems ok :)
But I still don'T understand the highlighted part. It checks if the protein is named "Blood Cells". Why?
I guess you mean ` !protein.InBloodCell()` ??

And also for VascEndo `!proteinInVascularEndothelium` and not `!proteinIsVascularEndothelium`?

msevestre: you are very much correct. I changed the code locally without saving and it did not commit those changes. I am going to add test for this use case where it's already set :)

msevestre: Very good eye @Yuri05 Thanks for catching this. 
I added a new test testing the second part of the if for VascularEndo and RBC . We are good

msevestre (commented):
ready to merge I think

Yuri05 (approved):

Yuri05:
looks good :)

PR #2064, opened 2022-01-20
Control size issue

msevestre:


msevestre (commented):
@Yuri05 I am experimenting wiht some new layout options here. I'd like to merge it and you tell me how it looks on your screen
also I have added a small x button on the Spalshscreen to dismiss it if youi don't want to see it. We can discuss about it. I just wanted to try the TableLayout stuff

msevestre:
Merging this for now. We can revert some of those later as I am experimenting some more

PR #2082, opened 2022-02-01
Fixes #2081 PKSim CLI filter simulation

### LINKED ISSUE #2081: PKSIm.CLI => Do not generate outputs ofr simulation not used (support argument)

msevestre:


src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -906,6 +906,9 @@ public static string MapToSnapshotNotSupportedWithoutContext(string snapshotType
 
          public static string CannotFindSimulationInSnapshot(string simulationName, string project) => CannotFindBuildingBlockInSnapshot(ObjectTypes.Simulation, simulationName, project);
 
+         public static string SimulationUsedInPlotsAreNotExported(IReadOnlyList<string> simulationNames, string project)
```
msevestre, line 909: Making sure that simulation used in plots are also referenced here. This is just a simple validation step.

src/PKSim.CLI.Core/RunOptions/QualificationRunOptions.cs
```diff
@@ -16,5 +16,10 @@ public class QualificationRunOptions
       ///    Should simulation be performed as part of the run?
       /// </summary>
       public bool Run { get; set; }
+
+      /// <summary>
+      /// Specifies if project files (snapshot and project should be exported)
+      /// </summary>
+      public bool ExportProjectFiles { get; set; }
```
msevestre, line 23: Should we export the project files or not. Default will be not because I don't think we need to export those for the qualification

msevestre (commented):
READY FOR REVIEW

Yuri05 (approved):

src/PKSim.Infrastructure/Services/ImportObservedDataTask.cs
```diff
@@ -306,13 +306,13 @@ public IEnumerable<string> PredefinedValuesFor(string name)
          return Enumerable.Empty<string>();
       }
 
-      public IReadOnlyList<string> DefaultMetaDataCategories => CoreConstants.ObservedData.DefaultProperties;
+      public IReadOnlyList<string> DefaultMetaDataCategories { get; } = CoreConstants.ObservedData.DefaultProperties;
 
-      public IReadOnlyList<string> ReadOnlyMetaDataCategories => new List<string> { };
+      public IReadOnlyList<string> ReadOnlyMetaDataCategories { get; } = new List<string> { };
+    
+      public bool MolWeightAlwaysEditable { get; } = false;
```
msevestre, line 313: @georgeDaskalakis I changed this stuff here. Also when you always return a constant, use {get;} instead 0f => false. A bit faster...

src/PKSim.CLI.Core/RunOptions/QualificationRunOptions.cs
Yuri05: ok, let's keep it like this. 
If this option can be passed to the QualificationRunner later I'm perfectly fine with that :)

PR #2092, opened 2022-02-09
2055 old values not scaled

msevestre:
I feel that I am patching a leak with some tap but it's better than nothing..

### DIFF
--- a/src/PKSim.Presentation/Presenters/ProteinExpression/ProteinExpressionsPresenter.cs
+++ b/src/PKSim.Presentation/Presenters/ProteinExpression/ProteinExpressionsPresenter.cs
@@ -19,14 +19,16 @@ using OSPSuite.Presentation.Presenters;
 
 namespace PKSim.Presentation.Presenters.ProteinExpression
 {
-   public enum ColumnNamesOfTransferTable
+   public static class ColumnNamesOfTransferTable
    {
-      Container,
-      DisplayName,
-      RelativeExpressionOld,
-      ExpressionValue,
-      RelativeExpressionNew,
-      Unit
+      public static string Container = "Container";
+      public static string DisplayName = "DisplayName";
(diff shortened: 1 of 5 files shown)


src/PKSim.Presentation/Presenters/ProteinExpression/ProteinExpressionsPresenter.cs
```diff
@@ -19,14 +19,16 @@
...
-   public enum ColumnNamesOfTransferTable
+   public static class ColumnNamesOfTransferTable
    {
-      Container,
-      DisplayName,
-      RelativeExpressionOld,
-      ExpressionValue,
-      RelativeExpressionNew,
-      Unit
+      public static string Container = "Container";
```
msevestre, line 24: I created those so that we could remove the .ToString() used EVERYWHERE!!

src/PKSim.UI/Views/ProteinExpression/TransferView.cs
```diff
@@ -133,58 +140,87 @@ public void SetData(DataTable transferTable, string selectedUnit)
...
+            Maximum = max,
+            Minimum = 0,
+            NullText = string.Empty,
+            PercentView = false,
+            ShowTitle = true,
+            DisplayFormat = {FormatType = FormatType.Numeric}
+         };
+      }
+
+      private int getPercentageMax(DataView dataView, string columnName)
```
msevestre, line 185: This is the change. Instead of setting the max to 100, I set the max to the max value in the column (int) and then this will be used as max for the repo. The scaling is then done automatically

msevestre:
@Yuri05 This cannot be tested, unfortunately. Everything happens in the view.

msevestre:
and this?

Yuri05 (approved):

PR #2093, opened 2022-02-09
Fixes #1755 aucr cmax r not calculated properly

### LINKED ISSUE #1755: AUCR and CmaxR not calculated if formulation with more than 1 Particle bin used
In the attached file I wanted to calculete AUCR and CMaxR in the simulation" DDI Treatment - Itraconazole/Midazolam - Prueksaritanont 2017_particle50mg". But all I see after clicking on the triangle button, that it is calculating for some time. then it stops and nothing chnged. No AUCR or Cmax R is visible. unlike for the other simulation using Solution or default particle 10 µM. 
Is that normal or a bug?
[Test-AUCR-calc.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/6466058/Test-AUCR-calc.zip)

image

msevestre:


msevestre:
@Yuri05 can you revuiew this?

src/PKSim.Core/Services/GlobalPKAnalysisTask.cs
```diff
@@ -319,17 +319,23 @@ private bool isMultipleOral(Simulation simulation, Compound compound)
...
+         if (schemaItem == null)
+            return false;
+
          return schemaItem.ApplicationType == ApplicationTypes.Intravenous || schemaItem.ApplicationType == ApplicationTypes.IntravenousBolus;
       }
 
       private ApplicationType applicationTypeFor(Simulation simulation, Compound compound)
       {
-         var numberOfApplications = schemaItemsFrom(simulation, compound).Count;
+         var numberOfApplications = simulation.AllApplicationParametersOrderedByStartTimeFor(compound.Name).Count;
```
Yuri05, line 338: This is the only relevant difference, correct?
For a metabolite (which is not administered) number of applications was zero before; now it is number of applications of its parent drug, right?

By the way, I would rename the function `AllApplicationParametersOrderedByStartTimeFor `to something like `AllOrderedApplicationStartTimes`. 
Because in `AllApplicationParameters `I would expect also all other application parameters like e.g. DrugMass, InfusionTime etc. (I had to look into the implementation of the function to understand what it does)

msevestre: ok but this is in core so not directly relevant to this problem. Please create an issue for this

>For a metabolite (which is not administered) number of applications was zero before; now it is number of applications of its parent drug, right?

yes

Yuri05: Issue created: https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1494
PR looks good for me

Yuri05 (approved):

PR #2097, opened 2022-02-17
Perf oroptimization: Create / Edit Simulation

msevestre:


src/PKSim.Presentation/Presenters/ConfigurableLayoutPresenter.cs
```diff
@@ -40,6 +40,7 @@ public void AddViews(IEnumerable<IView> views)
          if (allViews.Count > 1)
          {
             var configurableView = _configurableContainerLayoutViewFactory.Create();
+            configurableView.StartAddingViews();
```
msevestre, line 43: @Yuri05 I have added a step here but this is in fact not required as the views are only redrawn when visible. At any rate, it does not hurt and is symmetrical

src/PKSim.Presentation/Views/IResizableWithDefaultHeightView.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace PKSim.Presentation.Views
+{
+   public interface IResizableWithDefaultHeightView : IResizableView
+   {
+      /// <summary>
+      /// Default height placeholder that will be used to initialize the container of this view
+      /// This is required to avoid a flickering effect. 
+      /// </summary>
+      int DefaultHeight { get; }
```
msevestre, line 11: This is new. I am giving a default height so that we do not have the flickering effect on BUILDING The view

src/PKSim.UI/UIConstants.cs
```diff
@@ -39,6 +39,10 @@ public static class Size
          public static readonly int COMPOUND_VIEW_WIDTH = ScaleForScreenDPI(760);
 
          public static readonly int MODEL_PICTURE_HEIGHT = ScaleForScreenDPI(400);
+
+         public static readonly int SIMULATION_COMPOUND_CONFIGURATION_DEFAULT_HEIGHT = ScaleForScreenDPI(280);
```
msevestre, line 43: Just looks like it's working on my PC. Can you try on yours>

src/PKSim.UI/Views/Core/AccordionLayoutView.cs
```diff
@@ -11,14 +11,25 @@ public AccordionLayoutView()
...
+      }
+
       public void AddView(IView view)
       {
          var group = layoutControl.Root.AddGroup();
-         AddViewToGroup(group, view);
+         var layoutItem = AddViewToGroup(group, view);
+         var resizeView = view as IResizableWithDefaultHeightView;
+         if (resizeView != null)
+            this.AdjustLayoutItemSize(layoutItem, resizeView, resizeView.DefaultHeight);
```
msevestre, line 25: this is where it's being used

msevestre (commented):
@Yuri05 Building of UI is faster on my PC. But who knows

src/PKSim.UI/UIConstants.cs
Yuri05: will try it with the next nightly

Yuri05 (approved):

PR #2115, opened 2022-03-10
2112 fraction expressed should be formula

msevestre:
I don't think this will solve our issues quite yet as we have some tricky stuff done for other parameters such as InitialConcentration etc.. but this might work with extend in MoBi. tocheck
/cc @PavelBal @Yuri05


msevestre (commented):
This are the criteria that I have defined for now

| ContainerId | ContainerName | ContainerType | Tag | Condition | ParameterName |
| :--- | :--- | :--- | :--- | :--- | :--- |
| 5015 | PROTEIN | GENERAL | Interstitial | Has | Fraction expressed interstitial |
| 5015 | PROTEIN | GENERAL | Intracellular | Has | Fraction expressed intracellular |
| 5941 | TRANSPORTER | GENERAL | Brain | InContainer | Fraction expressed at blood brain barrier |
| 5941 | TRANSPORTER | GENERAL | Plasma | Has | Fraction expressed at blood brain barrier |
| 5941 | TRANSPORTER | GENERAL | Brain | InContainer | Fraction expressed brain tissue |
| 5941 | TRANSPORTER | GENERAL | Interstitial | Has | Fraction expressed brain tissue |
| 5015 | PROTEIN | GENERAL | Brain | NotInContainer | Fraction expressed interstitial |

src/PKSim.Core/Services/IndividualMoleculeFactory.cs
```diff
@@ -71,69 +78,48 @@ protected bool HasAgeParameter(ISimulationSubject simulationSubject)
...
-            Dimension = MOLAR_CONCENTRATION,
-            GroupName = CoreConstants.Groups.RELATIVE_EXPRESSION,
-            MinValue = 0,
-            MinIsAllowed = true
-         };
+         //In the db, we define those parameters as rate for now but we need to create a value meta data here so we update
+         var parameterMetaData = _parameterRateRepository.ParameterMetaDataFor(_containerPath, paramName);
+         var parameterValueMetaData = new ParameterValueMetaData {DefaultValue = defaultValue};
+         parameterValueMetaData.UpdatePropertiesFrom(parameterMetaData);
+         return parameterValueMetaData;
```
Yuri05, line 87: I don't quite get this. Why do we need **Value** Metadata and not **Rate** metadata here?

msevestre: this is a very good question. I am just trying for now to be compatible with the old way,
In short the default value is always 0 except for our undefined enzymes where it's 1
At the moment, the value in the DB is NaN, I could set it to 0. I guess and go from there. 
Is RelativeExpression only define in one compartment ? Or does it depend on the organ?

PavelBal: > Is RelativeExpression only define in one compartment ? Or does it depend on the organ?

We should only have **one** `RelativeExpression` per Organ. As we cannot have molecule properties on the organ level, I would suggest to put it into intracellular space.

msevestre: How is it NOW? Is it always in the Intracellular? Or do we have it sometimes in Lumen etc..
If we do, we have a problem

PavelBal: You are right... It is always in `Intracellular` but also in lumen segments. Which is a questionable concept anyways...

msevestre: Agreed. This is because we cannot create molecule parameter at the organ level. It comes back to the same issue. 
In the meantime, since we do not want to change this now, what does this mean for what we are trying to achieve?
I cannot create a condition for RElativeExpression that would work as we except (e.g. Intracellular OR Lumen) since we do not have OR. If I leave the RelativeExpression default value to NaN, then we still have the problem that we cannot extend. It's easier to do probably than adding formulas left and right but this is all not ideal

msevestre:
@PavelBal 
ok this is is done for our 3 parameters
What about fraction expressed basolateral and fraction expressed apical. Same problem isn't it?
However here, we have different formula based on location. How do we deal with this
[new_way.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/8236406/new_way.zip)

See mobi file here attached. Please used latest nightly in MoBi
https://ci.appveyor.com/project/open-systems-pharmacology-ci/mobi/builds/42873379/artifacts

We can update in PK-Sim. For now, fraction expressed basolateral and apical are still nan.

src/PKSim.Core/Services/IndividualEnzymeFactory.cs
```diff
@@ -37,11 +40,17 @@ public IndividualEnzyme AddUndefinedLiverTo(Individual individual)
...
-            AddContainerExpression(zone.Container(CoreConstants.Compartment.INTRACELLULAR), undefinedLiver.Name,
-               RelExpParam(REL_EXP, defaultValue: 1),
+            var intracellular = zone.Container(CoreConstants.Compartment.INTRACELLULAR);
+            AddContainerExpression(intracellular, undefinedLiver.Name,
+               RelExpParam(REL_EXP),
                FractionParam(FRACTION_EXPRESSED_INTRACELLULAR, CoreConstants.Rate.ONE_RATE),
                InitialConcentrationParam(CoreConstants.Rate.INITIAL_CONCENTRATION_INTRACELLULAR)
             );
+
+            var relExpParameter = intracellular.EntityAt<IParameter>(undefinedLiver.Name, REL_EXP);
```
msevestre, line 50: @Yuri05 Following up on your comments, I am updating the value of the parameter after the fact .No more messing around

src/PKSim.Infrastructure/ORM/Repositories/ParameterMetaDataRepository.cs
```diff
@@ -49,15 +59,35 @@ protected override void DoStart()
...
+         var allConditions = _flatContainerParameterDescriptorConditionRepository.All()
+            .Where(x => x.ContainerId == containerId && x.ParameterName == parameterName)
+            .ToList();
+
+         if (!allConditions.Any())
+            return null;
 
+         var operators = allConditions.Select(x => x.Operator).Distinct().ToList();
+         if (operators.Count != 1)
+            throw new ArgumentException(PKSimConstants.Error.MultipleOperatorFoundForContainer(containerId, parameterName));
```
msevestre, line 85: @Yuri05 Here checking only one operator (Or or And)

msevestre:
Here are the criteria I have defined so far
| ContainerId | ContainerName | ContainerType | ParameterName | Tag | Condition | Operator |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 5015 | PROTEIN | GENERAL | Fraction expressed interstitial | Interstitial | Has | And |
| 5015 | PROTEIN | GENERAL | Fraction expressed intracellular | Intracellular | Has | And |
| 5941 | TRANSPORTER | GENERAL | Fraction expressed at blood brain barrier | Brain | InContainer | And |
| 5941 | TRANSPORTER | GENERAL | Fraction expressed at blood brain barrier | Plasma | Has | And |
| 5941 | TRANSPORTER | GENERAL | Fraction expressed brain tissue | Brain | InContainer | And |
| 5941 | TRANSPORTER | GENERAL | Fraction expressed brain tissue | Interstitial | Has | And |
| 5015 | PROTEIN | GENERAL | Relative expression | Intracellular | Has | Or |
| 5015 | PROTEIN | GENERAL | Relative expression | LumenSegment | Has | Or |
| 5941 | TRANSPORTER | GENERAL | Relative expression | Intracellular | Has | Or |
| 5941 | TRANSPORTER | GENERAL | Relative expression | LumenSegment | Has | Or |

msevestre (commented):
I think this looks pretty good. All tests passing in PKSim and MoBi

src/Db/Dump/PKSimDB_dump.txt
```diff
@@ -93607,4 +93618,13 @@ SELECT TAB_CONTAINER_PARAMETERS.CONTAINER_ID                    AS ContainerId,
...
 
 FROM TAB_CONTAINER_PARAMETERS;
+CREATE VIEW VIEW_CONTAINER_PARAMETER_DESCRIPTOR_CONDITIONS AS SELECT
+     tab_container_parameter_descriptor_conditions.container_id AS ContainerId,
+     tab_container_parameter_descriptor_conditions.container_name AS ContainerName,
+     tab_container_parameter_descriptor_conditions.container_type as ContainerType,
+     tab_container_parameter_descriptor_conditions.parameter_name as ParameterName,
+     tab_container_parameter_descriptor_conditions.tag AS Tag,
+     tab_container_parameter_descriptor_conditions.condition AS Condition,
+     tab_container_parameter_descriptor_conditions.operator AS Operator
```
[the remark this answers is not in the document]
Yuri05: I think it's ok. Would not make the DB structure more complicated for that.

Yuri05 (approved):

src/PKSim.Core/Services/IndividualTransporterFactory.cs
```diff
@@ -141,22 +149,41 @@ private void addOrganWithLumenParameters(IContainer organ, IndividualTransporter
...
+         addContainerExpression(organ.Container(INTERSTITIAL), transporter,
+            TransportDirectionId.None,
+            FractionParam(FRACTION_EXPRESSED_BASOLATERAL, CoreConstants.Rate.PARAM_F_EXP_BASOLATERAL, visible: false),
+            InitialConcentrationParam(CoreConstants.Rate.INITIAL_CONCENTRATION_INTERSTITIAL_TRANSPORTER)
+         );
+
+         addContainerExpression(organ.Container(INTRACELLULAR), transporter,
+            DefaultTissueDirectionFor(transporter.TransportType),
+            RelExpParam(REL_EXP),
+            FractionParam(FRACTION_EXPRESSED_APICAL, CoreConstants.Rate.ZERO_RATE)
```
msevestre: @Yuri05 Apical parameter is defined everywhere (before we only add basolateral hidden with value 1)
Now basolateral formula is the same everwhere (1-f_abpical). Default value for apical is 0 and therefor all the same as bfore

msevestre (commented):
I think we are good here

msevestre:
@Yuri05 I am merging this. All tests are running. Let see if something broke in the nightly :)

Yuri05:
ok...

src/Db/Dump/PKSimDB_dump.txt
```diff
@@ -4555,7 +4555,7 @@ INSERT INTO tab_calculation_method_rates VALUES('ExpressionParameters','NaN_Rate
 INSERT INTO tab_calculation_method_rates VALUES('ExpressionParameters','PARAM_f_exp_basolateral','1 - f_exp_org_apical','Fraction');
 INSERT INTO tab_calculation_method_rates VALUES('ExpressionParameters','PARAM_f_exp_interstitial','1 - f_exp_org_cell','Fraction');
 INSERT INTO tab_calculation_method_rates VALUES('ExpressionParameters','InitialConcentrationIntracellular_Transporter','f_cell > 0 ? RC * rel_exp_org * f_exp_org_apical * 1 / f_cell : 0','Concentration (molar)');
-INSERT INTO tab_calculation_method_rates VALUES('ExpressionParameters','InitialConcentrationInterstitial_Transporter','f_int > 0 ? RC * rel_exp_org * f_exp_org_basolatateral * 1 / f_int : 0','Concentration (molar)');
+INSERT INTO tab_calculation_method_rates VALUES('ExpressionParameters','InitialConcentrationInterstitial_Transporter','f_int > 0 ? RC * rel_exp_org * f_exp_org_basolateral * 1 / f_int : 0','Concentration (molar)');
```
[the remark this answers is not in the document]
PavelBal: :D

PR #2137, opened 2022-03-25
Fixes update svg support

msevestre:


msevestre:
@Yuri05 This is what it takes to make SVG support work in PK-Sim. Icons are not blurry anymore on my screen, even at  250% pixel
We are missing 4 icons that will be updated very soon before the release :)

PR #2144, opened 2022-03-28
Fixes #2140 fraction basolateral not set properly

### LINKED ISSUE #2140: A simulation which was running in V10 does not run in V11 anymore
[DDI_Treatment-Rifampicin_Midazolam-Backman1998.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/8357967/DDI_Treatment-Rifampicin_Midazolam-Backman1998.zip)
Loading this snapshot works in V10:
grafik

In V11 the error message "Variables became negative" comes:
grafik

Also interesting: when loading the pkml exported for this simulation by the qualification runner in MoBi and running the simulation, 
the same error message comes - but at *different time point* 
grafik

After reducing the absolute tolerance from 1E-10 to 1E-11: simulation runs in V11.
However the simulation result is also slightly different.

Attached is a MoBi project with 2 simulations:
[DDI_Treatment-Rifampicin_Midazolam-Backman1998.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/8357977/DDI_Treatment-Rifampicin_Midazolam-Backman1998.zip)
grafik
* Simulation **_DDI...** exported from V10 (after importing the snapshot in PK-Sim 10)
* Simulation **DDI...** was created by the Qualification Runner (as part of the CYP3A4-DDI Quali workflow) and in MoBi AbsTol was reduced from 1E-10 to 1E-11 to be able to run the simulation

V10-Simulation results (_DDI...):
grafik

V11-Simulation results (DDI)
grafik

msevestre:


src/PKSim.Core/Services/TransportContainerUpdater.cs
```diff
@@ -73,6 +73,10 @@ private void updateFractionExpressedEpithelial(TransporterExpressionContainer tr
          if (fractionExpressedApical == null)
             return;
 
+         //Parameter is a hidden parameter. This is used for consistency purpose only and should not be updated
+         if (!fractionExpressedApical.Visible)
```
msevestre, line 77: A bit to eager here. The null above was covering exactly this. But now that we have this parameter everywhere, we should only update if we can actually set it

Yuri05: lol
always those nice side effects 😄

Yuri05 (approved):

PR #2147, opened 2022-03-28
Fixes #2034 info missing

### LINKED ISSUE #2034: Protein expression: info missing when transporter direction is not in DB
1) Create new project
2) Add individual
3) Add transporter protein from Dababase: OATP1B1

In PK-Sim 9:

image

In PK-Sim 10, the hint is missing:

image

msevestre:


src/PKSim.Core/Events/EntityEvents.cs
```diff
@@ -155,11 +155,11 @@ public SwapBuildingBlockEvent(IPKSimBuildingBlock oldBuildingBlock, IPKSimBuildi
       }
    }
 
-   public class NoTranporterTemplateAvailableEvent
+   public class NoTransporterTemplateAvailableEvent
    {
       public IndividualTransporter Transporter { get; }
 
-      public NoTranporterTemplateAvailableEvent(IndividualTransporter transporter)
+      public NoTransporterTemplateAvailableEvent(IndividualTransporter transporter)
```
msevestre, line 162: This was ideal before because the presenter was already created and we could react to this event. This is not the case before as we instantiate the presenter AFTER . I kept this event but I am not sure how I can use this in an effective way

src/PKSim.Core/Repositories/ITransporterContainerTemplateRepository.cs
```diff
@@ -36,5 +36,10 @@ public interface ITransporterContainerTemplateRepository : IStartableRepository<
       /// <param name="speciesName">Species where the transporter will be defined</param>
       /// <param name="containerName">Concrete transporter location (e.g Kidney, Liver etc...)</param>
       IEnumerable<TransporterContainerTemplate> TransportersFor(string speciesName, string containerName);
+
+      /// <summary>
+      /// Returns the name of all transporters defined in the PKSim Database (template transporters and their synonyms)
+      /// </summary>
+      IReadOnlyList<string> AllTransporterNames { get; }
```
msevestre, line 43: nothing to do with the issue but we added a bunch of molecules from the db and those were msissing

src/PKSim.Presentation/Presenters/Individuals/IndividualTransporterExpressionsPresenter.cs
```diff
@@ -99,6 +104,22 @@ public void ActivateMolecule(IndividualMolecule molecule)
          rebind();
       }
 
+      private void updateWarning()
+      {
+         if (_transporterExpressionDTO.DefaultAvailableInDatabase)
```
msevestre, line 109: so here is what I am doing. In the DTO I save if this came from DB or not. If yet, then nothing to show
If not, then I check if the transporter type is our default, (Efflux). If yes, I show the warning. Otherwise hide
This means that this warning will always be shown for transporter not from db and Efflux. Are we okay with this?
Otherwise, we need to save the info that the value was set by the user as an extra prop in the transporter. Doable
Thoughts?

msevestre (commented):
also added some missing tests

msevestre:
@Yuri05 Thoughts?

Yuri05 (approved):

Yuri05:
seems ok :)

PR #2155, opened 2022-03-31
2153 template issue

msevestre:


src/PKSim.Infrastructure/ORM/Repositories/RemoteTemplateRepository.cs
```diff
@@ -85,8 +85,16 @@ public async Task UpdateLocalTemplateSummaryFile()
 
       private IReadOnlyList<RemoteTemplate> expressionProfileFor(RemoteTemplate remoteTemplate, ISimulationSubject simulationSubject)
       {
-         //TODO Not implemented yet. It will be done with the profile defined for individual as separate building block
-         return Array.Empty<RemoteTemplate>();
+         //This will for now assume that the expression profile templates are in the same file
+         return simulationSubject.AllExpressionProfiles()
```
msevestre, line 89: For now I think this is a good enough assumptions. But I don't think this will work at that stage because simulation subject expression profiles are not set....we'll need to investigate a bit more

src/PKSim.Presentation/Presenters/Individuals/MoleculesPresenter.cs
```diff
@@ -146,14 +147,17 @@ public void ActivateNode(ITreeNode node)
          var rootNode = node.ParentNode.DowncastTo<RootNode>();
          var molecule = moleculeFrom(node);
          var expressionProfile = _simulationSubject.ExpressionProfileFor(molecule);
+         if (expressionProfile == null)
```
msevestre, line 150: This is where the expression profile might be null

src/PKSim.Presentation/Presenters/TemplatePresenter.cs
```diff
@@ -121,6 +120,10 @@ private bool shouldLoadTemplateWithReferences(TemplateType templateType)
          if (!selectionSupportsReference(templateType))
             return false;
 
+         //For simulation subject always save linked expression profiles
+         if (templateType.Is(TemplateType.SimulationSubject))
```
msevestre, line 124: Not saving the expression profile will result in a corrupted individual. It needs to be saved

msevestre (commented):
added a  bucnh of tests here to cover the usual use cases

Yuri05 (approved):
don't see anything wrong, however also don't really understand what happens here

msevestre:
The idea is to be able to load an individual from template using expression profile
Try in your version
Create Indiviudal with Expression Profile
Save individual to template
Load
Observe the problem
1- Expression profile is not there
2- open individual and click on the molecule => crash


This implementations ssavesd the expression profiles automatically to the db (user is not asked). 
If the expression profile is removed after the fact from db, the individual will still be corrupted. However the UI won't crahs

PR #2174, opened 2022-04-05
Fixes #2121 default values

### LINKED ISSUE #2121: Default value for the "Phenotype"-field in the expression profile
I propose to have a default value, e.g. "Healthy", for the "Phenotype"-field. This would give a user a better understanding of what is required here, and contribute to some consistency of the naming across different models.

msevestre:


src/PKSim.Core/Model/ExpressionProfileFactory.cs
```diff
@@ -50,6 +55,7 @@ public ExpressionProfile Create(Type moleculeType, Species species, string molec
       {
          var expressionProfile = _objectBaseFactory.Create<ExpressionProfile>();
          expressionProfile.IsLoaded = true;
+         expressionProfile.Category = _usedExpressionProfileCategoryRepository.All().FirstOrDefault();
```
msevestre, line 58: Default is not always healthy but the first one in the list. I think it makes much more sense if the project does not deal with healthy individual and one always has to change the value to sthg that was input previously. So either empty,. or this

msevestre:
@Yuri05 Merging as discussed this morning

PR #2177, opened 2022-04-06
Fixes #2175 renaming expression profile

### LINKED ISSUE #2175: Renaming Protein Expressions Building Block corrupts it
In the project attached some strange things happen with the protein expressions
[Pgp_Only.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/8426343/Pgp_Only.zip)

I have created a transport protein expression BB, changed the (global) transporter type to PgP and cloned this transporter 3 times
(more precisely, I cloned P1=>P2, P2=>P3, P3=>P4 (or P2 => P4))
Then I also renamed all created Protein expressions BBs.

Now I face different problems in the project:
1) In the individual BB, where I added all those transporter, they are all shown as Efflux (and if I export to MoBi - yes, they are all efflux). Only the "global" direction setting is shown as pgp
<details><summary><b>Screenshot</b></summary>

grafik

</details>

2) When I open a Protein expression-BB: only the vascular system part is shown - nothing else
<details><summary><b>Screenshot</b></summary>

grafik

</details>

After some playing around: it's a renaming which corrupts the building block. E.g. if I do the following in a new empty project:
* Create Transporter Protein expression Building Block `T1|Human|Healthy`
* Set global transporter type to PgP and close the Protein Expression View
* Rename the Building Block to `T2|Human|Healthy`
* Create human individual and add `T2|Human|Healthy` to it

Then I observed both problems from above: Transporter type in the individual is Efflux in every organ (only the global transporter type is shown as PgP) and reopening the Transporter-BB view shows only the vascular system, no organs.

Cloning works, so if I clone `T1|Human|Healthy` to `T2|Human|Healthy` - everything works as expected

msevestre:


src/PKSim.Core/Model/ExpressionProfile.cs
```diff
@@ -42,7 +42,12 @@ public override string Name
          set
          {
             if (string.Equals(Name, value))
+            {
+               //Even if the name has not changed, we raise the even here as the underlying composite name may have changed
```
msevestre, line 46: This is key to see the Node update on the right. Otherwise, because we have composite names, the name has changed already when the molecule was renamed

msevestre (commented):
That should fix the issue
@Yuri05 I could not reproduce the problem with Pgp stuff. So LEt's merge and try again?

Yuri05 (approved):

PR #2184, opened 2022-04-08
Fixes #2183 wrong error range

### LINKED ISSUE #2183: Range error should show error in display unit and not in base unit
Showing the error here
image

msevestre:


src/PKSim.Core/Model/ParameterRange.cs
```diff
@@ -86,71 +92,35 @@ public override string ToString()
...
-         {
-            get
-            {
-               return CreateRule.For<ParameterRange>()
-                  .Property(item => item.MaxValueInDisplayUnit)
-                  .WithRule(maxGreaterThanDbMin)
-                  .WithError((param, value) => PKSimConstants.Rules.Parameter.MaxGreaterThanDbMinValue(param.ParameterName, param.DbMinValue, param.Unit.ToString()));
-            }
-         }
+         public static IBusinessRule MinLessThanMax { get; } = CreateRule.For<ParameterRange>()
```
msevestre, line 95: @Yuri05 Just creating the variable once instead of every time. much more efficvient

src/PKSim.Core/Model/PopulationAnalyses/PopulationAnalysisFieldBase.cs
```diff
@@ -62,30 +62,21 @@ public static IEnumerable<IBusinessRule> All
...
-         private static IBusinessRule nameNotEmpty
-         {
-            get { return GenericRules.NonEmptyRule<PopulationAnalysisFieldBase>(x => x.Name); }
-         }
+         private static IBusinessRule nameNotEmpty { get; } = GenericRules.NonEmptyRule<PopulationAnalysisFieldBase>(x => x.Name);
 
-         private static IBusinessRule nameUnique
-         {
-            get
+         private static IBusinessRule nameUnique { get; } = CreateRule.For<PopulationAnalysisFieldBase>()
```
msevestre, line 67: Also changing some other rules to be more efficient while I am at it

PR #2189, opened 2022-04-12
Fixes #2188 jacobian option

### LINKED ISSUE #2188: PK-Sim Batch: Add options to disable, enable use as is Jacobian

msevestre:


src/PKSim.BatchTool/BatchRegister.cs
```diff
@@ -24,9 +23,6 @@ public override void RegisterInContainer(IContainer container)
          container.Register<IDiagramModelToXmlMapper, CLIDiagramModelToXmlMapper>();
          container.Register<IDiagramModel, CLIDiagramModel>();
          container.Register<IJournalDiagramManagerFactory, CLIJournalDiagramManagerFactory>();
-
-
-         container.Register(typeof(IInputAndOutputBatchView<>), typeof(InputAndOutputBatchView<>));
```
msevestre, line 29: I have removed this abstraction. It made life too complicated and we only have one implementation

src/PKSim.Core/Services/SimulationRunner.cs
```diff
@@ -71,8 +71,26 @@ public void StopSimulation()
          else
             updatePersistableFromSettings(simulation);
 
+         updateSolverSettings(simulation);
          await simulationEngine.RunAsync(simulation, _simulationRunOptions);
          simulation.HasChanged = true;
       }
+
+      private void updateSolverSettings<TSimulation>(TSimulation simulation) where TSimulation : Simulation
```
msevestre, line 79: it goes deep into the simulation runner., Default is AsIs and it should not change anything

Yuri05 (approved):

Issue #2194, opened 2022-04-13
V11: Wrong values when importing from Snapshot


Yuri05:
[Human_MultipleIV_Efflux.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/8479915/Human_MultipleIV_Efflux.zip)
E.g. open the Snapshot attached. It contains a transporter with parameters defined like this:
<img width="400" src="https://user-images.githubusercontent.com/25061876/163142911-7d1687fe-441a-4313-9f8b-350d288f484d.png">

When opening the snapshot in V10 - parameters are created correctly:
![grafik](https://user-images.githubusercontent.com/25061876/163143169-c840321d-466a-4bbf-b45c-26af3e2ee146.png)

In V11:
![grafik](https://user-images.githubusercontent.com/25061876/163143258-eb1fcf6d-d98c-413d-ae15-f51f832a6bd2.png)

So the reference concentration is wrong.
t1/2 values are correct, but shown in min instead of in h

PR #2196, opened 2022-04-13
WIP #2194 Wrong values in Snapshot

msevestre:


src/PKSim.Core/Snapshots/Mappers/ExpressionProfileMapper.cs
```diff
@@ -154,8 +154,16 @@ private Localization retrieveLocalizationFrom(SnapshotExpressionProfile snapshot
 
       private bool isV9Format(SnapshotExpressionProfile snapshot)
       {
-         //Localization is only defined for protein and was added in v10
-         return snapshot.Localization == null && snapshot.TransportType == null;
+         //This was a protein
```
msevestre, line 157: That was the problem here. Trying to decide if this is a v9 format or not
In fact, we have a project version defined in the project file .I think this is what I should have used instead. However, we need to pass it top down which we do not have at the moment. A good enhancement that I might had before we release

tests/PKSim.Tests/Data/ind_expression_trans_v9.json
```diff
@@ -0,0 +1,71 @@
+{
+  "Version": 77,
```
msevestre, line 2: here the version that corresponds to the version in the PKSIm project

   public static readonly ProjectVersion V9 = addVersion(77, "9");
      public static readonly ProjectVersion V10 = addVersion(78, "10");
      public static readonly ProjectVersion V11 = addVersion(79, "11");

Yuri05 (approved):

PR #2205, opened 2022-04-14
Fixes #1506 update value unstirred water

### LINKED ISSUE #1506: Value origin for unstirred water layer thickness is wrong
The value origin for the unstirred water layer thickness in the particle dissolution model seems to be wrong:
Annotation 2020-06-17 112529

In the [referenced publication](https://www.sciencedirect.com/science/article/pii/S0939641110001517), Table 3, a value of 30 µm is given.

@Incei 
@KatrinCoboeken

msevestre:


src/Db/Dump/PKSimDB_dump.txt
```diff
@@ -3780,7 +3780,7 @@ INSERT INTO tab_calculation_method_rates VALUES('Formulation','PARAM_Subcutaneou
 INSERT INTO tab_calculation_method_rates VALUES('Formulation','PARAM_Tablet_alpha','ln(2)/DissTime50^DissShape','Inversed time');
 INSERT INTO tab_calculation_method_rates VALUES('Formulation','PARAM_Tablet_beta','0.92','Dimensionless');
 INSERT INTO tab_calculation_method_rates VALUES('Formulation','PARAM_w_fact','0.2','Dimensionless');
-INSERT INTO tab_calculation_method_rates VALUES('Formulation','PARAM_WaterLayerThickness','0.0002','Length');
+INSERT INTO tab_calculation_method_rates VALUES('Formulation','PARAM_WaterLayerThickness','0.0003','Length');
```
msevestre, line 3783: @Yuri05 Here we go. Let see if something breaks because of this in the tests

Yuri05:
Yep, I think the value was wrong and 30 is indeed the correct value. It's also 30 in the initial paper (https://pubs.acs.org/doi/10.1021/la800694j) referenced by Willmann et al.

Yuri05 (approved):

Yuri05:
That was clear...
grafik

We could change the thickness parameter in the reference simulation as well, then the tests should be green again
https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/develop/tests/PKSim.Tests/Data/PrototypeParticlesDissolution_722.pkml

msevestre:
value changed. Test running again

msevestre:
@Yuri05 I have no idea what this test is doing/ It is now failing for obscure reasons (variable became negative or outputs not the same). Can you check this out?

Yuri05:
You changed the value in the building block in the pkml, but not in the simulation - in the simulation it is still 20.
If you change it in the simulation as well - should be fine (hopefully)
grafik

msevestre:
need to update the DB> Merging and hoping for the best

tests/PKSim.Tests/IntegrationTests/SimulationWithParticlesFormulationSpecs.cs
```diff
@@ -725,6 +725,9 @@ protected override void SetupSimulation()
       {
          //disable precipitation
          PrecipitatedDrugSoluble = true;
+
+         _simulation.Solver.AbsTol = 1e-10;
+         _simulation.Solver.RelTol = 1e-6;
```
msevestre, line 730: @Yuri05 This was required too.

PR #2206, opened 2022-04-14
Fixes #2187 blood sampling order

### LINKED ISSUE #2187: fBlood for sampling: Organs/Mucosa several times
In the newly visible Sampling/Fraction of Blood parameter group the categories Organs and Mucosa appear several times instead of only once
grafik

msevestre:


msevestre:
Trivial

PR #2210, opened 2022-04-22
Make PKSim compile again after breaking changes in core

msevestre:


msevestre:
@Yuri05 Letting all tests run and then if this is good, I'll merge
This is mostly about removing the name of quantity info as well as a bunch of other stuff due to compilation error (snapshot mapping etc)

PR #2216, opened 2022-04-29
2215 load disease state pop

msevestre:


src/PKSim.Core/Snapshots/Mappers/ExpressionProfileMapper.cs
```diff
@@ -90,10 +90,12 @@ public override async Task<ModelExpressionProfile> MapToModel(SnapshotExpression
          expressionProfile.Category = snapshot.Category;
 
          var (molecule, individual) = expressionProfile;
-         await _parameterMapper.MapLocalizedParameters(snapshot.Parameters, individual, snapshotContext, !snapshotContext.IsV9FormatOrEarlier);
-
+         //Update molecule properties first
          updateMoleculePropertiesToMolecule(molecule, snapshot, individual, snapshotContext);
```
msevestre, line 94: THis may set some default parameters which were set line before. We need to swap those lines

Yuri05 (approved):

msevestre:
This fixes all the CKD issues that I had. I think we are good

msevestre:
this was not merged

PR #2220, opened 2022-05-02
Fixes #2118 calculated values show error

### LINKED ISSUE #2118: Define VolumePlasma as SumFormula
see https://github.com/Open-Systems-Pharmacology/Forum/discussions/1174#discussioncomment-2371523

@Yuri05 Do you know why it's still a hard coded formula?
How could we defined it dynamically? Using Plasma as a tag I guess? Thoughts?

msevestre:


src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -1469,26 +1469,6 @@ public static string MinLessThanDbMaxValue(string parameterName, string displayM
                return $"Minimum value for {parameterName} should be less than or equal to {displayMaxValue} {unit}.";
             }
 
-            public static string ValueSmallerThanMax(string parameterName, string value, string unit)
```
msevestre, line 1472: not used anymore

src/PKSim.Presentation/DTO/Parameters/ParameterDTO.cs
```diff
@@ -78,8 +77,13 @@ public virtual double Value
       {
          get
          {
-            var (result,_) = Parameter.TryGetValueInDisplayUnit();
+            var (result, success) = Parameter.TryGetValueInDisplayUnit();
             //Permeability cannot be read in compound as references cannot be resolved
+            //We return zero for now so that the rules are not showing a broken state (>=0). 
+            //We probably need to update the rules framework to not evaluate on NaN.
+            if (!success)
+               return 0;
```
msevestre, line 85: This was returnin NaN. But in fact, we have a rule that the value should be 0 or greater and this was breaking the rule
So returning 0 for now until we make it more resiliehnt

src/PKSim.UI/Views/Core/CalculationMethodSelectionViewBase.cs
```diff
@@ -62,6 +62,8 @@ public override void InitializeBinding()
       public void BindTo(IEnumerable<CategoryCalculationMethodDTO> calculationMethodDtos)
       {
          _gridViewBinder.BindToSource(calculationMethodDtos);
+         //This is required to ensure that the view resizes itself as expected
+         AdjustHeight();
```
msevestre, line 66: this was missing. I could see someweird UI behavior

PR #2222, opened 2022-05-02
Fixes #2221 diffrerence acebutololl

### LINKED ISSUE #2221: Loading from Snapshot. Acebutolol results are different

msevestre:


src/PKSim.Infrastructure/ProjectConverter/v11/Converter10to11.cs
```diff
@@ -191,12 +191,17 @@ private void addExpressionProfilesUsedBySimulationSubjectToProject(ISimulationSu
...
 
             //Some parameters are probably marked as FixedValue event thought they have not changed (Formula=>constant) due to change in 
-            //definition of Fraction expressed basolateral going from a constant to a formula. We reset
+            //definition of Fraction expressed basolateral going from a constant to a formula. We reset is fixed value and default state
             expressionProfile.Individual.AllMoleculeParametersFor(expressionProfile.Molecule)
                .Where(x => !x.Visible)
                .Where(x => x.IsFixedValue)
                .Each(x => x.IsFixedValue = false);
 
+            expressionProfile.Individual.AllMoleculeParametersFor(expressionProfile.Molecule)
```
msevestre, line 200: @Yuri05 This is actually not fixing the issue but making sure we do not generate entries parameter that should not have changed in the first place

src/PKSim.Infrastructure/ProjectConverter/v11/Converter10to11.cs
```diff
@@ -191,12 +191,17 @@ private void addExpressionProfilesUsedBySimulationSubjectToProject(ISimulationSu
...
+               .Where(x => !x.Visible)
+               .Where(x => !x.IsDefault)
+               .Each(x => x.IsDefault = true);
```
Yuri05, line 203: not _visible_ or not _editable_?
visible but not editable parameters cannot change as well

msevestre: That should be equivalent in this case. I'll check it out

Yuri05 (approved):

PR #2227, opened 2022-05-10
include pk properties for population simulations

### LINKED ISSUE #2021: Include PK Properties for Population Simulations
@abdelr commented on [Tue Dec 21 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1406)

---

@PavelBal commented on [Tue Dec 21 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1406#issuecomment-998667985)

## Problem

For population simulations, not all of the PK parameters are calculated. Furthermore, the values are not calculated for the selected ranges (e.g., 5%-95%).
Following PK-parameters should be calculated for the selected outputs and are displayed but not calculated (values are NaN, see Figure 1):
•	AUC_inf_norm
•	AUC_tEnd_norm
•	C_max_norm
•	Total body clearance/F (not quite clear what it is, documentation missing)
•	Vd(plasma)/F
•	Vss(plasma)/F

image
Figure 1

Following parameters are calculated in individual simulation for all compounds and are not calculated for population simulations:
•	AUC Ratio (for DDI simulations only)
•	C_max Ratio (for DDI simulations only)
•	Total plasma clearance
•	Vd (plasma)
•	Vss(phys-chem) (iv simulations)
•	VSS(plasma)
•	Bioavailability (oral simulations)
•	Fraction absorbed (oral simulations)

## Proposed solution

1)	Present PK-parameter values for the selected ranges. PK-parameters for the selected outputs should be presented for all selected outputs, including ranges. In particular, these are:
•	Arithmetic Standard Deviation – present as 1234
•	Geometric Standard Deviation - present as 1234
•	Range 5% to 95% - show two columns for 5% and 95% percentiles
•	Range 2.5% to 97.5% - show two columns for 2.5% and 97.5% percentiles

2)	Investigate why some PK-Parameters are calculated as “NaN” and fix if possible
3)	Enable calculation of missing PK-Parameters. For parameters that require additional simulation runs (AUC_ratio, C_max_ratio, bioavailability), show a warning message to the user that this will require calculation of the whole population and might take much time.
It remains to clarify for which outputs these parameters should be calculated. Arithmetic mean in venous blood plasma?

abdelr:
Fix #2021 (include pk properties for population simulations)


abdelr:
@PavelBal: Just created a PR for easier revision but still in progress. Currently, it uses the mean for the population global PK. It also expects the user to select the [peripheral] venous blood plasma as an output. Still missing the second option for the normal PK.

msevestre:
FYI @Open-Systems-Pharmacology/suite-core-developers 

This should not be merged before we release v11

Yuri05:
>I do not think it will be ready before we release

Well.. who knows 😉

abdelr:
> > I do not think it will be ready before we release
> 
> Well.. who knows 😉

Well, I hope the release does not get that much delayed lol

abdelr:
@msevestre, @Yuri05 and @georgeDaskalakis: Tests are green now but there are missing tests on the new logic. You could though start taking a look on the PR if you like so we do not end up with a huge set of changes. If at some point you do review the PR then I will start to add new incremental features by new PR on these branch instead of committing directly to it so it will be easier to review. For now, the new tests I will commit directly unless you tell me me not to do it since you are reviewing so I start already with the new PRs policy.

src/PKSim.Core/Mappers/PopulationPKAnalysisToDataTableMapper.cs
```diff
@@ -37,7 +37,7 @@ private void addAnalysisToTable(PopulationPKAnalysis analysis, DataTable dataTab
 
       private string getCurveName(PopulationPKAnalysis analysis)
       {
-         return analysis.CurveData.Caption;
+         return analysis.CurveData.Caption + analysis.ExtraDescription;
```
msevestre, line 40: use string interpolation instead of concatenating strings. Not a big deal here but just something to simply do all the time so that strings are not allocated for nothing

src/PKSim.Core/Model/Extensions/DataColumnExtensions.cs
```diff
@@ -17,13 +17,13 @@ public static bool BelongsTo(this DataColumn dataColumn, Simulation simulation)
...
+            return true;
+
          if (dataColumn.DataInfo.Origin != ColumnOrigins.Calculation)
             return false;
 
-         if (individualSimulationContainsColumn(dataColumn, simulation))
-            return true;
-
-         return simulation.IsAnImplementationOf<PopulationSimulation>();
+         return individualSimulationContainsColumn(dataColumn, simulation);
```
msevestre, line 26: Don't quite understand this change. Seems to be equivalent to me?

src/PKSim.Core/Model/IndividualSimulation.cs
```diff
@@ -1,69 +1,26 @@
...
 
 namespace PKSim.Core.Model
 {
    public class IndividualSimulation : Simulation
    {
       private DataRepository _dataRepository;
-      private readonly ICache<string, CompoundPK> _compoundPKCache;
 
-      public IndividualSimulation()
+      public IndividualSimulation() : base()
```
msevestre, line 18: calling base() is not required. Remove

src/PKSim.Core/Model/IndividualSimulation.cs
```diff
@@ -113,5 +70,23 @@ public override void AcceptVisitor(IVisitor visitor)
          if(HasResults)
             DataRepository.AcceptVisitor(visitor);
       }
+
+      public override DataColumn PeripheralVenousBloodColumn(string compoundName)
```
msevestre, line 74: Inline all of this for readability. They are just `forward` methods

src/PKSim.Core/Model/IndividualSimulation.cs
```diff
@@ -1,69 +1,26 @@
...
 
-      public IndividualSimulation()
+      public IndividualSimulation() : base()
       {
-         _compoundPKCache = new Cache<string, CompoundPK>(x => x.CompoundName, x => new CompoundPK());
       }
 
       public virtual IEnumerable<SimulationTimeProfileChart> TimeProfileAnalyses => Analyses.OfType<SimulationTimeProfileChart>();
 
-      public void ClearPKCache()
```
msevestre, line 26: @abdelr It would be good to comment a bit on PR. I have no idea why this is removed. This has nothing to do with pk properties for simulations

src/PKSim.Core/Model/PopulationSimulation.cs
```diff
@@ -24,7 +24,7 @@ public class PopulationSimulation : Simulation, IPopulationDataCollector, IAdvan
 
       public virtual AgingData AgingData { get; }
 
-      public PopulationSimulation()
+      public PopulationSimulation() : base()
```
msevestre, line 27: remove base. This is always called

src/PKSim.Core/Model/PopulationSimulation.cs
```diff
@@ -279,5 +279,65 @@ public override void UpdateFromOriginalSimulation(Simulation originalSimulation)
          var sourcePopSimulation = originalSimulation as PopulationSimulation;
          sourcePopSimulation?.AdvancedParameters.Each(x => AddAdvancedParameter(x, generateRandomValues: true));
       }
+
+      private DataColumn aggregateDataColumns(IEnumerable<DataColumn> columns)
```
msevestre, line 283: Move this outside of the simulation. This is clearly some kind of an extension methods. We have quite a few to calculate "aggregates" stuff already

src/PKSim.Core/Model/PopulationSimulation.cs
```diff
@@ -279,5 +279,65 @@ public override void UpdateFromOriginalSimulation(Simulation originalSimulation)
...
+      {
+         var column = columns.First();
```
msevestre, line 285: this will crash if there is no items. Clearly you are suspecting that this might be null see line below

msevestre: What does first mean? We need to comment this type of code BTW.
Do not comment the obvious, but if it is not obvious, comment

src/PKSim.Core/Model/Simulation.cs
```diff
@@ -19,6 +20,25 @@ public abstract class Simulation : PKSimBuildingBlock, ISimulation, IWithChartTe
       private readonly List<ISimulationAnalysis> _allSimulationAnalyses = new List<ISimulationAnalysis>();
       private SimulationProperties _properties;
       private SimulationResults _results;
+      private readonly ICache<string, CompoundPK> _compoundPKCache;
+
+      public virtual DataColumn PeripheralVenousBloodColumn(string compoundName)
```
msevestre, line 25: I think this should be abstract. But I am actually not sure that this belongs here at all. Why are we implementing something specific for just a few outputs? Does not make much sense to me

src/PKSim.Core/Model/Simulation.cs
```diff
@@ -45,8 +65,50 @@ protected Simulation() : base(PKSimBuildingBlockType.Simulation)
...
+
+         return _compoundPKCache[compoundName];
+      }
+
+      public void AddCompoundPK(CompoundPK compoundPK)
+      {
+         _compoundPKCache.Add(compoundPK);
       }
 
+      private double? compoundPKValue(string compoundName, Func<CompoundPK, double?> pkValueFunc)
```
msevestre, line 90: Why was this moved to the base class?

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -96,17 +96,16 @@ public IEnumerable<PopulationPKAnalysis> CalculateFor(IPopulationDataCollector p
          if (timeProfileChartData == null)
             return pkAnalyses; // there are no analyses to calculate
 
-         var allColumns = timeProfileChartData.Panes.SelectMany(x => x.Curves).Select(x =>
-               new {curveData = x, column = columnFor(x, populationDataCollector)})
-            .Where(c => c.column != null)
+         var allColumns = timeProfileChartData.Panes.SelectMany(x => x.Curves).SelectMany(x =>
```
msevestre, line 99: I don't understand this change.

src/PKSim.Core/Services/PopulationPKAnalysis.cs
```diff
@@ -7,11 +7,13 @@ public class PopulationPKAnalysis
    {
       public CurveData<TimeProfileXValue, TimeProfileYValue> CurveData;
       public PKAnalysis PKAnalysis { get; set; }
+      public string ExtraDescription { get; private set; }
 
-      public PopulationPKAnalysis(CurveData<TimeProfileXValue, TimeProfileYValue> curveData, PKAnalysis pkAnalysis)
+      public PopulationPKAnalysis(CurveData<TimeProfileXValue, TimeProfileYValue> curveData, PKAnalysis pkAnalysis, string extraDescription = "")
```
msevestre, line 12: string,empty

msevestre (changes requested):
I need to understand what is going on here. I do not see in the #2021  issue some of the stuff matching the implementation and therefore I have no idea why this was changed

msevestre:
oK i see where the request to move stuff comes from
Enable calculation of missing PK-Parameters. For parameters that require additional simulation runs (AUC_ratio, C_max_ratio, bioavailability), show a warning message to the user that this will require calculation of the whole population and might take much time.

src/PKSim.Core/Model/Extensions/DataColumnExtensions.cs
abdelr: For what I have seen in the code, when we receive a population simulation the DataInfo.Origin is not properly set. My thought was that for the case of populations it is not meaningful (or at least in this case) but as we discussed, it makes sense to see why it has not been set. So the change is indeed very similar but the only difference is that for population simulations it will not check the origin at all. Before it would return false always. I will still check why the origin is not been set.

src/PKSim.Core/Model/IndividualSimulation.cs
abdelr: Why not? There is some code in the base constructor that needs to be executed still. More specifically 
``_compoundPKCache = new Cache<string, CompoundPK>(x => x.CompoundName, x => new CompoundPK());``

src/PKSim.Core/Model/IndividualSimulation.cs
abdelr: It was moved to the base class so population simulation can also use it.

src/PKSim.Core/Model/IndividualSimulation.cs
msevestre: Because the base() constructor is always called....or am I missing something?

src/PKSim.Core/Model/PopulationSimulation.cs
abdelr: Fair enough

src/PKSim.Core/Model/PopulationSimulation.cs
```diff
@@ -279,5 +279,65 @@ public override void UpdateFromOriginalSimulation(Simulation originalSimulation)
...
+      {
+         return aggregateDataColumns(drugColumnFor(CoreConstants.Organ.VENOUS_BLOOD, CoreConstants.Compartment.PLASMA, CoreConstants.Observer.CONCENTRATION_IN_CONTAINER, compoundName));
+      }
+
+      public override DataColumn FabsOral(string compoundName)
+      {
+         return aggregateDataColumns(drugColumnFor(CoreConstants.Organ.LUMEN, CoreConstants.Observer.FABS_ORAL, CoreConstants.Observer.FABS_ORAL, compoundName));
+      }
+
+      private IEnumerable<DataColumn> drugColumnFor(string organ, string compartment, string columnName, string compoundName)
```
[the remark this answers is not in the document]
abdelr: Good catch... compoundName should be used

src/PKSim.Core/Model/Simulation.cs
abdelr: It was already defined at IndividualSimulation and although you are right, it makes some sense since we are using this specific combination ([peripheral] blood plasma) a lot. It could be on a service though but keep in mind that it is extracted differently at Population/Individual Simulation. Also, it was not abstract since there are more implementations of Simulation and there it is not required but again, makes sense that it is abstract here and only on the concrete implementations we decide whether it returns something or not.

src/PKSim.Core/Model/Simulation.cs
abdelr: So it is used by both Individual/Population Simulation. There could be one class between Simulation and Population/Individual Simulation but it sounds like an overshooting to me.

src/PKSim.Core/Services/PKAnalysesTask.cs
abdelr: It refers to range results. Before range results only contained a single result by now it is calculated at both ends so the 5 and 95 % for example instead of 5% only.

src/PKSim.Core/Services/PopulationPKAnalysis.cs
abdelr: Compiler complains, it needs to be a compile time constant

abdelr:
Not a draft anymore but please, do not merge since we need to merge first an existing PR into this branch!!

PavelBal (approved):
No remarks from user perspective any more.

Plase create a feature request for including "Global" PK-Parameters into the box-plot/scatter plot.

PavelBal:
@abdelr I thought this was merged already?

msevestre:
Let me check this one last time

msevestre:
There was a bunch of comments that were not addressed and for some reason, the documentation file is being reverted. I'll have to spend some time on this. Also the build is failing so some tests are not ok....

msevestre:
@PavelBal The tests are failing in this PR ...
image

I'll need to investigate why

msevestre:
@PavelBal 
Is this supposed to look like this?
image

msevestre:
I guess it's only for old simulation where F_Abs was not available. But still..this is really ugly

PavelBal:
> I guess it's only for old simulation where F_Abs was not available. But still..this is really ugly

Yes this will only be shown when having an older simulation... all new simulations will show the parameters right away.

msevestre:
ok. it's not because it's visible once that it should not be shown properly
image

msevestre:
ok I have it
image

and

image


and

image

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -2629,6 +2631,10 @@ public static class PKAnalysis
          public static readonly string Unit = UI.Unit;
          public static readonly string Description = "Description";
          public static readonly string Warning = "Warning";
+         public static readonly string OnCurves = "On Curves";
```
msevestre, line 2634: @abdelr @PavelBal This constants were defined as per request...but not USED! So the hard coded strings were still used in the code behing

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -40,6 +41,17 @@ public interface IPKAnalysesTask : OSPSuite.Core.Domain.Services.IPKAnalysesTask
...
+      /// <summary>
+      ///    Resolves options and use the mapper to create a PKAnalysis out of the values and a simulation for a given compound
+      /// </summary>
+      /// <param name="pkValues">values to use</param>
+      /// <param name="simulation">the simulation</param>
+      /// <param name="compound">the compound containing its name and molweight</param>
+      /// <returns></returns>
+      PKAnalysis CreatePKAnalysisFromValues(PKValues pkValues, Simulation simulation, Compound compound);
+
+      IReadOnlyList<PopulationPKAnalysis> AggregatePKAnalysis(Simulation populationDataCollector, IEnumerable<QuantityPKParameter> pkParameters, IEnumerable<StatisticalAggregation> selectedStatistics, string captionPrefix);
```
msevestre, line 54: I have no idea what this captionPrefix means. But I did not have time to investigate. Probably the name of the output but hard to tell

src/PKSim.Presentation/Presenters/Charts/EditTimeProfileAnalysisChartPresenter.cs
```diff
@@ -183,7 +183,7 @@ private void showAnalysis()
       private void calculatePKAnalysis()
       {
          var chartData = CreateChartData();
-         _pkAnalysisPresenter.CalculatePKAnalysis(PopulationDataCollector, chartData);
+         _pkAnalysisPresenter.CalculatePKAnalyses(PopulationDataCollector, chartData, PopulationAnalysisChart.PopulationAnalysis);
```
msevestre, line 186: I have rewritten this completely so that the PKCalculation was done ion pkAnalysisPresenter and not in this presenter.
TELL DON'T ASK.

msevestre (commented):
@PavelBal @Yuri05 @StephanSchaller 
The new behavior at the presenter level is absolutely not tested. I could move the whole code without having any issue with tests not working.  I do not have time to add tests but in the future, we need to ensure that we test the behavior that we are adding,. This is our only way of survival with such a small team

msevestre:
I will merge as soon as the build is running

PR #2236, opened 2022-05-20
Adding automation for selecting necessary output and running in one click

abdelr:
As discussed, this is part of the missing features on the PR already existing. The logic is split for easier reviewing.

### DIFF
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -1987,6 +1987,9 @@ namespace PKSim.Assets
          public static readonly string MolWeight = "Molecular weight [g/mol]";
          public static readonly string CalculateBioavailability = "Calculate Bioavailability";
          public static readonly string CalculateDDIRatio = "Calculate DDI Ratio";
+         public static readonly string CalculateAll = "Calculate once";
+         public static readonly string RunForResults = "Include Venous Blood to the outputs and run for results";
+         public static readonly string RunForResultsDescription = "Run_| = Run once to calculate results";
          public static readonly string ExportPKAnalysesToExcel = $"Export to {Excel}";
          public static readonly string AnatomyAndPhysiology = "Anatomy && Physiology";
          public static readonly string AnatomyAndPhysiologyText = "All anatomical and physiological parameters are set according to selected species, (population,) gender, age, weight and height except the manually changed parameters listed in {0}.";
@@ -2629,6 +2632,7 @@ namespace PKSim.Assets
          public static readonly string Unit = UI.Unit;
          public static readonly string Description = "Description";
          public static readonly string Warning = "Warning";
+         public static string VenousBloodPath(string compound) => $"Organism|VenousBlood|Plasma|{compound}|Concentration in container";
       }
 
       public static class Comparison
--- a/src/PKSim.Presentation/Presenters/Simulations/GlobalPKAnalysisPresenter.cs
+++ b/src/PKSim.Presentation/Presenters/Simulations/GlobalPKAnalysisPresenter.cs
@@ -25,8 +25,10 @@ namespace PKSim.Presentation.Presenters.Simulations
       GlobalPKAnalysis GlobalPKAnalysis { get; }
       bool ShouldCalculateBioAvailability(string compoundName, string parameterName);
       bool ShouldCalculateDDIRatio(string compoundName, string parameterName);
+      bool ShouldCalculateAll();
       void CalculateBioAvailability(string compoundName);
       void CalculateDDIRatioFor(string compoundName);
+      void CalculateAll();
       string DisplayNameFor(string parameterName);
       bool HasParameters();
    }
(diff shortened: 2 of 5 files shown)


src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -1987,6 +1987,9 @@ public static class UI
          public static readonly string MolWeight = "Molecular weight [g/mol]";
          public static readonly string CalculateBioavailability = "Calculate Bioavailability";
          public static readonly string CalculateDDIRatio = "Calculate DDI Ratio";
+         public static readonly string CalculateAll = "Calculate once";
+         public static readonly string RunForResults = "Include Venous Blood to the outputs and run for results";
+         public static readonly string RunForResultsDescription = "Run_| = Run once to calculate results";
```
msevestre, line 1992: what does this do ? the _| =

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -2629,6 +2632,7 @@ public static class PKAnalysis
          public static readonly string Unit = UI.Unit;
          public static readonly string Description = "Description";
          public static readonly string Warning = "Warning";
+         public static string VenousBloodPath(string compound) => $"Organism|VenousBlood|Plasma|{compound}|Concentration in container";
```
msevestre, line 2635: NO!! why is this hard coded like this?

src/PKSim.Presentation/Presenters/Simulations/GlobalPKAnalysisPresenter.cs
```diff
@@ -78,6 +84,11 @@ public bool ShouldCalculateDDIRatio(string compoundName, string parameterName)
                 shouldCalculateGlobalPKParameter(compoundName, parameterName, CoreConstants.PKAnalysis.C_maxRatio);
       }
 
+      public bool ShouldCalculateAll()
+      {
+         return !GlobalPKAnalysis.Children.SelectMany(x => (x as IContainer).Children).Any();
```
msevestre, line 89: This x as IContaienr looks bad. No idea why this is done like this but this should be rewritten.
And please add comment explaining what your are testing here. 
Also this could be moved to GlobalPKAnalysis object and you create a property here. Tell don't ask

src/PKSim.Presentation/Presenters/Simulations/GlobalPKAnalysisPresenter.cs
```diff
@@ -146,12 +167,14 @@ public IEnumerable<Unit> AvailableUnitsFor(string parameterName)
 
       private IParameter pkParameterNamed(string parameterName)
       {
-         return GlobalPKAnalysis.PKParameters(parameterName).First();
+         return GlobalPKAnalysis.PKParameters(parameterName).FirstOrDefault();
```
msevestre, line 170: Why this change?

src/PKSim.UI/Views/Simulations/GlobalPKAnalysisView.cs
```diff
@@ -112,6 +123,11 @@ private bool hasWarning(PivotDrillDownDataSource ds)
 
       public void BindTo(GlobalPKAnalysisDTO globalPKAnalysisDTO)
       {
+         if (globalPKAnalysisDTO.DataTable.Rows.Count == 0)
```
msevestre, line 126: This does not belong in the view but in the presenter.

msevestre (changes requested):
I feel that this is not relevant . IMO, we should add the output instead automatically

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -1987,6 +1987,9 @@ public static class UI
          public static readonly string MolWeight = "Molecular weight [g/mol]";
          public static readonly string CalculateBioavailability = "Calculate Bioavailability";
          public static readonly string CalculateDDIRatio = "Calculate DDI Ratio";
+         public static readonly string CalculateAll = "Calculate once";
+         public static readonly string RunForResults = "Include Venous Blood to the outputs and run for results";
```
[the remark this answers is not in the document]
msevestre: @abdelr This is the service here
SimulationPersistableUpdater
It has an overload for POP Simulation. 
```
 public void UpdatePersistableFromSettings(IndividualSimulation individualSimulation)
      {
         UpdateSimulationPersistable(individualSimulation);

         var organism = individualSimulation.Model.Root.Container(Constants.ORGANISM);

         individualSimulation.Compounds.Each(compound => addRequiredOutputForSimulation(organism, compound));
      }

      public void UpdatePersistableFromSettings(PopulationSimulation populationSimulation)
      {
         UpdateSimulationPersistable(populationSimulation);
      }
```

WE just nee dto make sure the two outputs are doing the same stuff. Because sometime, it's not venous blood that you need but Perivpheral venous blood etc.. So I would simply unify the call to be the same for all simulations and that's it

A few more entries calculated, but this is required to calculate PK Analysis. And if this should be done on demand, I would add ALL the one required for PK calculation and not only Venous  Blood because this is wrong.

PR #2237, opened 2022-05-30
Default outputs population simulations

abdelr:
Makes sure the population simulation also have the required outputs


abdelr:
Merging now after @msevestre comment. I agree the action is no longer needed so  remove it from code before merging. Feel free to add any comment and I will address it.

PR #2238, opened 2022-05-31
Filter body weight dependent rows from population PK values

abdelr:
Do not show NaN rows

### DIFF
--- a/src/PKSim.Core/Mappers/PKValuesToPKAnalysisMapper.cs
+++ b/src/PKSim.Core/Mappers/PKValuesToPKAnalysisMapper.cs
@@ -7,12 +7,13 @@ using OSPSuite.Core.Domain.Data;
 using IParameterFactory = PKSim.Core.Model.IParameterFactory;
 using OSPSuite.Core.Domain.PKAnalyses;
 using OSPSuite.Core.Services;
+using System.Collections.Generic;
 
 namespace PKSim.Core.Mappers
 {
    public interface IPKValuesToPKAnalysisMapper
    {
-      PKAnalysis MapFrom(DataColumn dataColumn, PKValues pkValues, PKParameterMode mode, string moleculeName);
+      PKAnalysis MapFrom(DataColumn dataColumn, PKValues pkValues, PKParameterMode mode, string moleculeName, bool forPopulation);
    }
 
    public class PKValuesToPKAnalysisMapper : IPKValuesToPKAnalysisMapper
@@ -31,13 +32,18 @@ namespace PKSim.Core.Mappers
          _displayUnitRetriever = displayUnitRetriever;
       }
 
-      public PKAnalysis MapFrom(DataColumn dataColumn, PKValues pkValues, PKParameterMode mode, string moleculeName)
+      public PKAnalysis MapFrom(DataColumn dataColumn, PKValues pkValues, PKParameterMode mode, string moleculeName, bool forPopulation)
       {
          var pk = new PKAnalysis().WithName(moleculeName);
-         _pkParameterRepository.All().Where(parameter => parameter.Mode.Is(mode)).Each(parameter => pk.Add(createPKParameter(parameter, pkValues)));
+         _pkParameterRepository.All().Where(parameter => parameter.Mode.Is(mode) && (!forPopulation || !filterOnPopulationForDisplayName(parameter.DisplayName))).Each(parameter => pk.Add(createPKParameter(parameter, pkValues)));
 
          pk.MolWeight = dataColumn.DataInfo.MolWeight;
          return pk;
+      }
+
+      private bool filterOnPopulationForDisplayName(string displayName)
+      {
+         return displayName.EndsWith("_norm") || displayName.EndsWith("/F");
       }
 
       private IParameter createPKParameter(PKParameter pkParameter, PKValues pkValues)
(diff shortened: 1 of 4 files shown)


msevestre (changes requested):

src/PKSim.Core/Mappers/PKValuesToPKAnalysisMapper.cs
```diff
@@ -35,15 +34,16 @@ public class PKValuesToPKAnalysisMapper : IPKValuesToPKAnalysisMapper
...
          var pk = new PKAnalysis().WithName(moleculeName);
-         _pkParameterRepository.All().Where(parameter => parameter.Mode.Is(mode) && (!forPopulation || !filterOnPopulationForDisplayName(parameter.DisplayName))).Each(parameter => pk.Add(createPKParameter(parameter, pkValues)));
+         _pkParameterRepository.All().Where(parameter => parameter.Mode.Is(mode) && (!forPopulation || !filterOnPopulationForDisplayName(parameter.Name, pkValues))).Each(parameter => pk.Add(createPKParameter(parameter, pkValues)));
 
          pk.MolWeight = dataColumn.DataInfo.MolWeight;
          return pk;
       }
 
-      private bool filterOnPopulationForDisplayName(string displayName)
+      private bool filterOnPopulationForDisplayName(string displayName, PKValues pkValues)
```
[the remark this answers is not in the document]
abdelr: > Rename this method. Also it this really the display name?? As key? 

I forgot to rename the parameter. Initially was the display name but it is indeed the name now.

> Last why do you use float.IsNaN and cast to float. Isn't the value a double? Then you can use double.IsNaN

It is nullable float. Maybe it is easier to read now, after last commit?

src/PKSim.Core/Mappers/PKValuesToPKAnalysisMapper.cs
```diff
@@ -31,15 +31,21 @@ public class PKValuesToPKAnalysisMapper : IPKValuesToPKAnalysisMapper
          _displayUnitRetriever = displayUnitRetriever;
       }
 
-      public PKAnalysis MapFrom(DataColumn dataColumn, PKValues pkValues, PKParameterMode mode, string moleculeName)
+      public PKAnalysis MapFrom(DataColumn dataColumn, PKValues pkValues, PKParameterMode mode, string moleculeName, bool forPopulation)
```
msevestre, line 34: I Am not sure I understand the forPopulation
I think we should filter also for individual as well no ?
/cc @PavelBal

src/PKSim.Core/Mappers/PKValuesToPKAnalysisMapper.cs
```diff
@@ -31,15 +31,21 @@ public class PKValuesToPKAnalysisMapper : IPKValuesToPKAnalysisMapper
...
       {
          var pk = new PKAnalysis().WithName(moleculeName);
-         _pkParameterRepository.All().Where(parameter => parameter.Mode.Is(mode)).Each(parameter => pk.Add(createPKParameter(parameter, pkValues)));
+         _pkParameterRepository.All().Where(parameter => parameter.Mode.Is(mode) && (!forPopulation || !filteNaNValues(parameter.Name, pkValues))).Each(parameter => pk.Add(createPKParameter(parameter, pkValues)));
```
msevestre, line 37: THis is very hard to read
` parameter.Mode.Is(mode) && (!forPopulation || !filteNaNValues(parameter.Name, pkValues)))`

I would suggest to create a method `shouldExport(parameter, pkValues, mode)` and then you can make it easy on the eyes

src/PKSim.Core/Mappers/PKValuesToPKAnalysisMapper.cs
```diff
@@ -31,15 +31,21 @@ public class PKValuesToPKAnalysisMapper : IPKValuesToPKAnalysisMapper
...
 
          pk.MolWeight = dataColumn.DataInfo.MolWeight;
          return pk;
       }
 
+      private bool filteNaNValues(string name, PKValues pkValues)
+      {
+         var pkValue = pkValues.ValueFor(name);
+         return (!pkValue.HasValue || float.IsNaN(pkValue.Value));
```
msevestre, line 46: hard for me to read without thinking about it 
(!pkValue.HasValue || float.IsNaN(pkValue.Value)); a! and a ||  
Also the method name as a typo (filte). MAybe a better name would be isPKValueDefined so that you do not have to do a ! and the code would be pkValue.HasValue && !float.isNaN(pkValue.value)


&& and ! are, I believe easier to compute than || and ! (in brain power not in computer power)

src/PKSim.Core/Mappers/PKValuesToPKAnalysisMapper.cs
abdelr: Actually, you are right, and in the process it will be easier to read

abdelr:
I disaggregated the check, I hope it is easier to read now?

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -193,7 +193,17 @@ public string DescriptionFor(string pkParameterName)
          return pkParameter.Description;
       }
 
-      private IndividualPKAnalysis calculatePKFor(DataColumn dataColumn, string moleculeName, PKCalculationOptions options, GlobalPKAnalysis globalPKAnalysis = null)
+      private IndividualPKAnalysis calculatePKForPopulation(DataColumn dataColumn, string moleculeName, PKCalculationOptions options, GlobalPKAnalysis globalPKAnalysis = null)
```
msevestre, line 196: Why do we have two methods with the same signature, the same parameters, and calling the same method returning the same value? I don't undertand

msevestre (commented):
Looks good to me aside from the two methods doing the same thing without any difference.Wondering what am I missing?

PR #2239, opened 2022-06-01
Adding two ways of aggregate on populations

abdelr:
Fixes having two tabs, one for calculations based on aggregated curve and one for calculations directly aggregated from individuals

### DIFF
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
-using System.Text.RegularExpressions;
 using OSPSuite.Assets.Extensions;
 using OSPSuite.Utility.Extensions;
 
@@ -2632,27 +2631,6 @@ namespace PKSim.Assets
          public static readonly string Warning = "Warning";
          public static readonly string OnCurve = "On curves";
          public static readonly string OnIndividuals = "On individuals";
-         public static string LowerSuffix(string text)
-         {
-            var regex = new Regex(@"^(.*)Range (\d*)% to (\d*)%");
-            var match = regex.Match(text);
-
-            if (!match.Success)
-               return text;
-
-            return $"{match.Groups[1]}{match.Groups[2]}%";
-         }
-
-         public static string UpperSuffix(string text)
-         {
-            var regex = new Regex(@"(.*)Range (\d*)% to (\d*)%");
-            var match = regex.Match(text);
-
-            if (!match.Success)
-               return text;
-
-            return $"{match.Groups[1]}{match.Groups[3]}%";
-         }
       }
 
       public static class Comparison
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using OSPSuite.Assets.Extensions;
 using OSPSuite.Utility.Extensions;
 
@@ -2631,8 +2632,27 @@ namespace PKSim.Assets
          public static readonly string Warning = "Warning";
          public static readonly string OnCurve = "On curves";
          public static readonly string OnIndividuals = "On individuals";
-         public static string LowerSubfix(string text) => $"{text} (Lower)";
-         public static string UpperSubfix(string text) => $"{text} (Upper)";
+         public static string LowerSuffix(string text)
+         {
+            var regex = new Regex(@"^(.*)Range (\d*)% to (\d*)%");
+            var match = regex.Match(text);
+
+            if (!match.Success)
+               return text;
+
+            return $"{match.Groups[1]}{match.Groups[2]}%";
+         }
+
+         public static string UpperSuffix(string text)
+         {
+            var regex = new Regex(@"(.*)Range (\d*)% to (\d*)%");
+            var match = regex.Match(text);
+
+            if (!match.Success)
+               return text;
+
+            return $"{match.Groups[1]}{match.Groups[3]}%";
+         }
       }
 
       public static class Comparison
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2225,6 +2225,7 @@ namespace PKSim.Assets
          public static readonly string Select = "Select";
          public static readonly string Healthy = "Healthy";
          public static readonly string RerunSimulationToSeeResults = "Please re-run the simulation to calculate PK-parameters.";
+         public static readonly string GlobalPKAnalysisDescription = "Values calculated for the mean population curve";
 
          public static readonly string[] DefaultExpressionProfileCategories = { Healthy };
 
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2630,7 +2630,9 @@ namespace PKSim.Assets
          public static readonly string Description = "Description";
          public static readonly string Warning = "Warning";
          public static readonly string OnCurve = "On curves";
+         public static readonly string OnCurveTooltip = "PK parameter values are calculated for the aggregated curves as shown in the chart";
          public static readonly string OnIndividuals = "On individuals";
+         public static readonly string OnIndividualsTooltip = "PK parameter values are aggregated from individual values";
       }
 
       public static class Comparison
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2629,6 +2629,8 @@ namespace PKSim.Assets
          public static readonly string Unit = UI.Unit;
          public static readonly string Description = "Description";
          public static readonly string Warning = "Warning";
+         public static readonly string OnCurve = "On curves";
+         public static readonly string OnIndividuals = "On individuals";
       }
 
       public static class Comparison
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2631,6 +2631,8 @@ namespace PKSim.Assets
          public static readonly string Warning = "Warning";
          public static readonly string OnCurve = "On curves";
          public static readonly string OnIndividuals = "On individuals";
+         public static string LowerSubfix(string text) => $"{text} (Lower)";
+         public static string UpperSubfix(string text) => $"{text} (Upper)";
       }
 
       public static class Comparison
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2224,7 +2224,7 @@ namespace PKSim.Assets
          public static readonly string DiseaseState = "Disease State";
          public static readonly string Select = "Select";
          public static readonly string Healthy = "Healthy";
-         public static readonly string RerunSimulationToSeeResults = "Please re-run the simulation to calculate PK-parameters";
+         public static readonly string RerunSimulationToSeeResults = "Please re-run the simulation to calculate PK-parameters.";
 
          public static readonly string[] DefaultExpressionProfileCategories = { Healthy };
 
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2224,6 +2224,7 @@ namespace PKSim.Assets
          public static readonly string DiseaseState = "Disease State";
          public static readonly string Select = "Select";
          public static readonly string Healthy = "Healthy";
+         public static readonly string RerunSimulationToSeeResults = "Please re-run the simulation to calculate PK-parameters";
 
          public static readonly string[] DefaultExpressionProfileCategories = { Healthy };
 
--- a/src/PKSim.Core/Services/ChartsDataCreator.cs
+++ b/src/PKSim.Core/Services/ChartsDataCreator.cs
@@ -19,8 +19,7 @@ namespace PKSim.Core.Services
       ChartData<TX, TY> CreateFor(PivotResult pivotResult);
       ChartData<TX, TY> CreateFor<TPopulationAnalysis>(PopulationAnalysisChart<TPopulationAnalysis> populationAnalysisChart, Aggregate aggregate) where TPopulationAnalysis : PopulationPivotAnalysis;
       ChartData<TX, TY> CreateFor<TPopulationAnalysis>(TPopulationAnalysis populationAnalysis, IPopulationDataCollector populationDataCollector, ObservedDataCollection observedDataCollection, Aggregate aggregate) where TPopulationAnalysis : PopulationPivotAnalysis;
-      IEnumerable<ChartData<TX, TY>> CreateSetFor<TPopulationAnalysis>(PopulationAnalysisChart<TPopulationAnalysis> populationAnalysisChart, Aggregate aggregate) where TPopulationAnalysis : PopulationPivotAnalysis;
-      IEnumerable<PopulationPKAnalysis> Aggregate(IEnumerable<StatisticalAggregation> selectedStatistics, IEnumerable<IEnumerable<PopulationPKAnalysis>> setOfAnalysis);
+      IEnumerable<PopulationPKAnalysis> Aggregate(IEnumerable<StatisticalAggregation> selectedStatistics, IReadOnlyList<Compound> compounds, IEnumerable<QuantityPKParameter> pks);
    }
 
    public abstract class ChartDataCreator<TX, TY> : IChartDataCreator<TX, TY> where TX : IXValue where TY : IYValue
@@ -47,8 +46,7 @@ namespace PKSim.Core.Services
       protected abstract bool CheckFields();
 
       protected abstract ChartData<TX, TY> BuildChartsData();
-      protected virtual IEnumerable<ChartData<TX, TY>> BuildChartsDataSet() { return null; }
-
+      
       public virtual ChartData<TX, TY> CreateFor(PivotResult pivotResult)
       {
          try
@@ -79,38 +77,6 @@ namespace PKSim.Core.Services
          }
       }
 
-      public virtual IEnumerable<ChartData<TX, TY>> CreateSetFor(PivotResult pivotResult)
-      {
-         try
-         {
-            _analysis = pivotResult.Analysis;
-            _data = pivotResult.PivotedData;
-            _aggregationName = pivotResult.AggregationName;
-            _dataColumnName = pivotResult.DataColumnName;
-            _populationDataCollector = pivotResult.PopulationDataCollector;
-            _observedDataCollection = pivotResult.ObservedDataCollection;
-
-            if (!CheckFields())
-               return null;
-
-            var charts = BuildChartsDataSet();
-            charts.Each(chart =>
-            {
-               chart.CreatePaneOrder();
-               chart.Panes.Each(x => x.CreateCurveOrder());
-            });
-
-            return charts;
-         }
-         finally
-         {
-            _analysis = null;
-            _data = null;
-            _populationDataCollector = null;
-            _observedDataCollection = null;
-         }
-      }
-
       public ChartData<TX, TY> CreateFor<TPopulationAnalysis>(PopulationAnalysisChart<TPopulationAnalysis> populationAnalysisChart, Aggregate aggregate) where TPopulationAnalysis : PopulationPivotAnalysis
       {
          return CreateFor(populationAnalysisChart.PopulationAnalysis, populationAnalysisChart.Analysable.DowncastTo<IPopulationDataCollector>(), populationAnalysisChart.ObservedDataCollection, aggregate);
@@ -124,11 +90,6 @@ namespace PKSim.Core.Services
       protected virtual PivotResult CreateResult<TPopulationAnalysis>(TPopulationAnalysis populationAnalysis, IPopulationDataCollector populationDataCollector, ObservedDataCollection observedDataCollection, Aggregate aggregate) where TPopulationAnalysis : PopulationPivotAnalysis
       {
          return _pivotResultCreator.Create(populationAnalysis, populationDataCollector, observedDataCollection, aggregate);
-      }
-
-      public IEnumerable<ChartData<TX, TY>> CreateSetFor<TPopulationAnalysis>(PopulationAnalysisChart<TPopulationAnalysis> populationAnalysisChart, Aggregate aggregate) where TPopulationAnalysis : PopulationPivotAnalysis
-      {
-         return CreateSetFor(CreateResult(populationAnalysisChart.PopulationAnalysis, populationAnalysisChart.Analysable.DowncastTo<IPopulationDataCollector>(), populationAnalysisChart.ObservedDataCollection, aggregate));
       }
 
       protected virtual IReadOnlyDictionary<string, string> GetFieldValues(IEnumerable<string> fieldNames, DataRow row)
@@ -295,9 +256,9 @@ namespace PKSim.Core.Services
          return _analysis.AllFieldsOn(area);
       }
 
-      public virtual IEnumerable<PopulationPKAnalysis> Aggregate(IEnumerable<StatisticalAggregation> selectedStatistics, IEnumerable<IEnumerable<PopulationPKAnalysis>> setOfAnalysis)
+      public virtual IEnumerable<PopulationPKAnalysis> Aggregate(IEnumerable<StatisticalAggregation> selectedStatistics, IReadOnlyList<Compound> compounds, IEnumerable<QuantityPKParameter> pks)
       {
-         return setOfAnalysis.First();
+         return null;
       }
    }
 }
--- a/src/PKSim.Core/Services/ChartsDataCreator.cs
+++ b/src/PKSim.Core/Services/ChartsDataCreator.cs
@@ -19,6 +19,8 @@ namespace PKSim.Core.Services
       ChartData<TX, TY> CreateFor(PivotResult pivotResult);
       ChartData<TX, TY> CreateFor<TPopulationAnalysis>(PopulationAnalysisChart<TPopulationAnalysis> populationAnalysisChart, Aggregate aggregate) where TPopulationAnalysis : PopulationPivotAnalysis;
       ChartData<TX, TY> CreateFor<TPopulationAnalysis>(TPopulationAnalysis populationAnalysis, IPopulationDataCollector populationDataCollector, ObservedDataCollection observedDataCollection, Aggregate aggregate) where TPopulationAnalysis : PopulationPivotAnalysis;
+      IEnumerable<ChartData<TX, TY>> CreateSetFor<TPopulationAnalysis>(PopulationAnalysisChart<TPopulationAnalysis> populationAnalysisChart, Aggregate aggregate) where TPopulationAnalysis : PopulationPivotAnalysis;
+      IEnumerable<PopulationPKAnalysis> Aggregate(IEnumerable<StatisticalAggregation> selectedStatistics, IEnumerable<IEnumerable<PopulationPKAnalysis>> setOfAnalysis);
    }
 
    public abstract class ChartDataCreator<TX, TY> : IChartDataCreator<TX, TY> where TX : IXValue where TY : IYValue
@@ -45,6 +47,7 @@ namespace PKSim.Core.Services
       protected abstract bool CheckFields();
 
       protected abstract ChartData<TX, TY> BuildChartsData();
+      protected virtual IEnumerable<ChartData<TX, TY>> BuildChartsDataSet() { return null; }
 
       public virtual ChartData<TX, TY> CreateFor(PivotResult pivotResult)
       {
@@ -76,6 +79,38 @@ namespace PKSim.Core.Services
          }
       }
 
+      public virtual IEnumerable<ChartData<TX, TY>> CreateSetFor(PivotResult pivotResult)
+      {
+         try
+         {
+            _analysis = pivotResult.Analysis;
+            _data = pivotResult.PivotedData;
+            _aggregationName = pivotResult.AggregationName;
+            _dataColumnName = pivotResult.DataColumnName;
+            _populationDataCollector = pivotResult.PopulationDataCollector;
+            _observedDataCollection = pivotResult.ObservedDataCollection;
+
+            if (!CheckFields())
+               return null;
+
+            var charts = BuildChartsDataSet();
+            charts.Each(chart =>
+            {
+               chart.CreatePaneOrder();
+               chart.Panes.Each(x => x.CreateCurveOrder());
+            });
+
+            return charts;
+         }
+         finally
+         {
+            _analysis = null;
+            _data = null;
+            _populationDataCollector = null;
+            _observedDataCollection = null;
+         }
+      }
+
       public ChartData<TX, TY> CreateFor<TPopulationAnalysis>(PopulationAnalysisChart<TPopulationAnalysis> populationAnalysisChart, Aggregate aggregate) where TPopulationAnalysis : PopulationPivotAnalysis
       {
          return CreateFor(populationAnalysisChart.PopulationAnalysis, populationAnalysisChart.Analysable.DowncastTo<IPopulationDataCollector>(), populationAnalysisChart.ObservedDataCollection, aggregate);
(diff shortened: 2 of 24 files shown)


abdelr:
This is not ready for merging yet @msevestre, it lacks tests and I want to take a look on the general implementation now. Still, now it works fast, since it does not calculate the PK for every individual but instead it uses the stored values. So, some important differences now from the last time we spoke. 

1. It runs much faster, you will not notice the lag as before.
2. It calculates all PK values. Even those dependent on body weight.
3. Values are not similar anymore. Pavel will take a look on this but very surprisingly, now the aggregated values are not really the same. So, unless I have something wrong on the code, which is highly probable, the stored values differs a bit to those calculated on the fly.

Feel free to wait for the PR to be ready to merge or start taking a look and add any comment. It will be very welcomed ;)

src/PKSim.Core/Services/ChartsDataCreator.cs
```diff
@@ -205,8 +207,8 @@ protected IEnumerable<DataRow> RowsForDataField(string fieldName)
 
       private CurveData<TX, TY> getOrCreateCurve(PaneData<TX, TY> pane, IReadOnlyDictionary<string, string> seriesFieldValues, INumericValueField yAxisField)
       {
-         string curveCaption = captionFor(seriesFieldValues.Values);
-         string curveId = idFromCaption(curveCaption);
+         string curveCaption = captionFor(new[] { seriesFieldValues[STATISTICAL_AGGREGATION_DISPLAY_NAME] });
```
msevestre, line 210: Can you comment on these changes? I am not sure I understand what's going on?

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -40,6 +40,8 @@ public interface IPKAnalysesTask : OSPSuite.Core.Domain.Services.IPKAnalysesTask
 
       IndividualPKAnalysis CalculateFor(Simulation simulation, DataColumn dataColumn);
       PKValues CalculatePK(DataColumn column, PKCalculationOptions options);
+
+      PKAnalysis MapFrom(double? molWeight, PKValues pkValues, string moleculeName, Simulation simulation);
```
msevestre, line 44: Mapping typically does not belong in task but own mapper. Why is this different?

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -145,13 +147,13 @@ private IEnumerable<DataColumn> columnsFor(CurveData<TimeProfileXValue, TimeProf
          if (curveData.IsRange())
             return new[] 
             {
-               new DataColumn(" (Lower)", curveData.YAxis.Dimension, baseGrid)
```
msevestre, line 148: I don't understnad the code on the right or on the left.  " (Lower)" ??!? Was it like this before?

src/PKSim.Core/Services/TimeProfileChartsDataCreator.cs
```diff
@@ -48,6 +49,53 @@ protected override bool CheckFields()
          return _analysis.AllFieldsOn(PivotArea.DataArea).Any() && statisticalAnalysis.SelectedStatistics.Any();
       }
 
+      public override IEnumerable<PopulationPKAnalysis> Aggregate(IEnumerable<StatisticalAggregation> selectedStatistics, IReadOnlyList<Compound> compounds, IEnumerable<QuantityPKParameter> pks, Simulation simulation)
```
msevestre, line 52: name the parameter propertlyes. not pks but pkParameters

src/PKSim.Core/Services/TimeProfileChartsDataCreator.cs
```diff
@@ -48,6 +49,53 @@ protected override bool CheckFields()
...
+      {
+         var names = pks.Select(x => x.Name).Distinct();
+         var matrix = new FloatMatrix();
+         pks.Each(pk => {
```
msevestre, line 56: no need fpr { }

src/PKSim.Presentation/Views/Simulations/IPopulationPKAnalysisView.cs
```diff
@@ -7,9 +7,9 @@ namespace PKSim.Presentation.Views.Simulations
 {
    public interface IPopulationPKAnalysisView : IView<IPopulationPKAnalysisPresenter>
    {
-      void BindTo(PKAnalysisDTO dataTable);
+      void BindTo(PKAnalysisDTO pkAnalysisOnCurveDTO, PKAnalysisDTO pkAnalysisOnIndividualsDTO);
```
msevestre, line 10: WE typicallyh only bind to ONE object. This is the DTO you create to bind to the view

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
```diff
@@ -34,33 +41,44 @@ public class PopulationPKAnalysisPresenter : PKAnalysisPresenter<IPopulationPKAn
...
       protected override void BindToPKAnalysis()
       {
-         _view.BindTo(_populationPKAnalysisToDTOMapper.MapFrom(_allAnalyses));
+         _view.BindTo(_populationPKAnalysisToDTOMapper.MapFrom(_allAnalysesOnCurves), _populationPKAnalysisToDTOMapper.MapFrom(_allAnalysesOnIndividuals));
       }
 
       protected override IEnumerable<PKAnalysis> AllPKAnalyses
       {
-         get { return _allPKAnalyses.Select(x => x.PKAnalysis); }
+         get { return (View.IsOnCurvesSelected() ? _allPKAnalysesOnCurves : _allPKAnalysesOnIndividuals).Select(x => x.PKAnalysis); }
```
msevestre, line 81: I don't undertstand the IsOnCurvesSelected. What does this mean? Property boolean starting with Is should also be a property {get}

msevestre (changes requested):
You need to walk me through the changes here. Absolutely zero line of comments, etc... Very hard to understand.

src/PKSim.Core/Services/ChartsDataCreator.cs
abdelr: image

It used to show the name of the compound on the top separator, then it repeated the name of the compound, the observer and the statistical measure on the header of each column. The thing is that the top header is shared for several statistical measures and it makes no sense to repeat then the same data again and again on all columns if they all share the common top header. So I removed al repeated data from the second header (in red in the image) and only left what makes it unique. The compound is already on the top header (in yellow in the image). Currently we do not show the name of the observer but maybe could be inferred from the compound, otherwise I would add it too but to the top header only. This way the view is more compact and does not repeat the same information again and again (which at the end you only see if you roll over it since it does not fit in the header). If you think this is not a good idea we could discuss it and we could restore it to the way it was before.

src/PKSim.Core/Services/PKAnalysesTask.cs
abdelr: It is actually wrapping the mapper. In addition, it uses the service to create the options which we do not have from the caller. So I could still use the mapper from the caller and ask only for the options to the service and then we have a better decoupling.

src/PKSim.Core/Services/ChartsDataCreator.cs
abdelr: I am tagging @PavelBal since he could also give some useful insights here.

src/PKSim.Core/Services/PKAnalysesTask.cs
abdelr: It was already merged in some previous PR. This is for the ranges. So, before we had only one value for a range parameter. In case we have 5-95 percentile it will only show the lower value (5 percentile). Now, it adds the two values as two separate columns. The first column will be <5-95 percentile> (Lower) and the second will be <5-95 percentile> (Upper). I realized it was coded using the string built on the fly and moved it to the constants (and reuse it in the table headers for both calculations - on individuals and on curves).

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
abdelr: > Property boolean starting with Is should also be a property {get}

Fixed

abdelr: > I don't undertstand the IsOnCurvesSelected.

Notice there is a view now with a tabbed view. One tab shows the calculation based on individuals and the other the same based on curves. When exporting to an excel, there were two options now: exporting both on different sheets or exporting only the selected tab. In the feature description we formalized it as "export only selected tab". So the presenter needs to know whether the selected tab is "OnCurves" or not.

abdelr:
I forced a rebase with no conflicts... hope it does not change the diff too much.

src/PKSim.Core/Services/ChartsDataCreator.cs
abdelr: Actually, the caption is used to draw the chart where we will not have the table representation so we need to keep the naming as it was before. I am reverting this changes now.

abdelr:
Tagging @msevestre to remember this is ready for revision. I think it went under the radar. After merging this into #2021 the feature should be complete and we could merge also #2227

msevestre:
@abdelr Did not receive any notifications on this. Sorry

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -2629,6 +2630,29 @@ public static class PKAnalysis
...
+
+            if (!match.Success)
+               return text;
+
+            return $"{match.Groups[1]}{match.Groups[2]}%";
+         }
+
+         public static string UpperSuffix(string text)
+         {
+            var regex = new Regex(@"(.*)Range (\d*)% to (\d*)%");
```
msevestre, line 2648: This file should really be only about constant string and putting strings together. Think almost this is the english version of PKSim. The code with regex etc.. does not belong here.
Note on the regex .They are almost identical but not quite. There is not comment explaining what it does and the code is almost exactly the same. I think this needs to be refactored. 
And this needs to be commented, if we really need regex for this

src/PKSim.Core/Services/ChartsDataCreator.cs
```diff
@@ -19,6 +19,7 @@ public interface IChartDataCreator<TX, TY> where TX : IXValue where TY : IYValue
       ChartData<TX, TY> CreateFor(PivotResult pivotResult);
       ChartData<TX, TY> CreateFor<TPopulationAnalysis>(PopulationAnalysisChart<TPopulationAnalysis> populationAnalysisChart, Aggregate aggregate) where TPopulationAnalysis : PopulationPivotAnalysis;
       ChartData<TX, TY> CreateFor<TPopulationAnalysis>(TPopulationAnalysis populationAnalysis, IPopulationDataCollector populationDataCollector, ObservedDataCollection observedDataCollection, Aggregate aggregate) where TPopulationAnalysis : PopulationPivotAnalysis;
+      IEnumerable<PopulationPKAnalysis> Aggregate(IEnumerable<StatisticalAggregation> selectedStatistics, IReadOnlyList<Compound> compounds, IEnumerable<QuantityPKParameter> pks, Simulation simulation, string captionPrefix);
```
msevestre, line 22: pks => pkParameters.

src/PKSim.Core/Services/TimeProfileChartsDataCreator.cs
```diff
@@ -48,6 +50,52 @@ protected override bool CheckFields()
...
+      }
+
+      private CurveData<TimeProfileXValue, TimeProfileYValue> buildCurveData(QuantityPKParameter pk, List<float[]> values, int index, StatisticalAggregation statisticalAggregation, string captionPrefix)
+      {
+         var suffix = _representationInfoRepository.DisplayNameFor(statisticalAggregation);
+         if (values.Count > 1)
+            suffix = index == 0 ? PKSimConstants.PKAnalysis.LowerSuffix(suffix) : PKSimConstants.PKAnalysis.UpperSuffix(suffix);
+         var caption = captionFor(new[] { captionPrefix, suffix });
+
+         var curveData = new CurveData<TimeProfileXValue, TimeProfileYValue>()
```
msevestre, line 79: return here

src/PKSim.Core/Services/TimeProfileChartsDataCreator.cs
```diff
@@ -48,6 +50,52 @@ protected override bool CheckFields()
...
+            }
+         });
+         return results;
+      }
+
+      private CurveData<TimeProfileXValue, TimeProfileYValue> buildCurveData(QuantityPKParameter pk, List<float[]> values, int index, StatisticalAggregation statisticalAggregation, string captionPrefix)
+      {
+         var suffix = _representationInfoRepository.DisplayNameFor(statisticalAggregation);
+         if (values.Count > 1)
+            suffix = index == 0 ? PKSimConstants.PKAnalysis.LowerSuffix(suffix) : PKSimConstants.PKAnalysis.UpperSuffix(suffix);
```
msevestre, line 76: Explan!! What is this index ifluencing the suffix.

src/PKSim.Core/Services/TimeProfileChartsDataCreator.cs
```diff
@@ -48,6 +50,52 @@ protected override bool CheckFields()
...
+         {
+            Id = pk.Id,
+            Caption = caption,
+            YDimension = pk.Dimension,
+            QuantityPath = pk.QuantityPath,
+         };
+         return curveData;
+      }
+
+      private PopulationPKAnalysis buildPopulationPKAnalysis(CurveData<TimeProfileXValue, TimeProfileYValue> curveData, Compound compound, float[] values, IEnumerable<string> names, Simulation simulation)
```
msevestre, line 89: make name an IReadOnly here so that you don't have to use Count() and ElementAt() etc..

src/PKSim.Presentation/Core/SingleStartPresenterTask.cs
```diff
@@ -37,7 +37,7 @@ public void StartForSubject<T>(T subject)
          {
             presenter.Edit(subject);
          }
-         catch (Exception)
+         catch (Exception e)
```
msevestre, line 40: where is the e being useD?

src/PKSim.Presentation/Presenters/Charts/EditTimeProfileAnalysisChartPresenter.cs
```diff
@@ -183,7 +188,13 @@ private void showAnalysis()
       private void calculatePKAnalysis()
       {
          var chartData = CreateChartData();
-         _pkAnalysisPresenter.CalculatePKAnalysis(PopulationDataCollector, chartData);
+         _pkAnalysisPresenter.CalculatePKAnalysisOnCurves(PopulationDataCollector, chartData);
+
+         var fields = PopulationAnalysisChart.PopulationAnalysis.AllFields.OfType<PopulationAnalysisOutputField>().Select(x => x.QuantityPath);
+         var pks = fields.SelectMany(x => ((PopulationSimulation)PopulationDataCollector).PKAnalyses.AllPKParametersFor(x));
```
msevestre, line 194: why do you kno where that this is a PopulationSimulation? This  cast is suspicious but I am not sure

msevestre: I think it's used in population comparison. This would crash

msevestre (changes requested):
This needs to be reviewed. A lot of the code does not seem to be tested also. I need to take a closer look. I identified a bunch of refactoring potential, but also i believe a crash with the cast. 
Can you try to add some comments, specifically when using low level stuff such as index, loop etc... where the decisions made is not trivial

abdelr:
@msevestre, I think you might review this now. I cannot read why it is failing but I actually did not expect the tests to run on this since it is a PR on a feature branch. Could you tell me if there are failing tests please?

abdelr:
> @msevestre, I think you might review this now. I cannot read why it is failing but I actually did not expect the tests to run on this since it is a PR on a feature branch. Could you tell me if there are failing tests please?

Never mind... there are merge conflicts

abdelr:
Should be fine now @msevestre

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -40,6 +41,10 @@ public interface IPKAnalysesTask : OSPSuite.Core.Domain.Services.IPKAnalysesTask
 
       IndividualPKAnalysis CalculateFor(Simulation simulation, DataColumn dataColumn);
       PKValues CalculatePK(DataColumn column, PKCalculationOptions options);
+
+      PKAnalysis CreatePKAnalysisFromValues(PKValues pkValues, Simulation simulation, Compound compound);
+      string LowerSuffix(string text);
```
msevestre, line 46: Add a comment to this interface (and the one below) explaining what this methods do. Also rename parameter text to something meaningful

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -251,5 +258,45 @@ private void updateBodyWeightFromCurrentIndividual(IParameter bodyWeightParamete
...
          bodyWeightParameter.Value = allBodyWeights.Count > individualId ? allBodyWeights[individualId] : double.NaN;
       }
+
+      public PKAnalysis CreatePKAnalysisFromValues(PKValues pkValues, Simulation simulation, Compound compound)
+      {
+         var options = _pkCalculationOptionsFactory.CreateFor(simulation, compound.Name);
+         return _pkMapper.MapFrom(compound.MolWeight, pkValues, options.PKParameterMode, compound.Name);
+      }
+
+      private Regex _rangeRegex => new Regex(@"^(.*)Range (\d*)% to (\d*)%");
```
msevestre, line 268: This is a field. Move it at the top
Also why do you instantiate it all the time? It can be a readonly field, no?

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -251,5 +258,45 @@ private void updateBodyWeightFromCurrentIndividual(IParameter bodyWeightParamete
...
+      }
+
+      /// <summary>
+      /// Will extract the range values from texts with the following format: Range 5% to 95%
+      /// </summary>
+      /// <param name="text">Range 5% to 95%</param>
+      /// <returns>95%</returns>
+      public string UpperSuffix(string text)
+      {
+         var regex = _rangeRegex;
```
msevestre, line 293: same here. Use the field

src/PKSim.Presentation/Presenters/Charts/EditTimeProfileAnalysisChartPresenter.cs
```diff
@@ -79,6 +89,55 @@ protected override void RefreshData()
       {
          return _chartDataCreator.CreateFor(PopulationAnalysisChart, AggregationFunctions.QuantityAggregation);
       }
+      
+      protected IEnumerable<PopulationPKAnalysis> aggregatePKAnalysis(IPopulationDataCollector populationDataCollector, IEnumerable<QuantityPKParameter> pkParameters, string captionPrefix)
```
msevestre, line 93: This does not belong in the presenter. There is nothing related to presentation here. Move it so that it can be tested in isolation?

src/PKSim.Presentation/Presenters/Charts/EditTimeProfileAnalysisChartPresenter.cs
```diff
@@ -79,6 +89,55 @@ protected override void RefreshData()
...
+            {
+               var pk = pkParameters.ElementAt(aggregationIndex);
+               var curveData = buildCurveData(pk, aggregated, aggregationIndex, statisticalAnalysis, captionPrefix);
+               results.Add(buildPopulationPKAnalysis(curveData, populationDataCollector.Compounds.First(x => simulation.Model.MoleculeNameFor(curveData.QuantityPath) == x.Name), aggregated[aggregationIndex], names, simulation));
+            }
+         });
+         return results;
+      }
+
+      private CurveData<TimeProfileXValue, TimeProfileYValue> buildCurveData(QuantityPKParameter pk, List<float[]> values, int index, StatisticalAggregation statisticalAggregation, string captionPrefix)
```
msevestre, line 114: it feels that the values and index are only use to create the caption. Maybe a method that returns the caption can be created and the caption passe dto this function insteaD? Also this is not presentation related and could be moved

src/PKSim.Presentation/Presenters/Charts/EditTimeProfileAnalysisChartPresenter.cs
```diff
@@ -79,6 +89,55 @@ protected override void RefreshData()
...
+      {
+         var suffix = _representationInfoRepository.DisplayNameFor(statisticalAggregation);
+         //For those metrics returning two values, the first is the lower value and the second
+         //is the upper value so depending on the index we use lower or upper suffix.
+         if (values.Count > 1)
+            suffix = index == 0 ? _pKAnalysesTask.LowerSuffix(suffix) : _pKAnalysesTask.UpperSuffix(suffix);
+         var caption = (new[] { captionPrefix, suffix }).CaptionFrom();
```
msevestre, line 121: CaptionFrom wthout does not read well here because there is no from.

src/PKSim.Presentation/Presenters/Charts/EditTimeProfileAnalysisChartPresenter.cs
```diff
@@ -180,10 +239,34 @@ private void showAnalysis()
...
+      private bool populationDataCollectorSupportsDifferentAggregations(IPopulationDataCollector populationDataCollector)
+      {
+         return populationDataCollector is PopulationSimulation;
+      }
+
+      private IEnumerable<QuantityPKParameter> extractPKParameters(PopulationSimulation populationSimulation)
+      {
+         if (populationSimulation == null)
+            return Enumerable.Empty<QuantityPKParameter>();
+         var fields = PopulationAnalysisChart.PopulationAnalysis.AllFields.OfType<PopulationAnalysisOutputField>().Select(x => x.QuantityPath);
```
msevestre, line 251: Empty line after return . Can you run resharper?

src/PKSim.Presentation/Presenters/Charts/EditTimeProfileAnalysisChartPresenter.cs
```diff
@@ -180,10 +239,34 @@ private void showAnalysis()
...
+      }
+
       private void calculatePKAnalysis()
       {
          var chartData = CreateChartData();
-         _pkAnalysisPresenter.CalculatePKAnalysis(PopulationDataCollector, chartData);
+         _pkAnalysisPresenter.CalculatePKAnalysisOnCurves(PopulationDataCollector, chartData);
+         _pkAnalysisPresenter.PKAnalysisOnIndividualsEnabled = false;
+         if (!populationDataCollectorSupportsDifferentAggregations(PopulationDataCollector))
+            return;
```
msevestre, line 261: Empty line after return . Can you run resharper? (hopefully this does it for you)

src/PKSim.Presentation/Presenters/Charts/EditTimeProfileAnalysisChartPresenter.cs
```diff
@@ -180,10 +239,34 @@ private void showAnalysis()
...
       }
 
+      private bool populationDataCollectorSupportsDifferentAggregations(IPopulationDataCollector populationDataCollector)
+      {
+         return populationDataCollector is PopulationSimulation;
+      }
+
+      private IEnumerable<QuantityPKParameter> extractPKParameters(PopulationSimulation populationSimulation)
+      {
+         if (populationSimulation == null)
```
msevestre, line 249: can it be null here? private method only used after the cast.

src/PKSim.Presentation/Presenters/Charts/EditTimeProfileAnalysisChartPresenter.cs
```diff
@@ -180,10 +239,34 @@ private void showAnalysis()
...
          var chartData = CreateChartData();
-         _pkAnalysisPresenter.CalculatePKAnalysis(PopulationDataCollector, chartData);
+         _pkAnalysisPresenter.CalculatePKAnalysisOnCurves(PopulationDataCollector, chartData);
+         _pkAnalysisPresenter.PKAnalysisOnIndividualsEnabled = false;
+         if (!populationDataCollectorSupportsDifferentAggregations(PopulationDataCollector))
+            return;
+         _pkAnalysisPresenter.PKAnalysisOnIndividualsEnabled = true;
+
+         var pks = extractPKParameters(PopulationDataCollector.DowncastTo<PopulationSimulation>());
+         var captionPrefix = PopulationAnalysisChart.PopulationAnalysis.AllFieldNamesOn(PivotArea.DataArea);
```
msevestre, line 265: move after

src/PKSim.Presentation/Presenters/Charts/EditTimeProfileAnalysisChartPresenter.cs
```diff
@@ -180,10 +239,34 @@ private void showAnalysis()
...
       {
          var chartData = CreateChartData();
-         _pkAnalysisPresenter.CalculatePKAnalysis(PopulationDataCollector, chartData);
+         _pkAnalysisPresenter.CalculatePKAnalysisOnCurves(PopulationDataCollector, chartData);
+         _pkAnalysisPresenter.PKAnalysisOnIndividualsEnabled = false;
+         if (!populationDataCollectorSupportsDifferentAggregations(PopulationDataCollector))
+            return;
+         _pkAnalysisPresenter.PKAnalysisOnIndividualsEnabled = true;
+
+         var pks = extractPKParameters(PopulationDataCollector.DowncastTo<PopulationSimulation>());
```
msevestre, line 264: rename to pkParameters

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
```diff
@@ -34,33 +40,67 @@ public class PopulationPKAnalysisPresenter : PKAnalysisPresenter<IPopulationPKAn
...
       }
 
-      public void CalculatePKAnalysis(IPopulationDataCollector populationDataCollector, ChartData<TimeProfileXValue, TimeProfileYValue> timeProfileChartData)
+      public void CalculatePKAnalysisOnCurves(IPopulationDataCollector populationDataCollector, ChartData<TimeProfileXValue, TimeProfileYValue> timeProfileChartData)
       {
-         _allAnalyses.Clear();
+         CalculatePKAnalysis(populationDataCollector, _allAnalysesOnCurves, _pkAnalysesTask.CalculateFor(populationDataCollector, timeProfileChartData), _allPKAnalysesOnCurves);
+      }
+
+      public void CalculatePKAnalysisOnIndividuals(IPopulationDataCollector populationDataCollector, IEnumerable<PopulationPKAnalysis> pks)
```
msevestre, line 56: rename to pkAnalyses

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
```diff
@@ -34,33 +40,67 @@ public class PopulationPKAnalysisPresenter : PKAnalysisPresenter<IPopulationPKAn
...
+      {
+         CalculatePKAnalysis(populationDataCollector, _allAnalysesOnIndividuals, pks, _allPKAnalysesOnIndividuals);
+      }
+
+      private bool _pkAnalysisOnIndividualsEnabled = true;
```
msevestre, line 61: Move fields at the top of the class

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
```diff
@@ -34,33 +40,67 @@ public class PopulationPKAnalysisPresenter : PKAnalysisPresenter<IPopulationPKAn
...
+      }
+
+      private void CalculatePKAnalysis(IPopulationDataCollector populationDataCollector, List<PopulationPKAnalysis> allAnalysis, IEnumerable<PopulationPKAnalysis> sourcePKs, List<PopulationPKAnalysis> targetPKs)
+      {
+         allAnalysis.Clear();
          _populationDataCollector = populationDataCollector;
-         _allPKAnalyses = _pkAnalysesTask.CalculateFor(populationDataCollector, timeProfileChartData);
-         _allAnalyses.AddRange(_allPKAnalyses);
+         targetPKs.Clear();
+         targetPKs.AddRange(sourcePKs);
```
msevestre, line 80: Do not use PK. You are using the same acronyms for PK Parameter, PK analyses, which is in both cases wrong
PK means PharmacoKinetic.

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
```diff
@@ -34,33 +40,67 @@ public class PopulationPKAnalysisPresenter : PKAnalysisPresenter<IPopulationPKAn
...
+         allAnalysis.AddRange(targetPKs);
          LoadPreferredUnitsForPKAnalysis();
          BindToPKAnalysis();
-
          _globalPKAnalysisPresenter.CalculatePKAnalysis(new Simulation[] { populationDataCollector as Simulation });
       }
 
+      public void HandleTabChanged()
+      {
+         BindToPKAnalysis();
```
msevestre, line 89: why do you recreate on each tab change? Cannont you cache the results once? Everyting is static

src/PKSim.UI/Views/Simulations/GlobalPKAnalysisView.Designer.cs
```diff
@@ -31,75 +31,104 @@ protected override void Dispose(bool disposing)
...
          this.layoutControl.Text = "layoutControl1";
          // 
+         // labelControl
+         // 
+         this.labelControl.Location = new System.Drawing.Point(2, 493);
+         this.labelControl.Name = "labelControl";
+         this.labelControl.Size = new System.Drawing.Size(319, 16);
+         this.labelControl.StyleController = this.layoutControl;
+         this.labelControl.TabIndex = 5;
+         this.labelControl.Text = "Please re-run the simulation to calculate PK-parameters";
```
msevestre, line 71: No hard coded text in designer

src/PKSim.UI/Views/Simulations/PopulationPKAnalysisView.cs
```diff
@@ -29,26 +31,41 @@ public void AddGlobalPKAnalysisView(IGlobalPKAnalysisView view)
...
+      }
+
+      public bool IsOnCurvesSelected
+      {
+         get => populationPKAnalysisXtraTabControl.SelectedTabPageIndex == 0;
+      }
+
+      public void EnablePKAnalysisOnIndividualsTab(bool enabled)
+      {
+         this.xtraTabPage4.PageEnabled = enabled;
```
msevestre, line 54: That tab needs to be renamed...please

tests/PKSim.Tests/Presentation/EditTimeProfileAnalysisChartPresenterSpecs.cs
```diff
@@ -258,7 +266,100 @@ protected override void Context()
...
+            new TestQuantityPKParameter() { Name = "Name 3", QuantityPath = "Organism|PeripheralVenousBlood|Esomeprazole|Plasma (Peripheral Venous Blood)",   _values = new[] { 0.0f,   2.5f,   5.0f,   7.5f,   10.0f  } },
+            new TestQuantityPKParameter() { Name = "Name 1", QuantityPath = "Organism|PeripheralVenousBlood|Esomeprazole-2|Plasma (Peripheral Venous Blood)", _values = new[] { 0f,     0f,     0f,     0f,     0f } },
+            new TestQuantityPKParameter() { Name = "Name 2", QuantityPath = "Organism|PeripheralVenousBlood|Esomeprazole-2|Plasma (Peripheral Venous Blood)", _values = new[] { 0f,     0f,     0f,     0f,     0f } },
+            new TestQuantityPKParameter() { Name = "Name 3", QuantityPath = "Organism|PeripheralVenousBlood|Esomeprazole-2|Plasma (Peripheral Venous Blood)", _values = new[] { 0f,     0f,     0f,     0f,     0f } }
+         };
+
+         _pkAnalyses = (sut as TestEditTimeProfileAnalysisChartPresenter).AggregatePKAnalysis(_simulation, pkParameters, "Esomeprazole");
```
msevestre, line 347: you would not need to create this presenter with all those fakes if the logic was outside of presneetaiton lgoci

src/PKSim.UI/Views/Simulations/PopulationPKAnalysisView.cs
```diff
@@ -63,8 +80,8 @@ public override void InitializeResources()
          base.InitializeResources();
          btnExportToExcel.InitWithImage(ApplicationIcons.Excel, text: PKSimConstants.UI.ExportPKAnalysesToExcel);
          layoutItemExportToExcel.AdjustLargeButtonSize();
-         populationPKAnalysisItem.TextVisible = false;
          layoutControlItemGlobalPKAnalysis.TextVisible = false;
+         populationPKAnalysisXtraTabControl.SelectedPageChanged += (o, e) => OnEvent(_presenter.HandleTabChanged);
```
msevestre, line 84: Can we had icons to the tabs? One using the population icon and one using the timeProfile icon This would look a bit nicer. It might be enough to have the underling view return their corresponding icons depending on how the tabs are built

msevestre (changes requested):
All in all, very good. I think this has come a long way since the first implementation
I have a few suggestions still to make the code a bit more readable. Most suggestions are about renaming and running resharper however :)

Well done

src/PKSim.Core/Services/PKAnalysesTask.cs
abdelr: > does it need to be public at all?

It is being used outside the class.

src/PKSim.Presentation/Presenters/Charts/EditTimeProfileAnalysisChartPresenter.cs
abdelr: Right, already checked at `if (!populationDataCollectorSupportsDifferentAggregations(PopulationDataCollector))`

src/PKSim.Presentation/Presenters/Charts/EditTimeProfileAnalysisChartPresenter.cs
abdelr: > it feels that the values and index are only use to create the caption. Maybe a method that returns the caption can be created and the caption passe dto this function insteaD? Also this is not presentation related and could be moved

It is also used to match the results from the statistical aggregation to the original source. But I did refactor this a bit for better readability.

src/PKSim.Core/Services/PKAnalysesTask.cs
abdelr: > > does it need to be public at all?
> 
> It's being used outside the class.

Not anymore... it is now private ;)

msevestre:
looks awesome

PR #2266, opened 2022-07-29
Fixes #2251 also load metabolites

### LINKED ISSUE #2251: When loading a compound from github, "Also load metabolites?" is always shown
Even if the compound does not have any metabolite in the data base.

msevestre:


src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -1686,6 +1686,7 @@ public static class UI
          public static readonly string DefaultSolubilityName = "Solubility";
          public static readonly string DefaultPopulationAnalysisType = "Population analysis";
          public static readonly string PreferredViewLayout = "Preferred View Layout";
+         public static readonly string TemplateReferenceBehavior = "Load metabolite when loading compound";
```
Yuri05, line 1689: "Load metabolites... " (plural) is probably better.

msevestre: done

PR #2268, opened 2022-07-29
Fixes #2241 initial concentration value

### LINKED ISSUE #2241: Expression profile BB shows "0" for all initial concentrations
image

I think we discussed the values being "NaN" or something else, as they cannot be calculated without a linked individual. And the user can overwrite them with a fixed value.

Currently, all initial concentrations  values are shown as "0" which is misleading. In the Individuals-BB, the values are properly calculated:

image

msevestre:


src/PKSim.Presentation/DTO/Parameters/ParameterDTO.cs
```diff
@@ -81,11 +82,11 @@ public virtual double Value
...
             //We probably need to update the rules framework to not evaluate on NaN.
-            if (!success)
-               return 0;
+            if (success)
+               return result;
             
-            return result;
-
+            //the returned value will be based on the parameter. For Initial concentration, we will return NaN, otherwise 0
+            return Parameter.IsNamed(INITIAL_CONCENTRATION) ? result : 0;
```
msevestre, line 89: that simple... We nee dto resturn nan so that it is displayed as <enter a value>

Yuri05 (approved):

PR #2294, opened 2022-08-04
2248 empty sapces

msevestre:


msevestre:
Can you try and let me know. Works for me on 250% scaling

PR #2301, opened 2022-08-05
Plugging in simulation output mappings

### LINKED ISSUE #2292: Mapping of simulation outputs to observed data in simulations
issue created for the plug in of :

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1647

issue in Core

georgeDaskalakis:
closes #2292 

@msevestre this is just the same PR with the code review changes - as noted in the last PR just made sense to create a new PR to avoid having so many files in the diff.

### DIFF
--- a/src/PKSim.Core/Model/Simulation.cs
+++ b/src/PKSim.Core/Model/Simulation.cs
@@ -1,8 +1,6 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
-using OSPSuite.Utility.Collections;
-using OSPSuite.Utility.Extensions;
-using OSPSuite.Utility.Visitor;
 using OSPSuite.Core.Chart;
 using OSPSuite.Core.Diagram;
 using OSPSuite.Core.Domain;
@@ -10,7 +8,9 @@ using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Domain.Data;
 using OSPSuite.Core.Domain.ParameterIdentifications;
 using OSPSuite.Core.Domain.Services;
-using System;
+using OSPSuite.Utility.Collections;
+using OSPSuite.Utility.Extensions;
+using OSPSuite.Utility.Visitor;
 
 namespace PKSim.Core.Model
 {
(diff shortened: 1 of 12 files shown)


src/PKSim.Core/Model/Simulation.cs
```diff
@@ -303,6 +302,8 @@ public virtual void AddAnalyses(IEnumerable<ISimulationAnalysis> simulationAnaly
       /// </summary>
       public virtual IEnumerable<CurveChart> Charts => _allSimulationAnalyses.OfType<CurveChart>();
 
+      public OutputMappings OutputMappings { get; set; } = new OutputMappings();
```
msevestre, line 305: new is important here. That way we are compatible with older simulations

src/PKSim.Infrastructure/Serialization/ORM/MetaData/SimulationMetaData.cs
```diff
@@ -17,6 +18,7 @@ public class SimulationMetaData : BuildingBlockMetaData
       public virtual ICollection<UsedBuildingBlockMetaData> BuildingBlocks { get; set; }
       public virtual ICollection<string> UsedObservedData { get; set; }
       public virtual SimulationResults SimulationResults { get; set; }
+      public virtual OutputMappings OutputMappings { get; set; }
```
msevestre, line 21: Why do we need this?

src/PKSim.Infrastructure/Serialization/Xml/Serializers/SimulationXmlSerializer.cs
```diff
@@ -17,6 +17,7 @@ public override void PerformMapping()
          Map(x => x.ReactionDiagramModel);
          Map(x => x.Reactions);
          Map(x => x.SimulationSettings);
+         Map(x => x.OutputMappings);
```
msevestre, line 20: I think you could add a test to check serialization. This will probably fail. But at least we remember that we have to do sthg:)

georgeDaskalakis: I added the unit test, but yes indeed it fails. Let's discuss this a bit today.

georgeDaskalakis:
outdated, will be tackled in new PR

PR #2337, opened 2022-09-14
Fixes #2331 Rename tabs for pk analysis on population simulations

### LINKED ISSUE #2331: Should we rename "On curves" and "on individuals" to something like "Aggregated PK values" / "Individual PK values"?
Should we rename "On curves" and "on individuals" to something like "Aggregated PK values" / "Individual PK values"?

_Originally posted by @Yuri05 in https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2324#issuecomment-1241695896_

rwmcintosh:
Easing into the swing of things

### DIFF
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2632,9 +2632,9 @@ namespace PKSim.Assets
          public static readonly string Unit = UI.Unit;
          public static readonly string Description = "Description";
          public static readonly string Warning = "Warning";
-         public static readonly string OnCurves = "On Curves";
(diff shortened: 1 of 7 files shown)


msevestre:
since you have done more than renaming the caption but changing the variables (which I think make sense), there is some left

https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/e8d9821207ea31c86f0777ad36d3206947ba5f90/src/PKSim.UI/Views/Simulations/PopulationPKAnalysisView.cs#L18

msevestre (changes requested):
Also rename the pivotView defined in the view

PR #2340, opened 2022-09-15
Fixes #2324 PopSim: PK parameters not shown for all compounds

### LINKED ISSUE #2324: PopSim: PK parameters On Individuals are shown only for one compound, even if multiple are selected
I selected two compounds for the output, and the "On Curves"-tag shows pk parameter values for both compounds:

image

The "On Individuals" tab, however, shows only values for "Esomeprazole"

image

rwmcintosh:


src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
```diff
@@ -68,23 +69,26 @@ public void CalculatePKAnalyses(IPopulationDataCollector populationDataCollector
...
+         {
+            var pkAnalyses = _pkAnalysesTask.AggregatePKAnalysis(simulation, pkParameters.Value, populationAnalysis.SelectedStatistics, pkParameters.Key);
+            _allPKAnalysesIndividualPKValues.AddRange(pkAnalyses);
+         });
 
          updateView();
       }
 
-
-      private IReadOnlyList<QuantityPKParameter> extractPKParameters(PopulationStatisticalAnalysis populationAnalysis, PopulationSimulation populationSimulation)
```
rwmcintosh, line 84: This used to make a flat list with all the pkParameters combined. Now this makes a cache with a list per field.

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
```diff
@@ -68,23 +69,26 @@ public void CalculatePKAnalyses(IPopulationDataCollector populationDataCollector
...
+      private Cache<string, IReadOnlyList<QuantityPKParameter>> extractPKParameters(PopulationStatisticalAnalysis populationAnalysis, PopulationSimulation populationSimulation)
       {
-         var fields = populationAnalysis.AllFields.OfType<PopulationAnalysisOutputField>().Select(x => x.QuantityPath);
-         return fields.SelectMany(x => populationSimulation.PKAnalyses.AllPKParametersFor(x)).ToList();
+         var cache = new Cache<string, IReadOnlyList<QuantityPKParameter>>();
+         var fields = populationAnalysis.AllFields.OfType<PopulationAnalysisOutputField>().ToList();
```
msevestre, line 88: What about you call it output? Field is the generic term that means anything from parameters to pk etc..

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
```diff
@@ -68,23 +69,26 @@ public void CalculatePKAnalyses(IPopulationDataCollector populationDataCollector
...
+
+         fields.Each(field => { cache.Add(field.Name, populationSimulation.PKAnalyses.AllPKParametersFor(field.QuantityPath).ToList()); });
```
msevestre, line 90: Is name path? What is the name exactly? Is there a property path ? If this is the same, use it instead because name looks like it could be duplicated

msevestre: You do not need the { } after =>

Also why the ToList?is this required

msevestre: Or maybe is name unique. I can't remember

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
rwmcintosh: You mean outputFields or something like that. This isn't the output of the method.

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
rwmcintosh: I don't see it being obvious. Maybe make the Cache with the PopulationAnalysisOutputField instead and then use the field.Name to get the caption

rwmcintosh: The object itself should be unique.

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
msevestre: output is part of the "domain" of PKSim. output of the simulation. outputField is good too. I did not know what field was in the context at first glance

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
msevestre: is the name unique? Probably

msevestre:
Looks good to me. I am going to merge and try those changes

msevestre:
Well done. Not an easy contribution

PR #2342, opened 2022-09-15
We need to make sense of the AggregatePKAnalysis function

msevestre:


src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -299,6 +299,7 @@ public IReadOnlyList<PopulationPKAnalysis> AggregatePKAnalysis(Simulation simula
       {
          var pkParametersList = pkParameters.ToList();
          var matrix = new FloatMatrix();
+         //WHY THE DISTINCT HERE. THis array is then used later in buildPopulationPKAnalysis by Index. 
          var names = pkParametersList.Select(x => x.Name).Distinct().ToList();
```
msevestre, line 303: This is very weird here. In the function below (buildPopulationPKAnalysis), we are using the index of names and values where one is not created based of the other. Very weird

msevestre (commented):
I just want to get the ball rolling. I am not even sure the values are correctly calculated

rwmcintosh:
The Distinct is probably not required now. The previous implementation in the presenter who calls this made it necessary to remove all duplicate 'names' calculated by the Select. If you had two compounds with the same pk parameters, the names would have been all passed in at once duplicating the 'names'.

PR #2349, opened 2022-09-22
Fixes update core

rwmcintosh:


src/PKSim.Core/Model/Simulation.cs
```diff
@@ -625,7 +625,7 @@ public virtual Origin Origin
       }
 
       //This is not used in PKSim.
-      public IBuildConfiguration BuildConfiguration { get; } = null;
+      public IBuildConfiguration BuildConfiguration { get; set; } = null;
```
rwmcintosh, line 628: There is a code change in this PR

PR #2350, opened 2022-09-26
Fixes #2338 Aggregate population PK values with median instead of ari…

### LINKED ISSUE #2338: Population PK Values: aggregate with median instead of arithmetic mean
Please extend the caption with the type of the mean which is used for the calculation (_arithmetic_ or _geometric_)

grafik

rwmcintosh:
…thmetic mean

### DIFF
--- a/src/PKSim.Core/Model/PopulationSimulation.cs
+++ b/src/PKSim.Core/Model/PopulationSimulation.cs
@@ -8,6 +8,7 @@ using OSPSuite.Core.Domain.Services;
 using OSPSuite.Core.Extensions;
 using OSPSuite.Core.Maths.Random;
 using OSPSuite.Utility.Extensions;
+using PKSim.Core.Extensions;
 using PKSim.Core.Repositories;
 
 namespace PKSim.Core.Model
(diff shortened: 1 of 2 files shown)


src/PKSim.Core/Model/PopulationSimulation.cs
```diff
@@ -270,37 +271,41 @@ public override void UpdateFromOriginalSimulation(Simulation originalSimulation)
...
          if (column == null)
             return null;
 
-         return new DataColumn(column.Id, column.Dimension, column.BaseGrid)
+         return  new DataColumn(column.Id, column.Dimension, column.BaseGrid)
          {
-            Values = Enumerable.Range(0, column.Values.Count)
-               .Select(i => 
-                  columns.Select(x => x.Values[i]).ToList().ArithmeticMean()
+            Values = Enumerable.Range(0, column.Values.Count).Select(i =>
```
rwmcintosh, line 282: going through the data columns and taking the nth value in each case and then aggregating the median. There's no shortcut obvious to me. But the inside linq stuff can be hidden a bit by a method

PR #2354, opened 2022-09-29
update core to 11.1.22 and SimModel to 4.0.0.52

rwmcintosh:
updating and consolidating core and simmodel versions


msevestre:
close?

rwmcintosh:
Yeah, I don't want to merge it until SimModel is updated I guess.

PR #2359, opened 2022-10-05
Fixes #2260 - including "Global" PK-Parameters into the box-plot/scatter plot

### LINKED ISSUE #2260: including "Global" PK-Parameters into the box-plot/scatter plot
Derived from #2227

rwmcintosh:


src/PKSim.Presentation/Presenters/PopulationAnalyses/PopulationAnalysisPKParameterFieldsPresenter.cs
```diff
@@ -59,7 +59,7 @@ private PopulationAnalysisPKParameterField createFieldFrom(QuantityPKParameter p
       {
          var quantity = _allQuantities[pkParameter.QuantityPath];
          if (quantity == null)
-            return null;
+            return _populationAnalysisFieldFactory.CreateFor(pkParameter, QuantityType.Drug | QuantityType.Observer, quantityDisplayPath);
```
rwmcintosh, line 62: I'm not sure here. Clearly the quantity is not going to be found in the _allQuantities cache. But I'm not sure how get around this. Should there be additional values added int _allQuantities?

src/PKSim.Core/Services/GlobalPKAnalysisTask.cs
```diff
@@ -30,136 +32,187 @@ public interface IGlobalPKAnalysisTask
...
+
          _vssCalculator = vssCalculator;
          _interactionTask = interactionTask;
          _cloner = cloner;
+         _pkCalculationOptionsFactory = pkCalculationOptionsFactory;
+         _populationSimulationBodyWeightUpdater = populationSimulationBodyWeightUpdater;
       }
 
-      private GlobalPKAnalysis calculateGlobalPKAnalysisFor(Simulation simulation)
+      private IContainer calculateGlobalPKAnalysisFor(Simulation simulation, Compound compound, DataColumn peripheralVenousBloodCurve, DataColumn venousBloodCurve, PKCalculationOptions options)
```
rwmcintosh, line 70: This method is now called per compound so that it was easier to do the body weight scaling. The content should be the same otherwise.

tests/PKSim.Tests/Core/GlobalPKAnalysisTaskSpecs.cs
```diff
@@ -29,7 +30,7 @@ public abstract class concern_for_GlobalPKAnalysisTask : ContextSpecification<IG
       private IVSSCalculator _vssCalculator;
       protected DataColumn _peripheralVenousBloodPlasma;
       protected DataColumn _venousBloodPlasma;
-      protected IndividualSimulation _simulation;
```
rwmcintosh, line 32: I refactored this test a bit because the base abstract class was setting up everything for an individual simulation, then the population_based was subclassing that and adding some population simulations.  When I wrote my tests, I couldn't figure out why they were failing. The assertions were using the wrong simulation!

src/PKSim.Core/Model/PopulationSimulation.cs
```diff
@@ -280,27 +279,37 @@ private DataColumn medianAggregateDataColumns(IReadOnlyList<DataColumn> columns)
...
       /// <summary>
       ///    tries to find venous blood plasma if defined in the repository. returns null otherwise
       /// </summary>
       public override DataColumn VenousBloodColumn(string compoundName)
       {
-         return medianAggregateDataColumns(drugColumnFor(CoreConstants.Organ.VENOUS_BLOOD, CoreConstants.Compartment.PLASMA, CoreConstants.Observer.CONCENTRATION_IN_CONTAINER, compoundName));
+         return medianAggregateDataColumns(VenousBloodColumns(compoundName));
+      }
+
+      public IReadOnlyList<DataColumn> VenousBloodColumns(string compoundName)
```
msevestre, line 310: why is this public?

src/PKSim.Core/Model/PopulationSimulation.cs
```diff
@@ -280,27 +279,37 @@ private DataColumn medianAggregateDataColumns(IReadOnlyList<DataColumn> columns)
...
          return columns.Select(x => x.Values[i]).ToList();
       }
 
       public override DataColumn PeripheralVenousBloodColumn(string compoundName)
       {
-         return medianAggregateDataColumns(drugColumnFor(CoreConstants.Organ.PERIPHERAL_VENOUS_BLOOD, CoreConstants.Observer.PLASMA_PERIPHERAL_VENOUS_BLOOD, CoreConstants.Observer.PLASMA_PERIPHERAL_VENOUS_BLOOD, compoundName));
+         return medianAggregateDataColumns(PeripheralVenousBloodColumns(compoundName));
+      }
+
+      public IReadOnlyList<DataColumn> PeripheralVenousBloodColumns(string compoundName)
```
msevestre, line 297: why is this public?

tests/PKSim.Tests/Presentation/SimulationComparisonTreeNodeContextMenuFactorySpecs.cs
```diff
@@ -20,7 +20,7 @@ protected override void Context()
    public class When_checking_if_the_simulation_comparison_tree_node_context_menu_factory_can_create_a_context_menu_for_a_given_node : concern_for_SimulationComparisonTreeNodeContextMenuFactory
    {
       [Observation]
-      public void shouldu_return_true_if_the_node_is_a_classifiable_comparision_node()
+      public void should_return_true_if_the_node_is_a_classifiable_comparision_node()
```
msevestre, line 23: still a typo lol comparision => comaprison

src/PKSim.Core/Model/PopulationSimulation.cs
rwmcintosh: GlobalPKAnalysisTask needs to get access to all the columns for PeripheralVenousBlood and VenousBlood

tests/PKSim.Tests/Presentation/SimulationComparisonTreeNodeContextMenuFactorySpecs.cs
rwmcintosh: There are a few instances of that. I'll fix them all, but one is in a database query that I won't change.

```
migrateTo5_3
```

src/PKSim.Core/Model/PopulationSimulation.cs
```diff
@@ -280,35 +279,43 @@ private DataColumn medianAggregateDataColumns(IReadOnlyList<DataColumn> columns)
...
       public override DataColumn FabsOral(string compoundName)
       {
-         return medianAggregateDataColumns(drugColumnFor(CoreConstants.Organ.LUMEN, CoreConstants.Observer.FABS_ORAL, CoreConstants.Observer.FABS_ORAL, compoundName));
+         return medianAggregateDataColumns(drugColumnsFor(CoreConstants.Organ.LUMEN, CoreConstants.Observer.FABS_ORAL, CoreConstants.Observer.FABS_ORAL, compoundName));
       }
 
-      private IReadOnlyList<DataColumn> drugColumnFor(string organ, string compartment, string columnName, string compoundName)
+      private DataColumn drugColumnForIndividual(string organ, string compartment, string columnName, string compoundName, int individualId)
+      {
+         return columnsFor(Results.IndividualResultsAsArray().FirstOrDefault(result => result.IndividualId == individualId), organ, compartment, columnName, compoundName);
```
msevestre, line 315: There has to be away to access results by Id other than this? This looks very odd to me/.Plus the array conversion etc ..

src/PKSim.Core/Model/PopulationSimulation.cs
```diff
@@ -330,5 +337,15 @@ private DataColumn columnsFor(IndividualResults results, string organ, string co
             Values =  column.Values
          };
       }
+
+      public DataColumn VenousBloodColumnForIndividual(int individualId, string compoundName)
```
msevestre, line 341: do they still need to be public?

msevestre (commented):
on going review. One big comment already

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -95,18 +125,389 @@ public PopulationSimulationPKAnalyses CalculateFor(PopulationSimulation populati
...
+               {
+                  _populationSimulationBodyWeightUpdater.UpdateBodyWeightForIndividual(populationSimulation, individualId);
+               },
+               (individualId, options, compoundName) =>
+               {
+                  var globalPKAnalysis = globalPkAnalysisForIndividual(globalPKAnalysesForIndividuals, individualId);
+                  globalPKAnalysis.Add(calculateGlobalPKParameters(populationSimulation, individualId, compoundName, options, globalPKAnalysis));
+               });
+
+            mapQuantityPKParametersFromIndividualGlobalPKAnalyses(globalPKAnalysesForIndividuals.ToList()).Each(x => analyses.AddPKAnalysis(x));
```
msevestre, line 142: why the ToList. It's already an enumerable
globalPKAnalysesForIndividuals.Each(analyses.AddPKAnalysis)

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -95,18 +125,389 @@ public PopulationSimulationPKAnalyses CalculateFor(PopulationSimulation populati
...
+            });
+         });
+
+         return quantityPKList;
+      }
+
+      private static QuantityPKParameter quantityPKParameterFor(IReadOnlyList<GlobalPKAnalysis> globalIndividualPKParameterList, IParameter pKParameter, string quantityPath)
+      {
+         var pKValuesForPKParameter = globalIndividualPKParameterList.SelectMany(globalPKAnalysisForAnIndividual => globalPKAnalysisForAnIndividual.AllPKParameters.Where(x => pathsEqual(x, pKParameter))).ToList();
+         var quantityPKParameter = new QuantityPKParameter { Dimension = pKValuesForPKParameter.First().Dimension, Name = pKValuesForPKParameter.First().Name, QuantityPath = quantityPath };
```
msevestre, line 311: are we sure here that there will be at least one? (first)

src/PKSim.Core/Services/PopulationSimulationBodyWeightUpdater.cs
```diff
@@ -0,0 +1,44 @@
...
+   public class PopulationSimulationBodyWeightUpdater : IPopulationSimulationBodyWeightUpdater
+   {
+      private readonly IEntityPathResolver _entityPathResolver;
+
+      public PopulationSimulationBodyWeightUpdater(IEntityPathResolver entityPathResolver)
+      {
+         _entityPathResolver = entityPathResolver;
+      }
+
+      public void UpdateBodyWeightForIndividual(PopulationSimulation populationSimulation, int individualId)
```
msevestre, line 22: Is there a reason why this was moved out of the PK Stuff? Because you are doing line 24 to 26 now for EAch individual as opposed to one time.

src/PKSim.Presentation/Presenters/PopulationAnalyses/PopulationAnalysisAvailablePKParametersPresenter.cs
```diff
@@ -69,6 +69,16 @@ private void createAvailableQuantitiesListBasedOn(IPopulationDataCollector popul
       {
          _allPKParameters.Clear();
          _allPKParameters.AddRange(_entitiesInContainerRetriever.OutputsFrom(populationDataCollector).SelectMany(availablePKParametersFor));
+
+         populationDataCollector.Compounds.Select(x => x.Name).Each(compoundName =>
```
msevestre, line 73: Extension method AllNames

src/PKSim.Presentation/Presenters/PopulationAnalyses/PopulationAnalysisPKParameterFieldsPresenter.cs
msevestre: I think this is ok but this requires a comment explaining that this parameter is not found because it is a global PK Parameter (can we actually check if this is one?). It cil;d also be the the quantity was removed somehow (thinking about mobi models imported etc..)

msevestre (changes requested):
Looking very good. Just a few minor comments

src/PKSim.Core/Model/PopulationSimulation.cs
rwmcintosh: We are using that from PKAnalysesTask, or what do you mean? Do we need this to be a datacolumn at all? And can we work from the results directly?

src/PKSim.Core/Services/PopulationSimulationBodyWeightUpdater.cs
rwmcintosh: The only reason for it was that originally  I was reproducing the scaling logic in a second location. I think it makes sense to put it back now.

src/PKSim.Core/Model/PopulationSimulation.cs
rwmcintosh: `Results.ResultsFor(individualId)` :) Nice.

src/PKSim.Core/Model/PopulationSimulation.cs
msevestre: just wondering if we need tthis sutff outside of the class. But we do.  I wasn't sure

PR #2369, opened 2022-10-26
2358 calculation of global pk parameters

rwmcintosh:


src/PKSim.Core/Model/CompoundPK.cs
```diff
@@ -1,37 +1,89 @@
...
+using OSPSuite.Core.Domain;
+using OSPSuite.Utility.Collections;
 
 namespace PKSim.Core.Model
 {
    /// <summary>
-   /// Reperesents a cache containing specific PK-Values calculated for a given compound
+   /// Represents a cache containing specific PK-Values calculated for a given compound
    /// </summary>
    public class CompoundPK
```
rwmcintosh, line 10: This object is the basis for the context. We are collecting denominotors for ratio parameters as well as QuantityPKParameters where necessary. Sometimes the context is that we have the parameter already calculated, sometimes we have only the denominator

src/PKSim.Core/Model/PopulationSimulation.cs
```diff
@@ -323,10 +323,7 @@ private IReadOnlyList<DataColumn> drugColumnsFor(string organ, string compartmen
       private DataColumn columnsFor(IndividualResults results, string organ, string compartment, string columnName, string compoundName)
       {
          var column = results.FirstOrDefault(x =>
-               x.QuantityPath.Contains(organ) &&
-               x.QuantityPath.Contains(compartment) &&
-               x.QuantityPath.Contains(columnName) &&
-               x.QuantityPath.Contains(compoundName)
+               isQuantityPathFor(organ, compartment, columnName, compoundName, x)
```
rwmcintosh, line 326: This is that bug where if one compound contains the other as substring, you get problems

src/PKSim.Core/Services/GlobalPKAnalysisRunner.cs
```diff
@@ -48,10 +48,11 @@ public Simulation RunForDDIRatio(Simulation simulation)
       private Simulation createAndRun(Func<ISimulationFactory, Simulation> simulationCreator)
       {
          var simulation = simulationCreator(_simulationFactory);
+
          try
          {
             _registrationTask.Register(simulation);
-            _simulationRunner.RunSimulation(simulation).Wait();
+            _simulationRunner.RunSimulation(simulation, new SimulationRunOptions { RaiseEvents = false }).Wait();
```
rwmcintosh, line 55: Prevent raising events for the transient simulations

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -177,24 +178,78 @@ public GlobalPKAnalysis CalculateGlobalPKAnalysisFor(IEnumerable<Simulation> sim
...
          {
-            var pkCalculationOptions = _pkCalculationOptionsFactory.CreateFor(simulation, compound.Name);
-            var peripheralVenousBloodColumn = simulation.PeripheralVenousBloodColumn(compound.Name);
-            var venousBloodColumn = simulation.VenousBloodColumn(compound.Name);
-            var compoundContainer = calculateGlobalPKAnalysisFor(simulation, compound, peripheralVenousBloodColumn, venousBloodColumn, pkCalculationOptions);
-            globalPKAnalysis.Add(compoundContainer);
-         });
+            // Global PK Parameters are stored in containers with the compound name as the first child of the root container.
+            // There will be one container per compound with PK parameters stored within
+            var compoundContainers = createCompoundPKParameterContainers(populationSimulation);
```
rwmcintosh, line 187: For population simulations, calculating the GlobalPKAnalysis is just retrieving the quantity parameters from the populationSimulation.PKAnalyses.

src/PKSim.Core/Model/CompoundPK.cs
```diff
@@ -1,37 +1,89 @@
...
+      public void AddQuantityPKParameter(QuantityPKParameter quantityPKParameter)
+      {
+         _quantityPKParameters.Add(quantityPKParameter);
+      }
+
+      public ICache<string, QuantityPKParameter> QuantityPKParameters() => _quantityPKParameters;
+
+      public double? QuantityFor(int individualId, string parameterName)
+      {
+         var quantityParameter = _quantityPKParameters.FirstOrDefault(x => string.Equals(x.Name, parameterName));
```
msevestre, line 84: Why are you using firstOrDefault and not the fact that this is a cache? (If this is the serialization issue, verify if it contains and then use)

src/PKSim.Core/Model/CompoundPK.cs
```diff
@@ -1,37 +1,89 @@
...
    {
+      private readonly Cache<string, QuantityPKParameter> _quantityPKParameters = new Cache<string, QuantityPKParameter>(getKey:x => x.Name);
+
+      public CompoundPK()
+      {
+         bioAvailabilities = new Cache<int, double?>();
+         ddiAucInf = new Cache<int, double?>();
+         ddiCMax = new Cache<int, double?>();
+      }
       public string CompoundName { get; set; }
```
msevestre, line 20: Move all the private stuff and properties above the constructor. This is my expectations when I look at a class to have the members and then the constructor. Here it's a bit backward

src/PKSim.Core/Model/CompoundPK.cs
```diff
@@ -1,37 +1,89 @@
...
-      public double? CmaxDDI { get; set; }
+      private double? valueStoreFor(ICache<int, double?> allValues, int individualId)
+      {
+         if (!allValues.Contains(individualId))
+            allValues[individualId] = null;
+
+         return allValues[individualId];
+      }
+
+      public ICache<int, double?> AllBioAvailabilities => bioAvailabilities;
```
msevestre, line 41: you can probably crate and init in one go here .This is probably public for serializatiohn

src/PKSim.Core/Model/CompoundPK.cs
```diff
@@ -1,37 +1,89 @@
...
+      }
       public string CompoundName { get; set; }
-      
+
       /// <summary>
-      /// AUC_inf of plasma curve following a IV infustion of 15 min
+      /// AUC_inf of plasma curve following a IV infusion of 15 min
       /// </summary>
-      public double? AucIV { get; set; }
+      private Cache<int, double?> bioAvailabilities { get; set; }
```
msevestre, line 25: private=>naming is off. Create a _privateMember (not get set)

src/PKSim.Core/Model/CompoundPK.cs
```diff
@@ -1,37 +1,89 @@
...
+      public ICache<int, double?> AllBioAvailabilities => bioAvailabilities;
+      public ICache<int, double?> AllDDIAucInf => ddiAucInf;
+      public ICache<int, double?> AllDDICMax => ddiCMax;
+
+      public double? BioAvailabilityAucInfFor(int individualId)
+      {
+         return valueStoreFor(bioAvailabilities, individualId);
+      }
+
+      public double? DDIAucInfFor(int individualId)
```
msevestre, line 50: you can makle those a one line,. Small stuff but resharper can do it fo you and it's less verbose

```
   public double? DDIAucInfFor(int individualId)
      {
         return valueStoreFor(ddiAucInf, individualId);
      }
```
==>


```   
public double? DDIAucInfFor(int individualId) => valueStoreFor(ddiAucInf, individualId);
```

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -297,6 +350,46 @@ private IContainer calculateGlobalPKAnalysisFor(Simulation simulation, Compound
          return container;
       }
 
+      private IParameter cmaxRatioCalculator(int individualId, CompoundPK compoundPKFromContext, float? cMax)
+      {
+         var contextValue = compoundPKFromContext.QuantityFor(individualId, CoreConstants.PKAnalysis.C_maxRatio);
+         if (contextValue.HasValue)
```
rwmcintosh, line 356: There's some repetition here that can maybe get factored out in these three functions. The idea is that you first check if there's a QuantityPKParameter for the individual already in the context and if so, just use this value.

Otherwise you check if there's a denominator available to calculate the ratio parameter, otherwise it's just null

src/PKSim.Core/Model/IndividualSimulation.cs
```diff
@@ -78,5 +71,23 @@ public override DataColumn PeripheralVenousBloodColumn(string compoundName)
...
+      }
+
+      public IndividualSimulation()
+      {
+         AucDDI = new Cache<string, double?>(onMissingKey: x => null);
+         AucIV = new Cache<string, double?>(onMissingKey: x => null);
+         CMaxDDI = new Cache<string, double?>(onMissingKey: x => null);
+      }
+
+      public Cache<string, double?> AucDDI { get; }
```
msevestre, line 89: Initalize when you declare so that you don't nee the constructor

 public Cache<string, double?> AucDDI { get; } = new Cache<string, double?>(onMissingKey: x => null);

msevestre: but moves those guys up. Unexpected to see member declared 1/2 way through the class

src/PKSim.Core/Model/PopulationSimulation.cs
msevestre: good find

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -430,23 +523,97 @@ public void CalculateBioavailabilityFor(Simulation simulation, string compoundNa
...
+            }
+         });
+         simulation.HasChanged = true;
+
+      }
+
+      private CompoundPKContext createContextForPopulationSimulation(CompoundPKContext contextCompoundPK, Func<CompoundPK, CompoundPK> compoundPKMapper, string compoundName, PopulationSimulation populationSimulation, string[] parameterNames)
+      {
+         var contextForPopulationSimulation = new CompoundPKContext();
+         contextForPopulationSimulation.AddCompoundPK(compoundPKMapper(contextCompoundPK.CompoundPKFor(compoundName)));
```
rwmcintosh, line 598: This adds 'denominators' to the context for calculation of ratio parameters.

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -430,23 +523,97 @@ public void CalculateBioavailabilityFor(Simulation simulation, string compoundNa
...
+         contextForPopulationSimulation.InitializeQuantityPKParametersFrom(populationSimulation);
```
rwmcintosh, line 599: This adds fully calculated ratio parameters from the population simulation

src/PKSim.Core/Services/GlobalPKAnalysisRunner.cs
msevestre, line 55: add a comment explaining why we set it to false here

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -430,23 +523,97 @@ public void CalculateBioavailabilityFor(Simulation simulation, string compoundNa
...
+         parameterNames.Each(parameterName => { removeQuantityPKParameterFromCompound(compoundName, contextForPopulationSimulation, parameterName); });
```
rwmcintosh, line 600: Then after that, we have to remove the ratio parameters that are being calculated

src/PKSim.Core/Services/PopulationSimulationEngine.cs
```diff
@@ -12,12 +12,28 @@
 
 namespace PKSim.Core.Services
 {
-   public class PopulationSimulationEngine : IPopulationSimulationEngine
+   public class BaseSimulationEngine
```
msevestre, line 15: make it abtract

msevestre: also probably SImulationEngine in this case?

rwmcintosh:
You can start, but the tests are all broken and won't compile for now so I'd stop there.

tests/PKSim.Tests/Core/IndividualSimulationEngineSpecs.cs
```diff
@@ -56,7 +56,7 @@ protected override async Task Context()
          _simulation = A.Fake<IndividualSimulation>();
          _simulation.Name = "Hello";
          _simulation.DataRepository = new DataRepository();
-         _simulation.CompoundPKFor("TOTO").AucIV = 55;
+         _simulation.CompoundPKFor("TOTO").BioAvailabilityFor(0).AucIV = 55;
```
msevestre, line 59: I don't understand why we pass the 0 here. 
I would expect 
 _simulation.CompoundPKFor("TOTO").BioAvailability = 55;

msevestre (changes requested):
just starting here. I

src/PKSim.Core/Model/CompoundPK.cs
rwmcintosh: no, that's because it was a list, then I changed it to Cache and didn't clean that up.

src/PKSim.Core/Model/CompoundPK.cs
rwmcintosh: I think I don't need these separate private ones at all

tests/PKSim.Tests/Core/IndividualSimulationEngineSpecs.cs
rwmcintosh: These tests need attention. They don't work so there will be another commit coming.
In general though, for the individual simulations we don't need to keep track of the actual BioAvailability parameter. We just need to hold on to AucIV because we always calculate AucInf when we open the simulation editor

rwmcintosh: oh, and 0 was for the 'IndividualId'. This is just another case of changing everything and then cleaning up after.

src/PKSim.Core/Services/CompoundPKContext.cs
```diff
@@ -0,0 +1,55 @@
...
+      {
+         if (!_compoundPKCache.Contains(compoundName))
+            _compoundPKCache.Add(new CompoundPK { CompoundName = compoundName });
+
+         return _compoundPKCache[compoundName];
+      }
+
+      public void AddCompoundPK(CompoundPK compoundPK)
+      {
+         _compoundPKCache[compoundPK.CompoundName] = compoundPK;
```
msevestre, line 26: hum... why are you passing the key here? Can it exist already? Otherwise do _compoundPKCache.add(

msevestre (commented):
I need to look at the PKAnalysisTask in more details. The rest looks good

src/PKSim.Core/Model/IndividualSimulation.cs
```diff
@@ -78,5 +73,12 @@ public override DataColumn PeripheralVenousBloodColumn(string compoundName)
       public override DataColumn VenousBloodColumn(string compoundName) => DataRepository.VenousBloodColumn(compoundName);
 
       public override DataColumn FabsOral(string compoundName) => DataRepository.FabsOral(compoundName);
+
+      public void ClearRatioPKCache()
```
msevestre, line 77: Why is this call ratio?

src/PKSim.Core/Services/CompoundPKContext.cs
```diff
@@ -0,0 +1,55 @@
...
+      }
+
+      /// <summary>
+      /// Initializes the context compound PK ratio parameters from the calculated simulation PK Parameters
+      /// Adds BioAvailability, AUCRatio, and CMaxRatio to the context
+      /// </summary>
+      /// <param name="populationSimulation">The simulation that contains the PK Parameters</param>
+      public void InitializeQuantityPKParametersFrom(PopulationSimulation populationSimulation)
+      {
+         populationSimulation.CompoundNames.Each(compoundName =>
```
msevestre, line 36: There is a lot of "look the same"kid of code"
What about sthg like this (
```
           var pkAnalyses = populationSimulation.PKAnalyses;
           var compoundPK = CompoundPKFor(compoundName);
            var parameters =  [
              pkAnalyses.PKParameterFor(compoundName, Bioavailability);
               pkAnalyses.PKParameterFor(compoundName,AUCRatio);
               pkAnalyses.PKParameterFor(compoundName, C_maxRatio);
           ].Where(x=>x!null);
           parameters.Each(compoundPK.AddQuantityPKParameter)


```

that way if you add one more, this is just a one line to add here
even more to

```
           var pkAnalyses = populationSimulation.PKAnalyses;
           var compoundPK = CompoundPKFor(compoundName);
            var parameters =  new [Bioavailability, AUCRatio, C_maxRatio]
.select(x=>pkAnalyses.PKParameterFor(compoundName, x)).where(x=>x!-null);
           parameters.Each(compoundPK.AddQuantityPKParameter)


```

src/PKSim.Infrastructure/Serialization/Xml/Serializers/SimulationXmlSerializer.cs
```diff
@@ -57,7 +57,9 @@ public class IndividualSimulationXmlSerializer : SimulationXmlSerializer<Individ
       public override void PerformMapping()
       {
          base.PerformMapping();
-         MapEnumerable(x => x.AllCompoundPK, x => x.AddCompoundPK);
+         Map(x => x.AucIV);
```
msevestre, line 60: This would require project conversions
@Yuri05 I am ok to not do this at this stage. PEople would just have to rerun for individual simulations

src/PKSim.Presentation/Presenters/Simulations/GlobalPKAnalysisPresenter.cs
```diff
@@ -93,7 +93,7 @@ public void CalculateBioAvailability(string compoundName)
 
       public void CalculateDDIRatioFor(string compoundName)
       {
-         calculateGlobalPKAnalysis(x => x.CalculateDDIRatioFor(firstSimulation, compoundName));
+         calculateGlobalPKAnalysis(x => x.CalculateDDIRatioFor(firstSimulation));
```
msevestre, line 96: not convince about removing the name here. This is not used internally in the CalculateDDIRatioFor method, but this feels that the presenter should not know about this. Otherwise it's not really symetrical

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -177,24 +178,78 @@ public GlobalPKAnalysis CalculateGlobalPKAnalysisFor(IEnumerable<Simulation> sim
...
-            var pkCalculationOptions = _pkCalculationOptionsFactory.CreateFor(simulation, compound.Name);
-            var peripheralVenousBloodColumn = simulation.PeripheralVenousBloodColumn(compound.Name);
-            var venousBloodColumn = simulation.VenousBloodColumn(compound.Name);
-            var compoundContainer = calculateGlobalPKAnalysisFor(simulation, compound, peripheralVenousBloodColumn, venousBloodColumn, pkCalculationOptions);
-            globalPKAnalysis.Add(compoundContainer);
-         });
+            // Global PK Parameters are stored in containers with the compound name as the first child of the root container.
+            // There will be one container per compound with PK parameters stored within
+            var compoundContainers = createCompoundPKParameterContainers(populationSimulation);
+            compoundContainers.Each(globalPKAnalysis.Add);
```
msevestre, line 188: we probably have a AddChildren method in the container?

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -177,24 +178,78 @@ public GlobalPKAnalysis CalculateGlobalPKAnalysisFor(IEnumerable<Simulation> sim
...
+
+         // For individual simulations global pk parameters are calculated on the fly
+         if (simulation is IndividualSimulation individualSimulation)
+         {
+            individualSimulation.Compounds.Each(compound =>
+            {
+               var pkCalculationOptions = _pkCalculationOptionsFactory.CreateFor(individualSimulation, compound.Name);
+               var peripheralVenousBloodColumn = individualSimulation.PeripheralVenousBloodColumn(compound.Name);
+               var venousBloodColumn = individualSimulation.VenousBloodColumn(compound.Name);
+               var individualId = individualSimulation.Results.AllIndividualIds().FirstOrDefault();
```
msevestre, line 201: I don't like this. Can we create a method in the indiviudalSimulation? This is an ecampsulation details that is leaking here.

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -297,6 +350,44 @@ private IContainer calculateGlobalPKAnalysisFor(Simulation simulation, Compound
...
+      {
+         var parameter = contextParameterFrom(CoreConstants.PKAnalysis.AUCRatio, compoundPKFromContext.QuantityFor(individualId, CoreConstants.PKAnalysis.AUCRatio),
+            aucInf, compoundPKFromContext.DDIAucInfFor(individualId));
+
+         compoundPKFromContext.AddDDIAucInf(individualId, aucInf);
+
+         return parameter;
+      }
+
+      private IParameter bioAvailabilityCalculator(int individualId, CompoundPK compoundPKFromContext, double? aucInf)
```
msevestre, line 372: wondering about double here and why not float

msevestre: all those calculators are looking very much the same. Can we extract one one method?

msevestre: import static in this file. It will be so much easier to read

msevestre (approved):

src/PKSim.Presentation/Presenters/Simulations/GlobalPKAnalysisPresenter.cs
rwmcintosh: Trade undesirable asymmetry for unused parameters in method signatures. 

My preference is to leave it because the caller should not be under the impression that the compoundName has any effect on what's happening.
If we need to add it back later for feature enhancement it's not too much effort.

src/PKSim.Core/Services/PKAnalysesTask.cs
rwmcintosh: Do you mean ask the IndividualSimulation for it's IndividualId rather than having everyone know that the IndividualSimulation has an Enumerable of which there is only 1?

msevestre: yes. Suddently we know that the simulation as this REsults that look like a population etc..

src/PKSim.Core/Services/PKAnalysesTask.cs
rwmcintosh: > all those calculators are looking very much the same. Can we extract one one method?

That's just removing these functions. They all call the same method to create the parameter

src/PKSim.Presentation/Presenters/Simulations/GlobalPKAnalysisPresenter.cs
msevestre: Then I 'd argue ` public void CalculateDDIRatioFor(string compoundName)` 
why compoudName? 
But that's cool with me. REsolving this conversation

src/PKSim.Presentation/Presenters/Simulations/PopulationPKAnalysisPresenter.cs
```diff
@@ -70,7 +70,7 @@ public void CalculatePKAnalyses(IPopulationDataCollector populationDataCollector
          _globalPKAnalysisPresenter.CalculatePKAnalysis(new[] { simulation });
 
          var pkParametersCache = extractPKParameters(populationAnalysis, simulation);
-         if (!pkParametersCache.Any())
+         if (!pkParametersCache.Any() || pkParametersCache.All(x => !x.Any()))
```
rwmcintosh, line 73: That's the bug. The intent here is to make sure to stop if there's no parameters in the cache, but what's actually created is parameters without any values.

rwmcintosh: It's a cache of QuantityPKParameters. So they are parameters that are empty of results.

Issue #2275, opened 2022-08-02
UI: Expression-BB - make the "Localization" filed a bit higher so it does not require scrolling


PavelBal:
A couple of pixels more and the scroll bar is not required.

4k, 200%.

![image](https://user-images.githubusercontent.com/25903040/182317730-34362d92-55bf-405d-8eff-74b295823314.png)

PR #2375, opened 2022-10-28
Update dimensions, dlls, templates, pk parameters

msevestre:


src/PKSim.UI/UIConstants.cs
```diff
@@ -33,7 +33,7 @@ public static class Size
          public static readonly int PROTOCOL_VIEW_WIDTH = ScaleForScreenDPI(700);
          public static readonly int PROTOCOL_VIEW_HEIGHT = ScaleForScreenDPI(760);
 
-         public static readonly int EXPRESSION_PROFILE_LOCALIZATION_HEIGHT = ScaleForScreenDPI(80);
+         public static readonly int EXPRESSION_PROFILE_LOCALIZATION_HEIGHT = ScaleForScreenDPI(85);
```
msevestre, line 36: @PavelBal This is for your localization view issue. Can you try and let me know if this fixes #2275

msevestre:
@Yuri05 Dimensions, pkParameters, templates updated as well as latest core

PR #2413, opened 2022-11-07
Fixes #2410 some calculation in global pk wrong

### LINKED ISSUE #2410: 11.1 Global PK-Analysis in population simulation wrong?
[Pop_global_PK_Fluconazole-Midazolam-DDI.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/9944375/Pop_global_PK_Fluconazole-Midazolam-DDI.zip)
* I created a population with 10 individuals, then extracted all Individuals from the population
* Then created a simulation with the population and cloned this simulation 10 times, each time replacing the population by one of the extracted individuals
* Then calculated and exported global pk analyis values from each individual simulation to excel
* Finally, calculated the median value of each global pk-parameter in excel

Expectation: value calculated in Excel is the same as the global pk analysis value shown in the population simulation in PK-Sim
Reality: the values do not match

[global_pk_analysis_Ind0_Ind9_Ahonen_midazolam_po_7.5mg_fluconazole_400mg_i.v_DDI.xlsx](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/9944394/global_pk_analysis_Ind0_Ind9_Ahonen_midazolam_po_7.5mg_fluconazole_400mg_i.v_DDI.xlsx)

That's the result of the calculation in Excel:

Parameter | Compound | Unit | **MEDIAN** | Value | Value | Value | Value | Value | Value | Value | Value | Value | Value
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
AUC Ratio (AUCR) | Midazolam |   | **4,50** | 5,10 | 3,12 | 3,94 | 4,24 | 4,65 | 6,62 | 6,69 | 4,34 | 6,53 | 4,24
C_max Ratio (C_maxR) | Midazolam |   | **2,26** | 2,49 | 1,59 | 2,17 | 2,21 | 2,30 | 2,53 | 2,92 | 2,04 | 2,81 | 2,02
Vss (plasma) | Midazolam | ml/kg | **1321,10** | 1135,20 | 1089,44 | 1757,32 | 1167,40 | 1358,00 | 1300,75 | 1297,42 | 1341,46 | 1503,11 | 1527,08
Vss (plasma) | Fluconazole | ml/kg | **686,23** | 741,88 | 705,22 | 651,63 | 717,76 | 748,16 | 667,25 | 719,31 | 584,39 | 654,82 | 650,28
Vd (plasma) | Midazolam | ml/kg | **1128,81** | 934,25 | 1040,85 | 1691,81 | 987,52 | 1210,17 | 1125,58 | 1132,03 | 1123,41 | 1317,98 | 1325,13
Vd (plasma) | Fluconazole | ml/kg | **688,78** | 748,79 | 707,61 | 653,79 | 722,12 | 753,07 | 669,95 | 722,63 | 586,55 | 657,84 | 654,98
Total plasma clearance | Midazolam | ml/min/kg | **2,59** | 2,94 | 1,21 | 2,39 | 2,58 | 3,51 | 2,46 | 3,02 | 2,48 | 2,60 | 2,80
Total plasma clearance | Fluconazole | ml/min/kg | **0,20** | 0,38 | 0,13 | 0,13 | 0,24 | 0,26 | 0,16 | 0,19 | 0,14 | 0,21 | 0,30
Bioavailability | Midazolam |   | **0,63** | 0,75 | 0,73 | 0,59 | 0,63 | 0,53 | 0,60 | 0,56 | 0,74 | 0,63 | 0,67
Fraction absorbed | Midazolam |   | **1,00** | 1,00 | 1,00 | 1,00 | 1,00 | 1,00 | 1,00 | 1,00 | 1,00 | 1,00 | 1,00
Vss (phys-chem) | Fluconazole | ml/kg | **661,67** | 715,33 | 682,84 | 613,56 | 685,68 | 721,07 | 640,51 | 691,27 | 542,24 | 619,36 | 616,15

That's what calculated in the population simulation in PK-Sim:
grafik

rwmcintosh:


src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -245,7 +247,7 @@ private Container createCompoundContainer(PopulationSimulation populationSimulat
          var compoundContainer = new Container().WithName(compoundName);
          populationSimulation.PKAnalyses.AllPKParametersFor(compoundName).Each(quantityPKParameter =>
          {
-            double? defaultValue = quantityPKParameter.ValuesAsArray.Median();
```
rwmcintosh, line 248: That was most of the bug. This extension method expected the array to be sorted. But when you write the code like that it looks like it should work.

I added this more specific object to use for quantile type calculations on float arrays just to make it more obvious what the requirements are

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -260,8 +262,9 @@ private IContainer calculateGlobalPKParameters(PopulationSimulation simulation,
       {
          var peripheralVenousBloodColumn = simulation.PeripheralVenousBloodColumnForIndividual(individualId, moleculeName);
          var venousBloodColumn = simulation.VenousBloodColumnForIndividual(individualId, moleculeName);
+         var fractionAbsorbedColumn = simulation.FractionAbsorbedColumnForIndividual(individualId, moleculeName);
 
-         return calculateGlobalPKAnalysisFor(simulation, simulation.Compounds.FindByName(moleculeName), peripheralVenousBloodColumn, venousBloodColumn, calculationOptions, individualId, compoundPKContext);
+         return calculateGlobalPKAnalysisFor(simulation, simulation.Compounds.FindByName(moleculeName), peripheralVenousBloodColumn, venousBloodColumn, fractionAbsorbedColumn, calculationOptions, individualId, compoundPKContext);
```
rwmcintosh, line 267: There was also a bug with how the fraction absorbed was calculated.

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -315,14 +319,16 @@ private IContainer calculateGlobalPKAnalysisFor(Simulation simulation, Compound
          var vssPlasmaOverF = createParameter(VssPlasmaOverF, bloodPlasmaPK[Vss], VolumePerBodyWeight);
          var vdPlasma = createParameter(VdPlasma, bloodPlasmaPK[Vd], VolumePerBodyWeight);
          var vdPlasmaOverF = createParameter(VdPlasmaOverF, bloodPlasmaPK[Vd], VolumePerBodyWeight);
-         var vssPhysChem = createParameter(VssPhysChem, calculateVSSPhysChemFor(simulation, compoundName), VolumePerBodyWeight);
+         if(simulation is IndividualSimulation)
```
rwmcintosh, line 322: We're only going to calculate the VssPhysChem for individuals for now.

src/PKSim.Core/Model/PopulationAnalyses/BinsGroupingDefinition.cs
```diff
@@ -33,7 +34,7 @@ public virtual void CreateLimits(IPopulationDataCollector populationDataCollecto
 
          var limits = new List<double>();
          for (var i = 1; i < numberOfBins; i++)
-            limits.Add(sortedValues.Quantile((double) i / numberOfBins));
+            limits.Add(new SortedFloatArray(sortedValues, true).Quantile((double) i / numberOfBins));
```
msevestre, line 37: can we name the parmaeter here (instead of true: alreadySorted: true)

src/PKSim.Core/Services/RangeChartsDataCreator.cs
```diff
@@ -128,11 +129,12 @@ private static RangeXValue createRangeXValue(BinInterval xValueInterval, int num
       private RangeYValue createRangeYValue(IEnumerable<float> valuesInInterval)
       {
          var sortedArray = valuesInInterval.OrderedAndPurified();
+         var sortedFloatArray = new SortedFloatArray(sortedArray, true);
```
msevestre, line 132: same here. Can we name the parameter

src/PKSim.Core/Services/StatisticalDataCalculator.cs
```diff
@@ -119,7 +120,7 @@ private float[] calculateMinFor(FloatMatrix quantityResults)
 
       private float[] calculateMedianFor(FloatMatrix quantityResults)
       {
-         return calculateValueFor(quantityResults, x => x.Median());
+         return calculateValueFor(quantityResults, x => new SortedFloatArray(x, true).Median());
```
msevestre, line 123: name the parmaeter everywhere in fact :)

msevestre (changes requested):
little change for improve readability later on

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -315,14 +319,16 @@ private IContainer calculateGlobalPKAnalysisFor(Simulation simulation, Compound
          var vssPlasmaOverF = createParameter(VssPlasmaOverF, bloodPlasmaPK[Vss], VolumePerBodyWeight);
```
Yuri05, line 319: @msevestre Do you remember, why all blood pk parameters are calculated either from peripheral venous blood or from venous blood (depending on the species), but **bioavailability is always calculated from the venous blood**?

src/PKSim.Core/Services/PKAnalysesTask.cs
rwmcintosh: The calculation is different from every other pk parameter. It's adding up organ volumes for the individual and dividing by body weight. They body weight parameter is scaled for each individual for all the pk parameter calculations, but organ volumes are not. Basically, the VssPhysChem calculation for each individual is not as simple as the others.

Yuri05: OK, I see.
Actually it's because the function `calculateVSSPhysChemFor `lacks the individual id to pick the correct volumes (and other parameters like HCT) from the proper individual.
If it's too much work for this extension right now - can you please create an issue for this?

rwmcintosh: It's more about getting the correct values from the simulation given the IndividualId. Check in VSSCalculator.cs

```
public IParameter VSSPhysChemFor(Simulation simulation, string compoundName)
```
If there's an easy way to pull the required parameters for the individual, I can arrange for it.

msevestre: There is not..the function needs to be rewritten completely. (What we do with. Bodyweight scaling, we have to do for all parameters). We probably need a function tthet does it, reset the parameters etc. 
Just too much work I think

Yuri05: I think there is an easy way.. but let's ask @msevestre about that :)

msevestre: Is there an easy way? Am I missing something?

rwmcintosh: @msevestre was of the opinion yesterday that there's not an easy way. I'll open an issue for it anyway. Let's not make it part of this PR.

#2414

rwmcintosh:
All requested changes were made. New issue created for VssPhysChem

src/PKSim.Core/Services/PKAnalysesTask.cs
Yuri05: >Is there an easy way? Am I missing something?

Well, as far as I can see Individual Ids used/generated by the pop runner are always 0,..., N-1 - so an Individual Id can be just used as index when pulling a values array for a parameter?
Which is of course error prone and could be improved, but it's how the population runner works at the moment, right?
https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/1227a5c738c63a68b4c6e46bb2d5395d2bdc96b6/src/PKSim.Core/Services/IPopulationExportTask.cs#L150-L163

msevestre: sure that's the theory. But the runner does not do this at all. It creates a table that is then apply via variation in SimModel. The only place were we do something like this is for BW scaling.  

So in short. There is no easy way. Since this was not implemented before, I am perfectly happy to say that all global pk are available except VSS Phys Chem that will be implemented at a later time (or never)

PR #2426, opened 2022-11-12
Fixes #2424 output mapping serializatin extra

### LINKED ISSUE #2424: OutputMapping should be exported to MoBi?
I think this belongs to the simulation transfer between PKSim and MoBi 
@rwmcintosh @PavelBal @Yuri05 @georgeDaskalakis 

or is there a reason why this was not done?

msevestre:


msevestre:
This is a follow up of last PR to make sure the cloning is done properly

PR #2428, opened 2022-11-14
Fixes #2401 plots lost during clone

### LINKED ISSUE #2401: 11.1 Residuals vs. Time plot: no refresh
If I change the settings on the observed data tab in a simulation (scaling or weight): Residual vs. time Plots are not refreshed.
I have to close and reopen the simulation to see the changes.

And if I clear the mapping of observed data set: it is not removed from the plot even after reopening the simulation, so I need to create a new plot (this also applies to Predicted vs Observed plots)

msevestre:


src/PKSim.Core/Chart/SimulationTimeProfileChart.cs
```diff
@@ -0,0 +1,9 @@
+﻿using OSPSuite.Core.Chart;
+
+namespace PKSim.Core.Chart
+{
+   public class SimulationTimeProfileChart : AnalysisChart
```
msevestre, line 5: This was somewhat of a duplcate class. One is abstract in core now. This is the implementation

src/PKSim.Infrastructure/Services/ChartTemplatingTask.cs
```diff
@@ -90,14 +90,14 @@ private void addObservedDataToChart(IChartEditorPresenter chartEditorPresenter,
...
-      public SimulationTimeProfileChart CloneChart(SimulationTimeProfileChart originalChart, IndividualSimulation simulation)
+      public T CloneChart<T>(T originalChart, IndividualSimulation simulation) where T: AnalysisChart
       {
          var clonedChart = _chartFactory.Create(originalChart.GetType()).WithName(originalChart.Name);
          clonedChart.UpdatePropertiesFrom(originalChart, _cloneManager);
 
          initializeFromTemplate(originalChart, clonedChart, simulation);
 
-         return clonedChart.DowncastTo<SimulationTimeProfileChart>();
+         return clonedChart.DowncastTo<T>();
```
msevestre, line 100: needs to be generic here or at least not so specific

rwmcintosh (approved):

rwmcintosh:
@msevestre I'm looking at Open-Systems-Pharmacology/OSPSuite.Core#1780 and I can see that the charts are not cloned at all in PI. Is that the correct behavior in the case of PI?

msevestre:
Hard to tell. I would have expected the charts to be cloned as well

PR #2442, opened 2022-11-25
Fixes #2344 Implement "Parameter Value Export" for Expression Profile…

### LINKED ISSUE #2344: Implement "Parameter Value Export" for Expression Profiles BB in PK-Sim

rwmcintosh:


src/PKSim.Core/Model/ExpressionProfile.cs
```diff
@@ -27,6 +28,12 @@ public Individual Individual
       public ExpressionProfile() : base(PKSimBuildingBlockType.ExpressionProfile)
       {
       }
+      
+      public override IReadOnlyList<T> GetAllChildren<T>()
+      {
+         return Individual.AllMoleculeContainersFor(Molecule)
```
msevestre, line 34: Why not returning GetAllChildeen of individual? That should be equivalent but here you almost know too much

rwmcintosh: I pattern this from what I observed in the DTO mapper that creates the parameter list for the main display of BB. If you say it's adequate for Individual.GetAllChildren, then I believe you!

PR #2461, opened 2022-12-07
2364 export expression profile domain object 

### LINKED ISSUE #2364: Export Expression Profile domain object
1. Right click on Expression BB
2. Show Context Menu
3. Select 'Export'
4. Select the file
5. Save on disk as PKML

georgeDaskalakis:
closes #2364

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,6 +3,8 @@ image: Visual Studio 2019
 
 version: '{build}'
 
+# TODO make sure to revert this temporary 12.0.
+
 init:
 - ps: Update-AppveyorBuild -Version "12.0.$($env:appveyor_build_version)"
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,7 +3,7 @@ image: Visual Studio 2019
 
 version: '{build}'
 
-# TODO make sure to revert this temporary 12.0.
+# TODO make sure to revert this temporary 12.0. to $($env:ospsuite_version).
 
 init:
 - ps: Update-AppveyorBuild -Version "12.0.$($env:appveyor_build_version)"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,23 +4,24 @@ image: Visual Studio 2019
 version: '{build}'
 
 init:
-- ps: Update-AppveyorBuild -Version "$($env:ospsuite_version).$($env:appveyor_build_version)"
+- ps: Update-AppveyorBuild -Version "12.0.$($env:appveyor_build_version)"
 
 clone_depth: 1
 
 dotnet_csproj:
   patch: true
   file: '**\*.csproj'
-  version: '$(ospsuite_version).$(build_number)'
-  assembly_version: '$(ospsuite_version).$(build_number)'
-  file_version: '$(ospsuite_version).$(build_number)'
-  informational_version: '$(ospsuite_version).$(build_number)'
+  version: '12.0.$(build_number)'
+  assembly_version: '12.0.$(build_number)'
+  file_version: '12.0.$(build_number)'
+  informational_version: '12.0.$(build_number)'
 
 skip_branch_with_pr: true
 
 branches:
   only:
     - develop
+    - OSMOSES
 
 pull_requests:
   do_not_increment_build_number: true
--- a/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
+++ b/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
@@ -26,8 +26,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="11.1.65" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.65" />
+    <PackageReference Include="OSPSuite.Assets" Version="12.0.74" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="12.0.74" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
+++ b/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
@@ -26,8 +26,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="11.1.72" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.72" />
+    <PackageReference Include="OSPSuite.Assets" Version="11.1.65" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.65" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
+++ b/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
@@ -26,8 +26,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="11.1.62" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.62" />
+    <PackageReference Include="OSPSuite.Assets" Version="11.1.65-wqtsahcc" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.65-wqtsahcc" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
+++ b/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
@@ -26,8 +26,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="11.1.65-wqtsahcc" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.65-wqtsahcc" />
+    <PackageReference Include="OSPSuite.Assets" Version="11.1.65-ikjtreet" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.65-ikjtreet" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
+++ b/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
@@ -26,8 +26,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="11.1.59" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.59" />
+    <PackageReference Include="OSPSuite.Assets" Version="11.1.62" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.62" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
+++ b/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
@@ -26,8 +26,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="11.1.65-ikjtreet" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.65-ikjtreet" />
+    <PackageReference Include="OSPSuite.Assets" Version="11.1.72" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.72" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
+++ b/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
@@ -26,8 +26,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="12.0.0-SisUj" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="12.0.0-SisUj" />
+    <PackageReference Include="OSPSuite.Assets" Version="11.1.65" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.65" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
+++ b/src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
@@ -26,8 +26,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="11.1.65" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.65" />
+    <PackageReference Include="OSPSuite.Assets" Version="12.0.0-SisUj" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="12.0.0-SisUj" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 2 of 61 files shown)


src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
```diff
@@ -26,8 +26,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="11.1.65" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="11.1.65" />
+    <PackageReference Include="OSPSuite.Assets" Version="12.0.0-SisUj" />
```
msevestre, line 29: no please. Let's not create PR referencing unknow number

msevestre: @georgeDaskalakis You really have to check this out. I'd really appreciate it if you would review the PR (by going to the files) before asking me to review.

src/PKSim.Core/Mappers/ExpressionProfileToExpressionProfileBuildingBlockMapper.cs
```diff
@@ -0,0 +1,77 @@
...
+         _entityPathResolver = entityPathResolver;
+         _formulaFactory = formulaFactory;
+      }
+
+      public ExpressionProfileBuildingBlock MapFrom(ExpressionProfile expressionProfile)
+      {
+         var expressionProfileBuildingBlock = _objectBaseFactory.Create<ExpressionProfileBuildingBlock>();
+
+         expressionProfileBuildingBlock.Name = expressionProfile.Name;
+         expressionProfileBuildingBlock.PKSimVersion = ProjectVersions.Current;
```
msevestre, line 33: This  probably should be the Dipsplay version as opposed to the current which is a number that does not mean anything? I am not quite sure how this is supposed to be used but version 78 won't mean anything to anyone but us

src/PKSim.Core/Mappers/ExpressionProfileToExpressionProfileBuildingBlockMapper.cs
```diff
@@ -0,0 +1,77 @@
...
+            default:
+               expressionProfileBuildingBlock.Type = ExpressionTypes.ProteinBindingPartner;
+               break;
+         }
+
+         var allParameters = expressionProfile.GetAllChildren<IParameter>();
+
+         foreach (var parameter in allParameters)
+         {
+            var expressionParameter = _objectBaseFactory.Create<ExpressionParameter>();
```
msevestre, line 54: can we have a private method here that justs maps a parameter. Thsi is a fairly simple method but it's quite long and I think they are are clear sections that can be moved  out

1/ the ExpressionTyupes to expressionProfileBuildingBlock.Type (BTW, any reason why we have different types and not the same?) (so line 36 to 48 can be a method with simple return statements

Like 54 to 70 is also a mapper method so to speak. 

Thoughts?

src/PKSim.Presentation/DTO/ExpressionProfiles/ExpressionProfileDTO.cs
```diff
@@ -48,7 +48,7 @@ public virtual string MoleculeName
 
       private bool moleculeSpeciesCategoryValid(string name) => !_allExistingExpressionProfileNames.Contains(name?.TrimmedValue()?.ToLower());
 
-      public string Name => ExpressionProfileName(MoleculeName, Species, Category);
+      public string Name => Constants.ContainerName.ExpressionProfileName(MoleculeName, Species?.DisplayName, Category);
```
msevestre, line 51: ExpressionProfileName. Isn't this method also in core?

src/PKSim.Presentation/UICommands/ExportExpressionProfileToPkmlCommand.cs
```diff
@@ -0,0 +1,34 @@
...
+using PKSim.Core.Model;
+
+namespace PKSim.Presentation.UICommands
+{
+   public class ExportExpressionProfileToPkmlCommand : ObjectUICommand<ExpressionProfile>
+   {
+      private readonly IDialogCreator _dialogCreator;
+      private readonly IPKMLPersistor _pkmlPersistor;
+      private readonly IExpressionProfileToExpressionProfileBuildingBlockMapper _mapper;
+      public ExportExpressionProfileToPkmlCommand(IDialogCreator dialogCreator, IPKMLPersistor pkmlPersistor,
```
msevestre, line 16: space is missign here

tests/PKSim.Tests/Core/ExpressionProfileToExpressionProfileBuildingBlockMapperSpecs.cs
```diff
@@ -0,0 +1,98 @@
...
+using OSPSuite.Core.Domain.Formulas;
+using OSPSuite.Core.Domain.Services;
+using PKSim.Core.Mappers;
+using PKSim.Core.Model;
+using IFormulaFactory = PKSim.Core.Model.IFormulaFactory;
+
+namespace PKSim.Core
+{
+   public abstract class
+      concern_for_ExpressionProfileToExpressionProfileBuildingBlockMapper : ContextSpecification<ExpressionProfileToExpressionProfileBuildingBlockMapper>
```
msevestre, line 16: Please run resharper in this file

msevestre (changes requested):

src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
georgeDaskalakis: @msevestre sorry for that, you are right. My suggestion here would be: since we are using this made up version of the nugets locally, should we actually add the .csproj to .gitignore (ONLY for this branch of course).
If we don't I can guarantee you this will happen again: like yesterday: if I review the changes and then need to adjust a small thing in a file, and not stage only that file we will end up getting all these changes in the project. I am not saying I will not try to be carefull, but it will definetely happen again.
@rwmcintosh what do you think about the gitignore?

src/PKSim.Core/Mappers/ExpressionProfileToExpressionProfileBuildingBlockMapper.cs
georgeDaskalakis: changed to ProjectVersions.Current.VersionDisplay. @msevestre that is the correct version right? like v11.1?

georgeDaskalakis: @msevestre @rwmcintosh then actually we should also change PKSimVersion from int to string, right? Otherwise I would say the way this is written is how it was supposed to work....

src/PKSim.Presentation/UICommands/ExportExpressionProfileToPkmlCommand.cs
georgeDaskalakis: added

src/PKSim.Presentation/DTO/ExpressionProfiles/ExpressionProfileDTO.cs
georgeDaskalakis: @msevestre  this is the on :D in Constants.ContainerName

src/PKSim.Core/Mappers/ExpressionProfileToExpressionProfileBuildingBlockMapper.cs
georgeDaskalakis: there also exists .VersionAsString

src/PKSim.Core/Mappers/ExpressionProfileToExpressionProfileBuildingBlockMapper.cs
georgeDaskalakis: done at least those, if you want we can discuss

src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
rwmcintosh: This is my fault. We can always back out the CSProj for PR's but the issue will be that when you clone the repo, the stuff won't build anyway because there is no centrally available ospsuite.core libraries that match.

This is a drawback of making your own local nuget repo.

For me, it is not a big difference if you leave this saying 11.1.65 or 12.0.0-fdjka because in both cases you need to create your own OSPSuite core before you can do anything.

rwmcintosh: I wouldn't add them to .gitignore, because if you add a new file to a project, that change won't be available for the next person.
It's whack-a-mole. It makes debugging much easier, but introduced a new problem.

I'd sooner have the non-existent core number.

Although, at the moment, I'm the only one using this new method of producing core builds.

georgeDaskalakis: the files paths are not stored in the .csproj. we went through the files with @rwmcintosh and actually the gitignore idea could work. @msevestre @Yuri05 let's discuss on monday

src/PKSim.Core/Mappers/ExpressionProfileToExpressionProfileBuildingBlockMapper.cs
georgeDaskalakis: ok, VersionDisplay is the correct one, changing to that

src/PKSim.Core/Mappers/ExpressionProfileToExpressionProfileBuildingBlockMapper.cs
```diff
@@ -0,0 +1,90 @@
...
+
+         expressionParameter.Name = parameter.Name;
+
+         expressionParameter.Path = _entityPathResolver.ObjectPathFor(parameter);
+         expressionParameter.Dimension = parameter.Dimension;
+         expressionParameter.DisplayUnit = parameter.DisplayUnit;
+         return expressionParameter;
+      }
+
+      private static void setExpressionType(ExpressionProfile expressionProfile, ExpressionProfileBuildingBlock expressionProfileBuildingBlock)
```
msevestre, line 73: no this is not what I meant sorry. While this, works, try to avoid side effects in methods

sthg like this

mapExpressionType(moleculeType){
   switch (moleculeType)
         {
            case QuantityType.Enzyme:
              return  ExpressionTypes.MetabolizingEnzyme;
            case QuantityType.Transporter:
               return ExpressionTypes.TransportProtein;
            default:
               expressionProfileBuildingBlock.Type = ExpressionTypes.ProteinBindingPartner;
               break;
         }
etc..
}

ALSO NO DEFAULT PLEASE. The default case should throw an exception because we are missing a type.

src/PKSim.Core/Model/ExpressionProfile.cs
```diff
@@ -38,7 +45,7 @@ public virtual string Category
 
       public override string Name
       {
-         get => ExpressionProfileName(MoleculeName, Species, Category);
+         get => Constants.ContainerName.ExpressionProfileName(MoleculeName, Species?.DisplayName, Category);
```
msevestre, line 48: why aren't you using static here where you do it everywhere else?

src/PKSim.Presentation/DTO/ExpressionProfiles/ExpressionProfileDTO.cs
msevestre, line 51: or here. Why aren't you using static here? It looks like we are calling something else now

src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
msevestre: I do not understand the reason why we would want to commit to our main `OSMOSES` branch code that does not compile. This does not make sense to me so I am probably missing the point. 
In my mind, osmoses is like develop. It should compile and be runnable at ALL time. So we cannot commit anything that is going to break. You can create as many branches as you want, if the goal is to share code etc, but in osmoses...I would expect this stuff to build.

Which means that we need an osmoses build of ospsuite.core that builds also where we do all of our PR. 
I'd love to understand the reason because there is probably some frictions going on that I do not see

src/PKSim.Core/Mappers/ExpressionProfileToExpressionProfileBuildingBlockMapper.cs
```diff
@@ -0,0 +1,90 @@
...
+         _entityPathResolver = entityPathResolver;
+         _formulaFactory = formulaFactory;
+      }
+
+      public ExpressionProfileBuildingBlock MapFrom(ExpressionProfile expressionProfile)
+      {
+         var expressionProfileBuildingBlock = _objectBaseFactory.Create<ExpressionProfileBuildingBlock>();
+
+         expressionProfileBuildingBlock.Name = expressionProfile.Name;
+         expressionProfileBuildingBlock.PKSimVersion = ProjectVersions.Current.VersionDisplay;
```
msevestre, line 33: Display is also not good. WE do not have a one to one mapping. We only increment the version if we have a serialization change
You need to grab the ApplicationConfiugration or sethg like this and get the version from there.

msevestre: probably this one here
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/IApplicationConfiguration.cs#L45

you can see what we export in the PopulationExporTTask when we dump a population to csv. This should be the same logic

src/PKSim.Assets.Images/PKSim.Assets.Images.csproj
rwmcintosh: We would need the nuget repo to be centrally created and stored along with numbering etc. We turned that off for OSMOSES branch due to conflicts with the build numbering while we are working on develop at the same time.

If they repositories can be somehow independent that would be another way. But based on the fact that Core is changing, and there's no central NuGet repo the developer will always have to create a Core build that works before they can build the apps.

Until we merge OSMOSES to develop, we will always have either the friction of no central Core NuGet, or conflicting (or alternating) Core build numbers between develop and OSMOSES

msevestre: I understand what we need but I don't understand why we would want to merge non compiling code. 
What is the `purpose` of this?

>(or alternating) Core build numbers between develop and OSMOSES

This is fine with me. We will reset all the counters as soon as we release 11.1 anyways

rwmcintosh: > What is the purpose of this?

As soon as we turned off the nuget build of OSPSuite.Core OSMOSES branch on AppVeyor this was going to happen. The purpose was to not create build number confusion with develop.

rwmcintosh: You just need to run the batch file to create your local repo to get it working.

In any case, if you think the core builds should be turned back on then that's fine too.

The whole purpose was to avoid the build numbering conflict, but if the cost is too great, then fine.  But it's one or the other.

msevestre: I just don't get it. Let's discuss on Monday

appveyor.yml
```diff
@@ -4,23 +4,24 @@ image: Visual Studio 2019
...
 clone_depth: 1
 
 dotnet_csproj:
   patch: true
   file: '**\*.csproj'
-  version: '$(ospsuite_version).$(build_number)'
-  assembly_version: '$(ospsuite_version).$(build_number)'
-  file_version: '$(ospsuite_version).$(build_number)'
-  informational_version: '$(ospsuite_version).$(build_number)'
+  version: '12.0.$(build_number)'
```
msevestre, line 14: Can we put a comment here with the old code and that we need to revert?

src/PKSim.Presentation/Presenters/ContextMenus/ExpressionProfileContextMenu.cs
```diff
@@ -16,7 +21,32 @@ public ExpressionProfileContextMenu(ExpressionProfile expressionProfile) : base(
...
+         yield return SaveAsSystemTemplateMenuFor(expressionProfile);
+
+         yield return AddToJournalMenuFor(expressionProfile);
+
+         yield return ExsportToPkml(expressionProfile);
+      }
+
+      private  IMenuBarItem ExsportToPkml (ExpressionProfile expressionProfile)
+      {
+
```
msevestre, line 46: please run resharper here. Thsi is all out of place

msevestre (approved):
This can be merged. 
Appveyor file  to update however because this is just a WO

PR #2476, opened 2022-12-12
2475 expression profile from pksim

rwmcintosh:


src/PKSim.UI.Starter/ApplicationStartup.cs
```diff
@@ -0,0 +1,115 @@
...
+            InfrastructureRegister.LoadSerializers(container);
+         }
+
+         return container;
+      }
+
+      // /// <summary>
+      // /// Use this method to register all dependencies in Core.UI and Core.Presentation
+      // /// </summary>
+      // private static void registerImplementationsFromCore(IContainer container)
```
rwmcintosh, line 84: I left this in because I know I'll be working on this more for Individuals. Once we have the feature working for both, I can remove this full registration code instead of commenting it out.

src/PKSim.UI.Starter/ApplicationStartup.cs
```diff
@@ -0,0 +1,115 @@
...
+      {
+         if (_container != null)
+            return _container;
+
+         _container = InitializeForStartup(shell);
+
+         return _container;
+      }
+
+      public static IContainer InitializeForStartup(IShell shell)
```
rwmcintosh, line 44: This is cool - we are actually registering the MoBi main view as IShell and then all the sub-views are shown correctly because IoC makes sure they are using MoBi as the Parent view.

msevestre, line 44: Why is the method public?

src/PKSim.UI.Starter/ExpressionProfileCreator.cs
```diff
@@ -0,0 +1,41 @@
...
+      private static object createExpressionProfile<T>(IShell shell) where T : IndividualMolecule
+      {
+         var container = ApplicationStartup.Initialize(shell);
+
+         using (var presenter = container.Resolve<ICreateExpressionProfilePresenter>())
+         {
+            var workspace = container.Resolve<IWorkspace>();
+            workspace.Project = new PKSimProject();
+            var mapper = container.Resolve<IExpressionProfileToExpressionProfileBuildingBlockMapper>();
+            presenter.Create<T>();
```
msevestre, line 35: Feels like this guy should return the presenter.

msevestre: Sorry the expression profiles

msevestre (approved):
Looks good. Just one little update potentially to the API so that the create methods actually returns the created object? Unless it's about command?

rwmcintosh:
> Looks good. Just one little update potentially to the API so that the create methods actually returns the created object? Unless it's about command?

Yes, it's about those commands. They otherwise signal other callers of this Create method whether or not the dialog was canceled.

We could pull out another method that does basically the same stuff, but without the commands and then return the ExpressionProfile.

src/PKSim.Presentation/Presenters/ExpressionProfiles/CreateExpressionProfilePresenter.cs
```diff
@@ -47,8 +47,12 @@ public class CreateExpressionProfilePresenter : AbstractDisposableCommandCollect
          _view.Caption = PKSimConstants.UI.CreateExpressionProfile;
          _view.BindTo(_dto);
          _view.Display();
+
+         ExpressionProfile = null;
```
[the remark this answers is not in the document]
rwmcintosh: Now that you mention it, I should be able to use the same convention as other callers: if the command IsEmpty() then the expression profile dialog was canceled and the result should not be used.

PR #2494, opened 2022-12-21
2480 individual serialization

rwmcintosh:


src/PKSim.Core/Mappers/PathAndValueBuildingBlockMapper.cs
```diff
@@ -0,0 +1,94 @@
...
+         return buildingBlock;
+      }
+
+      private TBuilder mapBuilderParameter(IParameter parameter)
+      {
+         var builderParameter = _objectBaseFactory.Create<TBuilder>();
+
+         if (parameter.Formula != null && parameter.Formula.IsCachable())
+         {
+            if (!_formulaCache.Contains(parameter.Formula.Name))
```
rwmcintosh, line 49: I could not understand how to use the formula factory in this more generalized case. We had previously hard-coded the category to EXPRESSION_PARAMETERS but I couldn't generalize it.

In the end, I just did my own uniqueness cache scheme

src/PKSim.Core/Mappers/ExpressionProfileToExpressionProfileBuildingBlockMapper.cs
```diff
@@ -1,52 +1,34 @@
...
+         base(objectBaseFactory, entityPathResolver, applicationConfiguration, lazyLoadTask)
       {
-         _objectBaseFactory = objectBaseFactory;
-         _entityPathResolver = entityPathResolver;
-         _formulaFactory = formulaFactory;
-         _applicationConfiguration = applicationConfiguration;
       }
 
-      public ExpressionProfileBuildingBlock MapFrom(ExpressionProfile expressionProfile)
+      protected override IReadOnlyList<IParameter> AllParametersFor(ExpressionProfile sourcePKSimBuildingBlock)
```
msevestre, line 22: rename the parameter here. No need to keep using the generic name of the base class IMO

src/PKSim.Core/Mappers/PathAndValueBuildingBlockMapper.cs
msevestre: I don't understand

src/PKSim.Presentation/Mappers/IndividualToIndividualBuildingBlockMapper.cs
```diff
@@ -0,0 +1,80 @@
...
+      }
+
+      public override IndividualBuildingBlock MapFrom(Individual input)
+      {
+         var buildingBlock = base.MapFrom(input);
+         buildingBlock.Icon = input.Icon;
+
+         addOriginDataToBuildingBlock(buildingBlock, Assets.PKSimConstants.UI.DiseaseState, input.OriginData.DiseaseState?.DisplayName);
+         addOriginDataToBuildingBlock(buildingBlock, Assets.PKSimConstants.UI.Species, input.Species?.DisplayName);
+         addOriginDataToBuildingBlock(buildingBlock, Assets.PKSimConstants.UI.DiseaseState, input.OriginData.DiseaseState?.DisplayName);
```
msevestre, line 41: this is twice here. Something is off with the copy pasting.

src/PKSim.Presentation/Mappers/IndividualToIndividualBuildingBlockMapper.cs
```diff
@@ -0,0 +1,80 @@
...
+      {
+         if (parameter == null)
+            return;
+
+         addOriginDataToBuildingBlock(buildingBlock, keyForOriginDataParameter(key, parameter), $"{parameter.Value} {parameter.Unit}");
+      }
+
+      private string keyForOriginDataParameter(string key, OriginDataParameter parameter)
+      {
+         return string.IsNullOrEmpty(parameter.Name) ? key : parameter.Name;
```
msevestre, line 69: I don't quite get this. How can you have  a parameter without a name? and if this is the case, why add the it at all ?

msevestre (changes requested):

src/PKSim.Core/Mappers/PathAndValueBuildingBlockMapper.cs
msevestre: yeah but the code is not  equivalent. One you were creating anew instance, here you are adding the same reference

msevestre: plus you are not adding it to the building block FormulaCache.  they won';t be serialized

rwmcintosh: They are copied over to the bb somewhere else

src/PKSim.Core/Mappers/PathAndValueBuildingBlockMapper.cs
```diff
@@ -0,0 +1,94 @@
...
+      {
+         var allParameters = AllParametersFor(sourcePKSimBuildingBlock);
+
+         foreach (var parameter in allParameters)
+         {
+            var builderParameter = mapBuilderParameter(parameter);
+            buildingBlock.Add(builderParameter);
+         }
+
+         foreach (var formula in _formulaCache)
```
rwmcintosh, line 77: Here we copy all the formulae into the BB for serialization

src/PKSim.Presentation/Mappers/IndividualToIndividualBuildingBlockMapper.cs
rwmcintosh: I don't know how the OriginDataParameters are created for the individual, but many do not have the Name property set.

image

rwmcintosh:
Merging this - if OriginDataParameter without a name is an issue that needs to be resolved, let's open another issue to address that.

PR #2503, opened 2023-01-05
Fixes #2484 bug when exporting pop simulation

### LINKED ISSUE #2484: Bug when exporting population simulation
### Discussed in https://github.com/Open-Systems-Pharmacology/Forum/discussions/1411

<div type='discussions-op-text'>

<sup>Originally posted by **moPellowe** December 15, 2022</sup>
I have found an issue with the population simulation export in PK-Sim. I have a simulation that has been run in PK-Sim and I can see the upper and lower values of the confidence interval via the plot.
image
image

But when I export the results of the population simulation, the results for the lower and upper values of the CI are sometimes different.
image

It seems that it is an issue that occurs only if you have more than one compound in the plot and you use a non-molar unit. (thanks, @tobiasK2001, for figuring that out.)</div>

msevestre:


src/PKSim.Core/Chart/ChartData.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace PKSim.Core.Chart
 {
-   public class ChartData<TX, TY> :  IComparer<PaneData<TX, TY>> where TX : IXValue where TY : IYValue
```
msevestre, line 8: Just formatting in this file

src/PKSim.Core/Mappers/BoxWhiskerChartDataToDataTableMapper.cs
```diff
@@ -33,16 +33,18 @@ protected override IEnumerable<DataRow> AddSpecificChartValues(DataRow row, Curv
          for (var i = 0; i < curveData.XValues.Count; i++)
          {
             var newRow = row.Table.NewRow();
+            var yValue = curveData.YValues[i];
             newRow.ItemArray = row.ItemArray;
             newRow[_xValue] = curveData.XValues[i].ToString(curveData.XAxis);
-            newRow[_lowerWhisker] = ValueForDataTableFor(curveData.YAxis, curveData.YValues[i].LowerWhisker);
-            newRow[_lowerBox] = ValueForDataTableFor(curveData.YAxis, curveData.YValues[i].LowerBox);
-            newRow[_median] = ValueForDataTableFor(curveData.YAxis, curveData.YValues[i].Median);
```
msevestre, line 40: this was the error everywhere. Using the dimension of the Y axis instead of the YDimensiion defined at the curve level (was not easy to find!)

src/PKSim.Core/Mappers/ChartDataToDataTableMapper.cs
```diff
@@ -149,21 +151,22 @@ private IEnumerable<DataColumn> getEmptyColumns(DataTable table)
...
+      protected virtual IReadOnlyList<string> GetDataFields(CurveData<TXValue, TYValue> curveData) => new List<string>();
 
-      protected virtual IReadOnlyList<string> GetRowFields(CurveData<TXValue, TYValue> curveData)
-      {
-         return new List<string>();
-      }
+      protected virtual IReadOnlyList<string> GetRowFields(CurveData<TXValue, TYValue> curveData) => new List<string>();
+
+      //For x values, we can use the dimension of the x axis right away as it is always the same for all curves
+      protected object xValueForDataTableFor<TX, TY>(CurveData<TX, TY> curveData, double value) where TX : IXValue where TY : IYValue => ValueForDataTableFor(curveData.XAxis, curveData.XAxis.Dimension, value);
```
msevestre, line 162: I created two helper methods so that I don't have to duplicate the logic

src/PKSim.Core/Mappers/RangeChartDataToDataTableMapper.cs
```diff
@@ -42,16 +42,20 @@ protected override IEnumerable<DataRow> AddSpecificChartValues(DataRow row, Curv
          for (var i = 0; i < curveData.XValues.Count; i++)
          {
             var newRow = row.Table.NewRow();
+            var xValue = curveData.XValues[i];
```
msevestre, line 45: sometimes it's good to avoid a bit of code dup by introducing a variable lol...oh my

rwmcintosh (approved):

PR #2510, opened 2023-01-11
Fixes #2508 Do not export calculation methods with Individual when there is only one possible value

### LINKED ISSUE #2508: Do not export calculation methods with Individual when there is only one possible value

rwmcintosh:


src/PKSim.Presentation/Mappers/IndividualToIndividualBuildingBlockMapper.cs
```diff
@@ -46,13 +54,17 @@ public override IndividualBuildingBlock MapFrom(Individual input)
...
-         input.OriginData.AllCalculationMethods().MapAllUsing(_calculationMethodDTOMapper)
+         input.OriginData.AllCalculationMethods().Where(cm => hasMoreThanOneOption(cm, input.Species)).MapAllUsing(_calculationMethodDTOMapper)
             .Each(x => addOriginDataToBuildingBlock(buildingBlock, x.DisplayName, x.CategoryItem.DisplayName));
 
          buildingBlock.OriginData.ValueOrigin = input.OriginData.ValueOrigin.Clone();
 
          return buildingBlock;
       }
+      private bool hasMoreThanOneOption(CalculationMethod calculationMethod, Species species)
+      {
```
msevestre, line 65: create a method in ICalculationMethodCategoryRepository? TELL don't ASK?

msevestre (commented):
suggestion is to move the method to the repo because this behavior is not the individual Mapper concern

PR #2512, opened 2023-01-14
Fixes #2255 resetting concentration issue

### LINKED ISSUE #2255: Resetting initial concentrations in the simulation and is not propagated to BB after commit
In the example from https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2253 I click on the Reset button in the initial concentration **in the simulation**

Then I commit the Individual from the simulation back to the BB.
* In the **Individual**-BB the value is correctly resetted to 0 (btw, suddenly I see an entry for Endosome initial concentration which was not shown before. Why?)
grafik
* But in the Protein-Expression-BB the value is still 1. Closing/Reopening the Protein-Expression-BB does not help, so it's not a refresh progress. And also here I suddenly get the entry for Endosome...
grafik

msevestre:
One of the hardest bugs to understand..well done @Yuri05 for finding a way to reproduce it

### DIFF
--- a/src/PKSim.Core/Commands/RemoveMoleculeFromIndividualCommand.cs
+++ b/src/PKSim.Core/Commands/RemoveMoleculeFromIndividualCommand.cs
@@ -1,3 +1,4 @@
+using OSPSuite.Utility.Extensions;
 using PKSim.Assets;
 using PKSim.Core.Events;
 using PKSim.Core.Model;
@@ -29,8 +30,8 @@ namespace PKSim.Core.Commands
 
       protected override void PerformExecuteWith(IExecutionContext context)
       {
-         _individual.RemoveMolecule(_molecule);
-         context.Unregister(_molecule);
+         var removedContainer = _individual.RemoveMolecule(_molecule);
+         removedContainer.Each(context.Unregister);
          context.PublishEvent(new RemoveMoleculeFromSimulationSubjectEvent<Individual> {Entity = _molecule, Container = _individual});
       }
    }
--- a/src/PKSim.Core/Commands/RemoveMoleculeFromPopulationCommand.cs
+++ b/src/PKSim.Core/Commands/RemoveMoleculeFromPopulationCommand.cs
@@ -1,4 +1,5 @@
-﻿using PKSim.Assets;
+﻿using OSPSuite.Utility.Extensions;
+using PKSim.Assets;
 using PKSim.Core.Events;
 using PKSim.Core.Model;
 
@@ -29,8 +30,8 @@ namespace PKSim.Core.Commands
 
       protected override void PerformExecuteWith(IExecutionContext context)
       {
-         _population.RemoveMolecule(_molecule);
-         context.Unregister(_molecule);
+         var removedContainer= _population.RemoveMolecule(_molecule);
+         removedContainer.Each(context.Unregister);
          context.PublishEvent(new RemoveAdvancedParameterContainerFromPopulationEvent(_population));
          context.PublishEvent(new RemoveMoleculeFromSimulationSubjectEvent<Population> {Entity = _molecule, Container = _population});
       }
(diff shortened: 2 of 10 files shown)


src/PKSim.Core/Commands/RemoveMoleculeFromIndividualCommand.cs
```diff
@@ -29,8 +30,8 @@ protected override void ClearReferences()
 
       protected override void PerformExecuteWith(IExecutionContext context)
       {
-         _individual.RemoveMolecule(_molecule);
-         context.Unregister(_molecule);
+         var removedContainer = _individual.RemoveMolecule(_molecule);
```
msevestre, line 33: This was also a bug that I fixed at the same time (same logical error as the actual bug). Removing a molecule only unregistered the TOP container but not all sub containers.

src/PKSim.Core/Commands/RemoveMoleculeFromIndividualCommand.cs
```diff
@@ -29,8 +30,8 @@ protected override void ClearReferences()
...
+         removedContainer.Each(context.Unregister);
```
msevestre, line 34: So i unregister them all when removing

src/PKSim.Core/Commands/RemoveMoleculeFromPopulationCommand.cs
```diff
@@ -29,8 +30,8 @@ protected override void ClearReferences()
 
       protected override void PerformExecuteWith(IExecutionContext context)
       {
-         _population.RemoveMolecule(_molecule);
-         context.Unregister(_molecule);
+         var removedContainer= _population.RemoveMolecule(_molecule);
+         removedContainer.Each(context.Unregister);
```
msevestre, line 34: Probably a way to unfiy those two implementations but at the ned of the day, they are also different use cases so I am leaving it like this for now

src/PKSim.Core/Services/MoleculeExpressionTask.cs
```diff
@@ -67,7 +71,11 @@ public ICommand RemoveExpressionProfileFor(IndividualMolecule molecule, TSimulat
       public ICommand AddExpressionProfile(TSimulationSubject simulationSubject, ExpressionProfile expressionProfile)
       {
          var moleculeFactory = _individualMoleculeFactoryResolver.FactoryFor(expressionProfile.Molecule);
+         //this calls not only creates a new molecule but modify the subject in place to have the link to the expression profile
+         //by adding sub containers in the individual structure. The factory does not register the container and parameters so we need to do it
          var molecule = moleculeFactory.AddMoleculeTo(simulationSubject, expressionProfile.MoleculeName);
+         //we register the whole individual again.
+         _registrationTask.Register(simulationSubject);
```
msevestre, line 78: This was the problem here.
The parameters of the molecule in the individual were not registered (only at creation time...when loading the project, they were registered=>NO BUG)

msevestre: there is really not much else to this bug.. But this took me one day lol

Yuri05: Cool, well done! 👍

Yuri05 (approved):

src/PKSim.Core/Services/MoleculeExpressionTask.cs
msevestre: Actually this is the right fix but not the right place. I need to move this to the command (symmetrical to the unregister)

src/PKSim.Core/Commands/AddMoleculeToPopulationCommand.cs
```diff
@@ -15,6 +15,8 @@ public AddMoleculeToPopulationCommand(IndividualMolecule molecule, Population po
       protected override void PerformExecuteWith(IExecutionContext context)
       {
          base.PerformExecuteWith(context);
+         //we register the whole individual again as sub container for the molecule were added to the individual
```
rwmcintosh, line 18: Should this comment should be revised to 'population' instead of 'individual'?

rwmcintosh (approved):

rwmcintosh (approved):

PR #2526, opened 2023-01-31
2525 creating individual from MoBi (PK-Sim code changes)

### LINKED ISSUE #2525: Create individual from PK-Sim
@rwmcintosh commented on [Wed Nov 02 2022](https://github.com/Open-Systems-Pharmacology/MoBi/issues/783)

You should see only the view of PK-sim where individuals are created

georgeDaskalakis:
closes #2525


rwmcintosh:
We did some pair programming this morning to remove some of the registrations and their dependencies. I found one more after we hung up that removed the need for any changes in OSPSuite.Core.

I know there's another issue around stating that we don't need to support the ExpressionProfiles tab when we open the dialog from MoBi. If we can somehow avoid the registration, then more of the manual registrations could be removed that support that tab.

The registration list is quite long, and we should investigate if we could register by convention for specific namespaces. When I did a quick look around, we might have to organize the namespaces more optimally for that to happen.

It just means that we won't have as much headache in the future maintaining this.

If you prefer to leave this branch open as a mini-feature branch, we can look at a couple of these optimizations in their own issues and merge them into this branch before merging the whole thing into OSMOSES.

src/PKSim.UI.Starter/ApplicationStartup.cs
```diff
@@ -54,24 +74,86 @@ private static IContainer initializeForStartup(IShell shell)
...
 
-         // registerImplementationsFromCore(container);
          registerMinimalImplementations(container);
 
 
          using (container.OptimizeDependencyResolution())
          {
+            container.RegisterImplementationOf(shell as BaseShell);
             container.RegisterImplementationOf(shell);
+            container.RegisterImplementationOf(UserLookAndFeel.Default);
```
rwmcintosh, line 84: @georgeDaskalakis simpler than we imagined

src/PKSim.UI.Starter/ApplicationStartup.cs
```diff
@@ -54,24 +75,93 @@ private static IContainer initializeForStartup(IShell shell)
...
+            container.Register<IUserSettings, UserSettings>();
+            container.Register<INumericFormatterOptions, NumericFormatterOptions>();
+            container.Register<ISkinManager, SkinManager>();
+            container.Register<DirectoryMapSettings, DirectoryMapSettings>();
+            container.Register<IPresentationSettingsTask, PresentationSettingsTask>();
+            container.Register<IWorkspace, Infrastructure.Workspace>();
+            container.Register<IWithWorkspaceLayout, Infrastructure.Workspace>();
+            container.Register<IWorkspaceLayout, WorkspaceLayout>();
+            container.Register<ITreeNodeContextMenuFactory, TreeNodeContextMenuFactory>();
+            container.Register<IIndividualMoleculesPresenter, IndividualMoleculesPresenter>();
```
rwmcintosh, line 140: It might be possible to create a static, simple presenter with no dependencies, like a NullIndividualMoleculesPresenter or something. Then when it's used it could be checked against that static instance to see if the tab should be displayed.

rwmcintosh (changes requested):

src/PKSim.UI.Starter/ApplicationStartup.cs
georgeDaskalakis: @georgeDaskalakis true :)

src/PKSim.UI.Starter/ApplicationStartup.cs
georgeDaskalakis: OK I am looking into it

georgeDaskalakis:
@msevestre this is ready for review :)

src/PKSim.Core/ICoreUserSettings.cs
```diff
@@ -57,5 +57,10 @@ public interface ICoreUserSettings : OSPSuite.Core.ICoreUserSettings
       ///    Path of the template database containing the template building block
       /// </summary>
       string TemplateDatabasePath { get; set; }
+
+      /// <summary>
+      ///    Default grouping mode for grouping presenter
+      /// </summary>
+      ParameterGroupingModeId DefaultParameterGroupingMode { get; set; }
```
msevestre, line 64: Why is this movd here? What does it have to do with Core Settings? This is purely presentational. Explain

src/PKSim.Presentation/Presenters/Individuals/CreateIndividualPresenter.cs
```diff
@@ -92,9 +90,58 @@ public override void WizardNext(int previousIndex)
...
+         _view.NextEnabled = PresenterAt(IndividualItems.Settings).CanClose && indexThatWillHaveFocus != IndividualItems.Expression.Index;
+         _view.OkEnabled = CanClose;
+         _view.SetControlEnabled(IndividualItems.Expression, PresenterAt(IndividualItems.Settings).IndividualCreated);
+         _view.SetControlEnabled(IndividualItems.Parameters, PresenterAt(IndividualItems.Settings).IndividualCreated);
+      }
+   }
+   }
+
+//try moving to the Starter namespace
+   public class CreateIndividualPresenterForMoBi : AbstractCreateIndividualPresenter
```
msevestre, line 120: I am wondering why this is not just a simple flag that says Expression supported or not? RIght now it's an exact copy of the other code without the last tab. All those expressions such as
```
_view.SetControlEnabled(IndividualItems.Expression, PresenterAt(IndividualItems.Settings).IndividualCreated);
         _view.SetControlEnabled(IndividualItems.Parameters, PresenterAt(IndividualItems.Settings).IndividualCreated);
`` can be rewritten to use sthg like all except first

File: src/PKSim.Presentation/Presenters/Individuals/CreateIndividualPresenter.cs [hunk 5]
```diff
@@ -92,9 +90,58 @@ public override void WizardNext(int previousIndex)
...
+   }
+
+      public override void CreateIndividual()
+      {
+         if (PresenterAt(IndividualItems.Settings).IndividualCreated) return;
+         //reset commands before generating a new individual
+         _macroCommand.Clear();
+         PresenterAt(IndividualItems.Settings).CreateIndividual();
+         if (Individual == null) return;
+         PresenterAt(IndividualItems.Parameters).EditIndividual(Individual);
```
msevestre, line 143: Exact same code.

src/PKSim.Presentation/Presenters/Individuals/CreateIndividualPresenter.cs
```diff
@@ -92,9 +90,58 @@ public override void WizardNext(int previousIndex)
...
+      public CreateIndividualPresenter(ICreateIndividualView view, ISubPresenterItemManager<IIndividualItemPresenter> subPresenterItemManager,
+         IDialogCreator dialogCreator,
+         IBuildingBlockPropertiesMapper propertiesMapper, IObjectBaseDTOFactory buildingBlockDTOFactory)
+         : base(view, subPresenterItemManager, dialogCreator, propertiesMapper, buildingBlockDTOFactory, IndividualItems.All)
+      { }
+
+      protected override void UpdateControls(int indexThatWillHaveFocus)
+      {
+         UpdateViewStatus();
+         _view.NextEnabled = PresenterAt(IndividualItems.Settings).CanClose && indexThatWillHaveFocus != IndividualItems.Expression.Index;
```
msevestre, line 111: here there is a condition on the last entry...

src/PKSim.Presentation/Presenters/Individuals/CreateIndividualPresenter.cs
```diff
@@ -92,9 +90,58 @@ public override void WizardNext(int previousIndex)
...
+      public CreateIndividualPresenterForMoBi(ICreateIndividualView view, ISubPresenterItemManager<IIndividualItemPresenter> subPresenterItemManager,
+         IDialogCreator dialogCreator,
+         IBuildingBlockPropertiesMapper propertiesMapper, IObjectBaseDTOFactory buildingBlockDTOFactory)
+         : base(view, subPresenterItemManager, dialogCreator, propertiesMapper, buildingBlockDTOFactory, IndividualItems.AllExceptExpression)
+      {}
+      
+      protected override void UpdateControls(int indexThatWillHaveFocus)
+      {
+         UpdateViewStatus();
+         _view.NextEnabled = PresenterAt(IndividualItems.Settings).CanClose;
```
msevestre, line 131: the condition from above was removed. So next is enabled on the last tab, which does not make sense.

src/PKSim.UI.Starter/ApplicationStartup.cs
```diff
@@ -54,24 +74,82 @@ private static IContainer initializeForStartup(IShell shell)
...
+            container.Register<IReportPartToToolTipPartsMapper, ReportPartToToolTipPartsMapper>();
+            container.Register<IIndividualDefaultValueUpdater, IndividualDefaultValuesUpdater>();
+            container.Register<IIndividualSettingsDTOToOriginDataMapper, IndividualSettingsDTOToOriginDataMapper>();
+            container.Register<IParameterToParameterDTOMapper, ParameterToParameterDTOMapper>();
+            container.Register<ISubPopulationToSubPopulationDTOMapper, SubPopulationToSubPopulationDTOMapper>();
+            container
+               .Register<IParameterValueVersionToCategoryParameterValueVersionDTOMapper,
+                  ParameterValueVersionToCategoryParameterValueVersionDTOMapper>();
+            container.Register<ICalculationMethodToCategoryCalculationMethodDTOMapper, CalculationMethodToCategoryCalculationMethodDTOMapper>();
+            container.Register<IIndividualToIIndividualSettingsDTOMapper, IndividualToIIndividualSettingsDTOMapper>();
```
msevestre, line 114: WHAT. ARE WE DOING HERE?

msevestre (changes requested):
@georgeDaskalakis @rwmcintosh 
I am not sure what we are trying to do here. But registering 50 components by hand, duplicating lots of code just to hide one tab seems like this is not the right approach.
Happy to have a chat to discuss what to do here but in the current state, I cannot approve

src/PKSim.Presentation/Presenters/Individuals/CreateIndividualPresenter.cs
rwmcintosh: The idea was to reduce the number of registrations required. We didn't want to register the presenter for Expressions and all its dependencies if it's not going to be shown.

src/PKSim.Core/ICoreUserSettings.cs
rwmcintosh: It was in IUserSettings, but it was moved so that IUserSettings didn't need to be registered. A better spot for it is probably IPresentationUserSettings, which is registered in the starter. 

This property is required by ParameterGroupsPresenter.

src/PKSim.UI.Starter/ApplicationStartup.cs
rwmcintosh: What do you mean? Manually registering dependencies for the CreateIndividualPresenter.

Are you asking why the list is long, or asking a question about one line of code?

src/PKSim.Core/ICoreUserSettings.cs
msevestre: yep so it should be moved to IPresentationUserSettings

src/PKSim.Presentation/Presenters/Individuals/CreateIndividualPresenter.cs
msevestre: Hum. registration costs nothing. Instantiation is. Do you register separately based on the use case? I can't seem to see a difference in the registration code

src/PKSim.UI.Starter/ApplicationStartup.cs
```diff
@@ -54,24 +74,82 @@ private static IContainer initializeForStartup(IShell shell)
...
+            container.RegisterImplementationOf(shell as BaseShell);
             container.RegisterImplementationOf(shell);
+            container.RegisterImplementationOf(UserLookAndFeel.Default);
             container.AddRegister(x => x.FromType<CoreRegister>());
             container.AddRegister(x => x.FromType<InfrastructureRegister>());
             container.Register<ICreateExpressionProfilePresenter, CreateExpressionProfilePresenter>();
             container.Register<ICreateExpressionProfileView, CreateExpressionProfileView>();
+            container.Register<ICreateIndividualPresenter, CreateIndividualPresenterForMoBi>();
+            container.Register<ICreateIndividualView, CreateIndividualView>();
             container.Register<IExpressionProfileToExpressionProfileDTOMapper, ExpressionProfileToExpressionProfileDTOMapper>();
```
msevestre, line 91: So this is registered no matter what. I don't understand what you meant with the comment above @rwmcintosh

src/PKSim.UI.Starter/IndividualCreator.cs
```diff
@@ -0,0 +1,27 @@
...
+         var container = ApplicationStartup.Initialize(shell);
+
+         using (var presenter = container.Resolve<ICreateIndividualPresenter>())
+         {
+            presenter.Initialize();
+            var workspace = container.Resolve<IWorkspace>();
+            workspace.Project = new PKSimProject();
+            var mapper = container.Resolve<IIndividualToIndividualBuildingBlockMapper>();
+
+            return presenter.Create().IsEmpty() ? null : mapper.MapFrom(presenter.Individual);
```
msevestre, line 23: This is a bit funky. 
Is this better?
```
return presenter.Individual ==null ? null : mapper.MapFrom(presenter.Individual);
```
not sure. Feels more natural

src/PKSim.UI.Starter/UIStarterUserSettings.cs
```diff
@@ -5,14 +5,35 @@
...
+using OSPSuite.Presentation.Presenters;
 using OSPSuite.Presentation.Presenters.ParameterIdentifications;
 using OSPSuite.Presentation.Presenters.SensitivityAnalyses;
+using OSPSuite.Presentation.Services;
 using OSPSuite.Presentation.Settings;
 using PKSim.CLI.Core.MinimalImplementations;
 
 namespace PKSim.UI.Starter
 {
+   public class UIStarterPresentationSettingsTask : IPresentationSettingsTask
```
msevestre, line 19: Why cannot we use the real task?

tests/PKSim.UI.Tests/UI.StarterTests/ApplicationStartupTests.cs
```diff
@@ -37,23 +38,55 @@ public void the_presenter_should_be_resolved()
...
+
+         protected override void Because()
+         {
+            _mapper = _container.Resolve<IIndividualToIndividualBuildingBlockMapper>();
+         }
+
+         [Observation]
+         public void the_mapper_should_be_resolved()
+         {
+            _mapper.ShouldBeAnInstanceOf<IndividualToIndividualBuildingBlockMapper>();
```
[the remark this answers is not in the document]
rwmcintosh: The idea with these tests are to make sure that the objects needed to create the building block can be resolved when the container is initialized. If someone comes along later and adds a dependency to one of the presenters this is how we're going to know that we are not registering that new depenedency

src/PKSim.Core/PKSimStarterRegistrationConvention.cs
```diff
@@ -0,0 +1,51 @@
...
+   {
+      public override void Process(Type concreteType, IContainer container, LifeStyle lifeStyle)
+      {
+         if (!shouldRegisterType(concreteType))
+            return;
+
+         base.Process(concreteType, container, lifeStyle);
+      }
+
+      private bool shouldRegisterType(Type concreteType)
```
msevestre, line 25: Why is this done here and not in the register with the Include exclude. ?

rwmcintosh: Like this => ```scan.Include(shouldRegisterType);```?

src/PKSim.UI.Starter/PKSimStarterUserInterfaceRegister.cs
```diff
@@ -12,10 +13,37 @@ public override void RegisterInContainer(IContainer container)
...
             scan.AssemblyContainingType<UserInterfaceRegister>();
-            scan.WithConvention<PKSimStarterRegistrationConvention>();
+            scan.WithConvention<PKSimRegistrationConvention>();
+            scan.Include(shouldRegisterType);
          });
          container.Register<OSPSuite.UI.Services.IToolTipCreator, IToolTipCreator, ToolTipCreator>(LifeStyle.Transient);
          container.Register<IExitCommand, ExitCommand>();
       }
+
+      private bool shouldRegisterType(Type concreteType)
```
msevestre, line 23: no not like this. I don't understand why we would incluise the namespaces by name and not using type
=>IncudeNamespacingContaining<X>

msevestre: scan.IncludeNamespaceContainingType<>();

rwmcintosh: Oh, I see what you're getting at. concreteType.FullName will include the namespace. So anything in ExpressionProfile namespace will be registered by this. But also, anything where the class includes 'ExpressionProfile' in the class name that's not in the 'ExpressionProfile' namespace. For example, if there's a mapper.

I can revise for dedicated namespace scanning.

rwmcintosh: Yeah, this is registering a mixture of components from various namespaces some with class names that match, some which match the namespace.

PR #2548, opened 2023-02-24
Fixes #2534 Imported expression profile is not populated from the PKS…

### LINKED ISSUE #2534: Imported expression profile is not populated from the PKSim Database
@rwmcintosh commented on [Wed Dec 14 2022](https://github.com/Open-Systems-Pharmacology/MoBi/issues/839)

Instead, the empty version of the building block is transferred.

We probably need a button similar to PK-Sim to populate. This would then query PK-Sim to transfer the values

rwmcintosh:
…im Database

### DIFF
--- a/src/PKSim.Core/Mappers/MoleculeToQueryExpressionSettingsMapper.cs
+++ b/src/PKSim.Core/Mappers/MoleculeToQueryExpressionSettingsMapper.cs
@@ -32,13 +32,17 @@ namespace PKSim.Core.Mappers
 
       public QueryExpressionSettings MapFrom(ExpressionProfileBuildingBlock expressionProfileBuildingBlock)
       {
-         var containerGroups = expressionProfileBuildingBlock.
+         var expressionContainers = expressionProfileBuildingBlock.
             Where(x => x.Value.HasValue && x.HasExpressionName()).
-            Select(x => (categoryName:x.ContainerNameForRelativeExpressionParameter(), expressionValue:x.Value.Value)).ToList();
+            Select(expressionContainerInfoForExpressionParameter).ToList();
 
-         var expressionContainer = new List<ExpressionContainerInfo>(containerGroups.Select(x => new ExpressionContainerInfo(x.categoryName, _representationInfoRepository.ContainerInfoFor(x.categoryName).DisplayName, x.expressionValue)));
+         return new QueryExpressionSettings(expressionContainers, string.Empty, expressionProfileBuildingBlock.MoleculeName);
+      }
 
-         return new QueryExpressionSettings(expressionContainer, string.Empty, expressionProfileBuildingBlock.MoleculeName);
+      private ExpressionContainerInfo expressionContainerInfoForExpressionParameter(ExpressionParameter expressionParameter)
+      {
+         var containerName = expressionParameter.ContainerNameForRelativeExpressionParameter();
+         return new ExpressionContainerInfo(containerName, _representationInfoRepository.ContainerInfoFor(containerName).DisplayName, expressionParameter.Value.Value);
       }
 
       private ExpressionContainerInfo mapFrom(string containerName, IParameter relExpParameter)
--- a/src/PKSim.Core/Mappers/MoleculeToQueryExpressionSettingsMapper.cs
+++ b/src/PKSim.Core/Mappers/MoleculeToQueryExpressionSettingsMapper.cs
@@ -10,7 +10,7 @@ namespace PKSim.Core.Mappers
    public interface IMoleculeToQueryExpressionSettingsMapper
    {
       QueryExpressionSettings MapFrom(IndividualMolecule molecule, ISimulationSubject simulationSubject, string moleculeName);
-      QueryExpressionSettings MapFrom(ExpressionProfileBuildingBlockUpdate expressionProfile);
+      QueryExpressionSettings MapFrom(ExpressionProfileBuildingBlock expressionProfileBuildingBlock);
    }
 
    public class MoleculeToQueryExpressionSettingsMapper : IMoleculeToQueryExpressionSettingsMapper
@@ -30,15 +30,15 @@ namespace PKSim.Core.Mappers
          return new QueryExpressionSettings(expressionContainer, molecule.QueryConfiguration, moleculeName);
       }
 
-      public QueryExpressionSettings MapFrom(ExpressionProfileBuildingBlockUpdate expressionProfile)
+      public QueryExpressionSettings MapFrom(ExpressionProfileBuildingBlock expressionProfileBuildingBlock)
       {
-         var containerGroups = expressionProfile.ExpressionParameters.
-            Where(x => x.OriginalValue.HasValue && x.IsExpression()).
-            Select(x => (categoryName:x.ContainerNameForRelativeExpressionParameter(), expressionValue:x.OriginalValue.Value)).ToList();
+         var containerGroups = expressionProfileBuildingBlock.
+            Where(x => x.Value.HasValue && x.HasExpressionName()).
+            Select(x => (categoryName:x.ContainerNameForRelativeExpressionParameter(), expressionValue:x.Value.Value)).ToList();
 
          var expressionContainer = new List<ExpressionContainerInfo>(containerGroups.Select(x => new ExpressionContainerInfo(x.categoryName, _representationInfoRepository.ContainerInfoFor(x.categoryName).DisplayName, x.expressionValue)));
 
-         return new QueryExpressionSettings(expressionContainer, string.Empty, expressionProfile.MoleculeName);
+         return new QueryExpressionSettings(expressionContainer, string.Empty, expressionProfileBuildingBlock.MoleculeName);
       }
 
       private ExpressionContainerInfo mapFrom(string containerName, IParameter relExpParameter)
--- a/src/PKSim.Core/Mappers/MoleculeToQueryExpressionSettingsMapper.cs
+++ b/src/PKSim.Core/Mappers/MoleculeToQueryExpressionSettingsMapper.cs
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
 using PKSim.Core.Model;
 using PKSim.Core.Repositories;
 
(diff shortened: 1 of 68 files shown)


src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -26,23 +27,43 @@ public static void SetPercentile(this IParameter parameter, double percentile)
...
 
       public static bool IsIndividualMoleculeGlobal(this IParameter parameter) =>
          CoreConstants.Parameters.AllGlobalMoleculeParameters.Contains(parameter.Name);
 
+      private static bool isExpression(IWithName parameter)
+      {
+         return isGlobalExpression(parameter) || parameter.IsNamed(REL_EXP);
+      }
+
+      public static bool IsExpression(this ExpressionParameterUpdate parameter)
```
msevestre, line 53: what's this class?`ExpressionParameterUpdate`
Having a hard time to understand what it is. Somekind of parameter?
Also why do we need extension methods on **concrete** types?

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -216,4 +237,26 @@ public static void ScaleDistributionBasedOn(this IDistributedParameter parameter
...
+   public static class ExpressionParameterExtensions
+   {
+      public static string ContainerNameForRelativeExpressionParameter(this ExpressionParameterUpdate expressionParameter)
+      {
+         if (expressionParameter.IsGlobalExpression())
+            return CoreConstants.ContainerName.GlobalExpressionContainerNameFor(expressionParameter.Name);
+         
+         var objectPath = expressionParameter.Path;
+         var containerName = parentOfParent(objectPath);
+         var key = Equals(containerName, CoreConstants.Compartment.INTRACELLULAR) ? objectPath[objectPath.Count - 4] : containerName;
```
msevestre, line 250: Is this code that is/was in PKSim already?

msevestre (changes requested):
so I have probably missed a PR somewhere but I have no idea what those ExpressionBuildingBlockUpdate and ExpressionParameterUpdate are. 

They are name like function but are actually class? Where are those defined?

rwmcintosh:
> so I have probably missed a PR somewhere but I have no idea what those ExpressionBuildingBlockUpdate and ExpressionParameterUpdate are.
> 
> They are name like function but are actually class? Where are those defined?

This was a PR that you looked at last week and made the comment to remove DTO from the name. It's just the data needed to query the database and return updated values without all the extra building block stuff.

msevestre:
Oh yeah. I just mentioned that DTO wasn't a good name as it was in your question.  But didn't look any further to this PR. 
Well I don't like.tbis name.either. I can't even tell if this is a parameter or not (ExpressionParamterUpdate)

src/PKSim.Core/Model/ParameterExtensions.cs
rwmcintosh: The concrete types are defined in Core and all REL_EXP is defined in PK-Sim. I just made it this way to prevent moving these constant string into Core.

e.g. ```AllGlobalRelExpParameters```

src/PKSim.Core/Model/ParameterExtensions.cs
rwmcintosh: Not exactly, It follows the code from

```AllExpressionParametersFor``` from ```Individual.cs```
But this is using object paths instead of IParameters in containers.

File: src/PKSim.Core/Model/Individual.cs [hunk 3]
```diff
@@ -226,10 +226,10 @@ public override void UpdatePropertiesFrom(IUpdatable sourceObject, ICloneManager
       public virtual ICache<string, IParameter> AllExpressionParametersFor(IndividualMolecule molecule)
       {
          var cache = new Cache<string, IParameter>(onMissingKey: x => null);
-         var allExpressionParameters = GetAllChildren<IParameter>(x => x.IsExpression() && x.ParentContainer.IsNamed(molecule.Name));
+         var allExpressionParameters = GetAllChildren<IParameter>(x => x.HasExpressionName() && x.ParentContainer.IsNamed(molecule.Name));
```
msevestre, line 229: I don't understand why you removed this one. IsExpression does read better no?

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -26,28 +27,15 @@ public static void SetPercentile(this IParameter parameter, double percentile)
...
-         if (parameter == null)
-            return false;
-
-         return parameter.IsGlobalExpression() || parameter.IsNamed(REL_EXP);
-      }
-
       public static bool IsExpressionOrOntogenyFactor(this IParameter parameter)
       {
-         if (parameter.IsExpression())
+         if (parameter.HasExpressionName())
```
msevestre, line 38: I'd like this to be as it was before. Move the extension method, make it use an IWithName but don't rename it 
Or remove HasExpreesionName. This doesn't read like a sentence

src/PKSim.Core/Model/ParameterExtensions.cs
msevestre: I don't like this 4 without any comment ad to why this is like this. If it was like this already, it's poor from (me probably)

msevestre (changes requested):
Still reviewing

src/PKSim.Core/Mappers/MoleculeToQueryExpressionSettingsMapper.cs
```diff
@@ -28,6 +30,17 @@ public QueryExpressionSettings MapFrom(IndividualMolecule molecule, ISimulationS
          return new QueryExpressionSettings(expressionContainer, molecule.QueryConfiguration, moleculeName);
       }
 
+      public QueryExpressionSettings MapFrom(ExpressionProfileBuildingBlock expressionProfileBuildingBlock)
+      {
+         var containerGroups = expressionProfileBuildingBlock.
+            Where(x => x.Value.HasValue && x.HasExpressionName()).
```
msevestre, line 36: value is a nullable?

src/PKSim.Core/Mappers/MoleculeToQueryExpressionSettingsMapper.cs
```diff
@@ -28,6 +30,17 @@ public QueryExpressionSettings MapFrom(IndividualMolecule molecule, ISimulationS
...
+            Select(x => (categoryName:x.ContainerNameForRelativeExpressionParameter(), expressionValue:x.Value.Value)).ToList();
```
msevestre, line 37: let's call this container and not category?

msevestre: and why don't you create a method that reurnd an ExceprresionContainerInfo directly instead of loopng twice in the next line?

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -216,4 +204,26 @@ public static void ScaleDistributionBasedOn(this IDistributedParameter parameter
...
+         var key = Equals(containerName, CoreConstants.Compartment.INTRACELLULAR) ? objectPath[objectPath.Count - 4] : containerName;
+         if (expressionParameter.ContainerPath.Contains(CoreConstants.Organ.LUMEN))
+            key = CoreConstants.ContainerName.LumenSegmentNameFor(key);
+
+         return key;
+      }
+
+      private static string parentOfParent(ObjectPath objectPath)
+      {
+         return objectPath[objectPath.Count - 3];
```
msevestre, line 226: I am wondering about the 3 here. Is it possible that we have less? that 3 or 4 entries?

src/PKSim.Infrastructure/Serialization/Xml/Serializers/ApplicationSettingsXmlSerializer.cs
```diff
@@ -22,7 +21,7 @@ public override void PerformMapping()
 
       public override IApplicationSettings CreateObject(XElement element, SerializationContext serializationContext)
       {
-         return IoC.Resolve<IApplicationSettings>();
+         return serializationContext.Container.Resolve<IApplicationSettings>();
```
msevestre, line 24: good catch

src/PKSim.Presentation/Services/ExpressionProfileProteinDatabaseTask.cs
```diff
@@ -19,51 +18,81 @@ public interface IExpressionProfileProteinDatabaseTask
...
+         return _geneExpressionsDatabasePathManager.HasDatabaseFor(expressionProfile.Species);
+      }
+
+      public QueryExpressionResults QueryDatabase(ExpressionProfileBuildingBlock expressionProfileBuildingBlock)
+      {
+         using (_geneExpressionsDatabasePathManager.ConnectToDatabaseFor(expressionProfileBuildingBlock.Species))
+         using (var presenter = _applicationController.Start<IProteinExpressionsPresenter>())
+         {
+            presenter.InitializeSettings(_queryExpressionSettingsMapper.MapFrom(expressionProfileBuildingBlock));
+            return getQueryResults(presenter);
```
msevestre, line 73: That's good

src/PKSim.Core/Model/ParameterExtensions.cs
rwmcintosh: If it's named IsExpression then I orefer it is on iparameter rather than Iwithname. Just because an iwithname has the appropriate name doesn't make it an expression parameter. I'll keep the ones as-is and put them back for Iparameter

src/PKSim.Core/Mappers/MoleculeToQueryExpressionSettingsMapper.cs
rwmcintosh: Yes, it's a PathAndValueEntity like start value

src/PKSim.Core/Model/ParameterExtensions.cs
rwmcintosh: Well, this code is based on similar code that uses containers to find grand parent and great grand parent. I had to modify it to use ObjectPath arrays but in the code I found, there doesn't seem to be any checks for ParentContainer == null.

I can do a count check and return the root element if there's not enough.

msevestre (approved):

PR #2559, opened 2023-03-14
Fixes #2558 project created with v10< do not export properly

### LINKED ISSUE #2558: Projects created with <= V10 are sometimes not properly exported to snapshot with >= V11
The problem is: if an individual or a population in the V10 was containing a protein expression **and this individual/population was not converted (loaded once)** in V11 before exporting to the snapshot: the corresponding protein expressions are not exported to the snapshot.
E.g.
1. Open the project [V10.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/10969168/V10.zip) in V11. Do not open the individual or population BB.
2. Export it to snapshot: the snapshot does not contain any protein expression and crashes during loading [V11_1.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/10969206/V11_1.zip)
3. Export to snapshot again: now the snapshot is valid (because individual/population BBs were loaded and converted during the 1st snapshot export) [V11_2.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/10969224/V11_2.zip)

### Workaround for V11: 
when opening an older project (<= V10) - make sure to load all individual/populations once before exporting the project to snapshot.

msevestre:


src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -69,6 +69,9 @@ public override async Task<SnapshotProject> MapToSnapshot(ModelProject project)
             x.Description = SnapshotValueFor(project.Description);
          });
 
+         //first pass. We load all building blocks to ensure that there are no issues with the building blocks loading order
+         project.All<IPKSimBuildingBlock>().Each(_lazyLoadTask.Load);
```
msevestre, line 73: @Yuri05 This is the fix. It seems more natural to me than saving individual and expression profile in this order. We normally define expression profile first so having to save them backward looks like a hack to me

tests/PKSim.Tests/ProjectConverter/v11/When_exporting_the_expression_and_individual_v10_to_snapshot.cs
```diff
@@ -0,0 +1,36 @@
...
+
+      protected override void Because()
+      {
+         _snapshot = _projectMapper.MapToSnapshot(_project).Result;
+      }
+
+      [Observation]
+      public void should_have_exported_the_expression_for_the_individual()
+      {
+         _snapshot.ExpressionProfiles.ShouldNotBeNull();
```
msevestre, line 32: failed before the fix. Passes after.

Yuri05 (approved):

PR #2560, opened 2023-03-14
Fixes #2536 loading project from snapshot observed data

### LINKED ISSUE #2536: Loading project from snaphot: observed data which was previously added to a simulation is plotted but not added to the curves browser
Seems to happen with any project in 11.1
Example: [P1.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/10700161/P1.zip)

grafik

msevestre:


src/PKSim.Core/Services/IChartTask.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace PKSim.Core.Services
+{
+   public interface IChartTask
+   {
+      /// <summary>
+      /// updates all observed data defined in the simulation to the underlying charts of this simulation
+      /// </summary>
+      /// <param name="simulation">Simulation used to retrieve the charts to update</param>
+      /// <param name="project">Optional project. If undefined, it will be retrieved from the project retriever</param>
+      void UpdateObservedDataInChartsFor(Simulation simulation, IProject project = null);
```
msevestre: I moved this baby to core (where it should have been in the first place) so that I could use it from snapshot

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
```diff
@@ -305,6 +308,7 @@ public override async Task<ModelSimulation> MapToModel(SnapshotSimulation snapsh
          simulation.AddAnalyses(await populationAnalysesFrom(simulation, snapshot.PopulationAnalyses, snapshotContext));
 
          _simulationParameterOriginIdUpdater.UpdateSimulationId(simulation);
+         _chartTask.UpdateObservedDataInChartsFor(simulation, snapshotContext.Project);
```
msevestre, line 311: I am not sure WHY this is required now and not before. I am assuming that a change in core did not add the repository automatically. Maybe @georgeDaskalakis Can comment on this? At any rate, this should be done anyways IMO so this is not a bug per see in core

Yuri05 (approved):

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
georgeDaskalakis: @msevestre I am not quite sure....maybe we needed this because of the grouping of the chart lines  - let's briefly discuss this in the meeting today.

PR #2561, opened 2023-03-15
Fixes #2553 pksim changes molecule name

### LINKED ISSUE #2553: PK-Sim changes molecule name while importing a project from a snapshot file
In the attached original PK-Sim file "7E3_original.pksim5" (see zip file)  and the corresponding snapshot file "7E3.json", the molecule name in the metadata of the observed data is "7E3". After loading the project from the snapshot file, the name is 7000 (see file "7E3_importedFromSnapshot.pksim5").
   
original
imported

[files.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/10858790/files.zip)

msevestre:


src/PKSim.Core/Snapshots/Mappers/ExtendedPropertyMapper.cs
```diff
@@ -34,20 +35,60 @@ private Task<ModelExtendedProperty> mapExtendedProperty<T>(SnapshotExtendedPrope
...
+         if (type == typeof(bool))
+            return ExtendedPropertyType.Boolean;
+
+         return ExtendedPropertyType.String;
+      }
+
       public override Task<ModelExtendedProperty> MapToModel(SnapshotExtendedProperty snapshot, SnapshotContext snapshotContext)
       {
-         if (double.TryParse(snapshot.Value.ToString(), out var doubleResult))
-            return mapExtendedProperty(snapshot, option => double.Parse(option.ToString()), doubleResult);
```
msevestre, line 44: @Yuri05 This was the reason. 7E3 is a valid string as a double. Naming a molecule like a double is completely crazy so that's why it was assumed to be ok. But clearly not lol

src/PKSim.Core/Snapshots/Mappers/ExtendedPropertyMapper.cs
```diff
@@ -34,20 +35,60 @@ private Task<ModelExtendedProperty> mapExtendedProperty<T>(SnapshotExtendedPrope
...
+               return mapExtendedProperty(snapshot, double.Parse, double.Parse(valueAsString));
+            case ExtendedPropertyType.Boolean:
+               return mapExtendedProperty(snapshot, bool.Parse, bool.Parse(valueAsString));
+            //string
+            default:
+               return mapExtendedProperty(snapshot, option => option, valueAsString);
+         }
+      }
+
+      private Task<ModelExtendedProperty> mapExtendedPropertyBasedOnValue(ExtendedProperty snapshot, string valueAsString)
```
msevestre, line 83: @Yuri05 Old way

src/PKSim.Core/Snapshots/Mappers/ExtendedPropertyMapper.cs
```diff
@@ -34,20 +35,60 @@ private Task<ModelExtendedProperty> mapExtendedProperty<T>(SnapshotExtendedPrope
...
+         if (snapshotType.HasValue)
+            return mapExtendedPropertyBasedOnType(snapshot, snapshotType.Value, valueAsString);
 
+         return mapExtendedPropertyBasedOnValue(snapshot, valueAsString);
+      }
 
-         return mapExtendedProperty(snapshot, option => option.ToString(), snapshot.Value.ToString());
+      private Task<ModelExtendedProperty> mapExtendedPropertyBasedOnType(ExtendedProperty snapshot, ExtendedPropertyType snapshotType, string valueAsString)
+      {
+         switch (snapshotType)
```
msevestre, line 69: @Yuri05 New way.
This should fail if someone messes around with the snamsphot as we are parsing without trying

tests/PKSim.Tests/Core/ExtendedPropertyMapperSpecs.cs
```diff
@@ -144,4 +179,12 @@ protected override void CreateSnapshot()
          _snapshot = new ExtendedProperty {Description = "Description", ReadOnly = true, FullName = "Full Name", ListOfValues = new List<object> {4.5, 6.5}, Name = "Name", Value = 5.5};
       }
    }
+
+   public class When_mapping_snapshot_to_extended_property_string_that_can_be_parsed_as_double : When_mapping_snapshot_to_extended_property<string>
+   {
+      protected override void CreateSnapshot()
+      {
+         _snapshot = new ExtendedProperty {Description = "Description", ReadOnly = true, FullName = "Full Name", Name = "Name", Value = "7E3", Type = ExtendedPropertyType.String};
```
msevestre, line 187: This failed before the fix. This is good now
Note that old project will still be wrong so in the case of this issue, the type needs to be added by hand

msevestre: and we cannot assume it's string al the time because we are exporting also extended properties for PI who have numbers if I am not mistaken. Not sure if this is used or not so we could default to string in this case. Let me know

src/PKSim.Core/Snapshots/Mappers/ExtendedPropertyMapper.cs
```diff
@@ -21,10 +21,11 @@ public override Task<SnapshotExtendedProperty> MapToSnapshot(ModelExtendedProper
             snapshot.Name = extendedProperty.Name;
             snapshot.FullName = SnapshotValueFor(extendedProperty.FullName);
             snapshot.ReadOnly = SnapshotValueFor(extendedProperty.ReadOnly);
+            snapshot.Type = mapExtendedPropertyType(extendedProperty.Type);
```
Yuri05, line 24: where is actually the type of extended properties set initially?
E.g. wehn importing observed data: is the type of all the metadata set anywhere?

msevestre: Obs data is always string

Yuri05: ok

Yuri05 (approved):

Yuri05 (approved):

PR #2567, opened 2023-03-19
No more psv

msevestre:


msevestre:
@PavelBal @Yuri05 
I am tagging Pavel here just to make sure we are on the same page.
PSV created from PKSim should be completely empty UNLESS the user has set some parameters in the simulation directly, to which case it should be reflected into the PSV. Is this right?

Forget about the build failing. This is because it is using a local version of ospsuite.core

src/PKSim.Infrastructure/Services/NeighborhoodFinalizer.cs
```diff
@@ -20,29 +21,35 @@ public NeighborhoodFinalizer(IFlatNeighborhoodRepository flatNeighborhoodReposit
...
          neighborhood.FirstNeighbor = neighborFrom(flatNeighborhood.FirstNeighborId, organism);
          neighborhood.SecondNeighbor = neighborFrom(flatNeighborhood.SecondNeighborId, organism);
       }
 
-      private IContainer neighborFrom(int neighborId, Organism organism)
+      private void setNeighborsIn(NeighborhoodBuilder neighborhood, Organism organism)
+      {
+         var flatNeighborhood = _flatNeighborhoodRepository.NeighborhoodFrom(neighborhood.Name);
+
+         neighborhood.FirstNeighborPath = neighborPathFrom(flatNeighborhood.FirstNeighborId, organism);
```
msevestre, line 44: Here this is the real change. For builder, we cannot set the container but we need to use the path instead

msevestre (commented):
All other changes are because I merged changes from develop into OSMOSES so that we do not forget. Something that should have been done long ago IMO

src/PKSim.Core/Services/PKSimParameterStartValuesCreator.cs
```diff
@@ -1,56 +1,39 @@
...
-
-            return _defaultStartValues.WithName(simulation.Name);
-         }
-         finally
-         {
-            _defaultStartValues = null;
-         }
+         //default default parameter start values matrix
+         _defaultStartValues = _objectBaseFactory.Create<ParameterStartValuesBuildingBlock>();
+         updateSimulationParameters(simulation);
```
msevestre, line 35: @PavelBal @Yuri05 Those are now the only parameters updated. In case of a brand new simulation , this will be empty

src/PKSim.Core/Services/PKSimParameterStartValuesCreator.cs
```diff
@@ -75,40 +58,6 @@ private bool isChangedSimulationParameter(IParameter parameter)
...
-         var parameterStartValue = getStartValueFor(parameter);
-         if (parameterStartValue == null)
-            return;
-
-         if (parameter.Formula.IsExplicit())
-         {
-            var formula = _formulaFactory.RateFor(EXPRESSION_PARAMETERS, parameter.Formula.Name, _defaultStartValues.FormulaCache);
-            parameterStartValue.Formula = formula;
-            //There is a formula, make sure we use it. We set this flag to false to ensure that the formula will not be replaced with a constant formula
-            parameterStartValue.OverrideFormulaWithValue = false;
```
msevestre, line 101: OverrideFormulaWithValue  I am not sure that this flag should be kept now. This was because we were creating values on the fly for everything and wanted to keep the formula even for constant value. But in this case...we won't do that anymore
@Yuri05 I'd vote to remove this flag. Thought?

src/PKSim.Core/Services/PKSimParameterStartValuesCreator.cs
```diff
@@ -75,40 +58,6 @@ private bool isChangedSimulationParameter(IParameter parameter)
...
-         }
-
-         if (parameter.IsConstantParameter())
-         {
-            parameterStartValue.StartValue = parameter.Value;
-            //we reset the formula to null to ensure that the start value will be used when constructing the simulation
-            parameterStartValue.Formula = null;
```
msevestre, line 108: @Yuri05 Interestingly enough, constant parameter seem to take over formula anyways as I found out with @rwmcintosh last time. I am a a bit surprised by this behavior

PavelBal:
> PSV created from PKSim should be completely empty UNLESS the user has set some parameters in the simulation directly, to which case it should be reflected into the PSV. Is this right?

Yepp, this is how it should be (at least this is how we specified this, hopefully the concept works out ;) )

msevestre:
@PavelBal So let say i set a value of 10 Initial concentration in Bone intracellular. Before, this value was set in PSV. Now it should be in the expression profile correct?

PavelBal:
@msevestre  yes.

msevestre:
@all. I need to merge this to move forward. @Yuri05 Please review if you have time

PR #2572, opened 2023-03-21
Fixes #2570 Move IndividualToIndividualBuildingBlockMapper from prese…

### LINKED ISSUE #2570: Move IndividualToIndividualBuildingBlockMapper from presentation to core
I missed this during review but the interface should definitely be in core and I'd argue that the implementation as well
Right now, some DTO are being used to convert one model object to another which is very surprising (and scares me a bit...hopefully no nightmare tonight)

Can you move that to core? 

`ExpressionProfileToExpressionProfileBuildingBlockMapper` is where it belongs

rwmcintosh:
…ntation to core

### DIFF
--- a/src/PKSim.Core/Mappers/IndividualToIndividualBuildingBlockMapper.cs
+++ b/src/PKSim.Core/Mappers/IndividualToIndividualBuildingBlockMapper.cs
@@ -6,13 +6,10 @@ using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Domain.Services;
 using OSPSuite.Utility.Extensions;
 using PKSim.Assets;
-using PKSim.Core;
-using PKSim.Core.Mappers;
 using PKSim.Core.Model;
 using PKSim.Core.Repositories;
-using PKSim.Presentation.DTO.Mappers;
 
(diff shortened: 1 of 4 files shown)


src/PKSim.Core/Mappers/IndividualToIndividualBuildingBlockMapper.cs
```diff
@@ -50,14 +47,20 @@ public override IndividualBuildingBlock MapFrom(Individual input)
...
+            .Each(x => addCalculationMethodOriginDataToBuildingBlock(buildingBlock, x));
 
          buildingBlock.OriginData.ValueOrigin = input.OriginData.ValueOrigin.Clone();
 
          return buildingBlock;
       }
 
+      private void addCalculationMethodOriginDataToBuildingBlock(IndividualBuildingBlock buildingBlock, CalculationMethod calculationMethod)
+      {
+         var repInfo = _representationInfoRepository.InfoFor(RepresentationObjectType.CATEGORY, calculationMethod.Category);
```
rwmcintosh, line 60: I had to replace the dto mapper with the representationInfoRepository to gather the DisplayName property.
This is what the mapper would do when creating the DTO

PR #2582, opened 2023-04-02
Small change to make test pass again

msevestre:


msevestre:
@Yuri05 Just to be consistent with the way cattle was created

PR #2600, opened 2023-04-19
Fixes #2599 Module icon is missing in Module explorer

### LINKED ISSUE #2599: Module icon is missing in Module explorer
Loading the simulation S
The module icon is missing
image
[S.zip](https://github.com/Open-Systems-Pharmacology/MoBi/files/11263524/S.zip)

rwmcintosh:


rwmcintosh:
I think this is the right spot for this fix. Before export

src/PKSim.Core/Services/SimulationConfigurationTask.cs
```diff
@@ -71,7 +72,7 @@ IApplicationConfiguration applicationConfiguration
 
       public SimulationConfiguration CreateFor(Simulation simulation, bool shouldValidate, bool createAgingDataInSimulation)
       {
-         var module = _objectBaseFactory.Create<Module>().WithName(simulation.Name);
+         var module = _objectBaseFactory.Create<Module>().WithName(simulation.Name).WithIcon(ApplicationIcons.Module.IconName);
```
msevestre, line 75: I think the icon (name) should be set in the constructor. Do we do the same for all other building blocks?

rwmcintosh: indeed

PR #2625, opened 2023-05-17
Core updates

rwmcintosh:


rwmcintosh:
@msevestre FYI. I'm going to merge this right away. It's a simple core update with the ```GetExecutingAssembly``` call done in the configuration constructor.

PR #2633, opened 2023-05-23
update core

rwmcintosh:


src/PKSim.Core/Model/PKSimSpatialStructureFactory.cs
```diff
@@ -126,6 +127,6 @@ private void addModelStructureTo(IContainer container, OriginData originData, Mo
          }
       }
 
-      protected override SpatialStructure CreateSpatialStructure() => _objectBaseFactory.Create<PKSimSpatialStructure>();
+      protected override SpatialStructure CreateSpatialStructure() => _objectBaseFactory.Create<PKSimSpatialStructure>().WithName(DefaultNames.SpatialStructure);
```
rwmcintosh, line 130: In some cases we are counting on the service to create something with a meaningful default name, but ObjectBaseFactory can't do that

PR #2637, opened 2023-05-25
Fixes #2624. Fixes #2623

### LINKED ISSUE #2623: Disease state: disabled state in the individual view
Should be displayed as all other readonly fields - locked, but not "disabled" font
grafik

### LINKED ISSUE #2624: Create Individual: change Disease State selection label
Would suggest to change the label "Select" to "Disease state":
grafik

Like it's already the case in the Expression Profile
grafik

msevestre:


src/PKSim.UI/Views/DiseaseStates/DiseaseStateSelectionView.cs
```diff
@@ -80,9 +87,10 @@ public override void InitializeResources()
          base.InitializeResources();
          layoutControl.Margin = new Padding(0);
          layoutControl.Root.Padding = new DevExpress.XtraLayout.Utils.Padding(0);
-         layoutItemDiseaseState.Text = PKSimConstants.UI.Select.FormatForLabel();
+         layoutItemDiseaseState.TextVisible = false;
```
msevestre, line 90: @Yuri05 Hiding select for now. It's clear that you have to choose

PR #2641, opened 2023-05-26
2632 errorvalue cannot be null parameter name source

msevestre:


src/PKSim.Presentation/Presenters/Simulations/GlobalPKAnalysisPresenter.cs
```diff
@@ -167,15 +173,13 @@ public bool HasParameters()
 
       public bool CanCalculateGlobalPK()
       {
-         return firstSimulation.Compounds.Any(compound =>
-         {
-            var schemaItems = _protocolToSchemaItemsMapper.MapFrom(firstSimulation.CompoundPropertiesFor(compound).ProtocolProperties.Protocol);
-            return !isMultipleIV(schemaItems);
```
msevestre, line 173: @PavelBal @Yuri05 
Can someone explain to me why the condition to calculate global pk is to not have multiple IV?
Multiple oral is fine, mix of oral and iv is fine, but two multiple no? I don't get it

This also comes to the next error that @rwmcintosh is looking at where we have user defined stuff going on. Feels like this was not specified properly

src/PKSim.Presentation/Presenters/Simulations/GlobalPKAnalysisPresenter.cs
```diff
@@ -167,15 +173,13 @@ public bool HasParameters()
...
-         });
+         return firstSimulation.Compounds.Select(compound => firstSimulation.CompoundPropertiesFor(compound).ProtocolProperties.Protocol)
+            .Where(p => p != null)
```
msevestre, line 177: at any rate, I am checking that we have a protocol that's not null

src/PKSim.Core/Mappers/ProtocolToSchemaItemsMapper.cs
```diff
@@ -28,6 +29,9 @@ public ProtocolToSchemaItemsMapper(ISimpleProtocolToSchemaMapper simpleProtocolT
 
       public IReadOnlyList<SchemaItem> MapFrom(Protocol protocol)
       {
+         if (protocol == null)
```
msevestre, line 32: also added this to the mix. Why not be safe everywhere :)

rwmcintosh (approved):

src/PKSim.Presentation/Presenters/Simulations/GlobalPKAnalysisPresenter.cs
Yuri05: Yep, I think `!isMultipleIV` is wrong. Single_IV, Single_Oral and Multiple_Oral must be allowed, anything else probably not.
In the individual simulations:
* no global analysis is shown for IV_Oral_Mix, IV_UserDef_Mix etc.
* global analysis is shown for UserDef Single - which is wrong I think. All calculated values are then NaN, Infinity, Zero etc.
grafik grafik

So I think it makes sense to show global analysis only for the 3 scenarios:
* IV single
* Oral single
* Oral Multiple

msevestre: Sounds good. Agree with this and I'll make an update for this

PR #2644, opened 2023-05-27
Fixes #2643 when to show global pk analysis

### LINKED ISSUE #2643: When to show global PK Analysis
So I think it makes sense to show global analysis only for the 3 scenarios:

IV single
Oral single
Oral Multiple

msevestre:


src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -79,6 +79,13 @@ public interface IPKAnalysesTask : OSPSuite.Core.Domain.Services.IPKAnalysesTask
       ///    Starts the calculation of the Auc DDI Ratio for all compounds /> by switching off all other applications
       /// </summary>
       void CalculateDDIRatioFor(Simulation simulation);
+
+      /// <summary>
+      ///    Returns true if global pk-analysis can be calculated for the given <paramref name="protocol" /> otherwise false
+      /// </summary>
+      /// <param name="protocol"></param>
+      /// <returns></returns>
+      bool CanCalculateGlobalPKFor(Protocol protocol);
```
msevestre, line 88: Seems to be the right place, where we calculate global pk, to wonder if we can calculate global pk. Doing this in the presenter was not cohesive.

src/PKSim.Core/Services/PKAnalysesTask.cs
```diff
@@ -592,6 +593,21 @@ public void CalculateDDIRatioFor(Simulation simulation)
...
       }
 
+      public bool CanCalculateGlobalPKFor(Protocol protocol)
+      {
+         var schemaItems = _protocolToSchemaItemsMapper.MapFrom(protocol);
+         if (schemaItems == null || schemaItems.Count == 0)
+            return false;
+
+         //only one item, supported for IV or ORAL
+         var schemaItem = schemaItems[0];
```
msevestre, line 603: @Yuri05 As discussed in the issue

src/PKSim.Presentation/Presenters/Simulations/GlobalPKAnalysisPresenter.cs
```diff
@@ -175,12 +167,9 @@ public bool CanCalculateGlobalPK()
       {
          return firstSimulation.Compounds.Select(compound => firstSimulation.CompoundPropertiesFor(compound).ProtocolProperties.Protocol)
```
msevestre, line 168: arguably, this belongs also in the PKAnalysisTask

msevestre (commented):
Well I moved it where it belongs after all

Yuri05 (approved):

PR #2645, opened 2023-05-29
2627 no expression profiles created when exporting a simulation from pk sim to mobi

msevestre:


msevestre:
@Yuri05 @rwmcintosh 
Little bit of a hard one here. We realized that expression profile were not saved as reference into the individual building block, IN THE SIMULATION, when the project was loaded from file. This aims at solving this and also ensuring that we synchronize everything when we need to...I think

src/PKSim.Core/Model/SimulationFactory.cs
```diff
@@ -109,6 +109,7 @@ public Simulation CreateFrom(ISimulationSubject simulationSubject, IReadOnlyList
 
          _simulationBuildingBlockUpdater.UpdateUsedBuildingBlockInSimulationFromTemplate(simulation, simulationSubject, PKSimBuildingBlockType.SimulationSubject);
          _simulationBuildingBlockUpdater.UpdateMultipleUsedBuildingBlockInSimulationFromTemplate(simulation, compounds, PKSimBuildingBlockType.Compound);
+         _simulationBuildingBlockUpdater.UpdateMultipleUsedBuildingBlockInSimulationFromTemplate(simulation, simulationSubject.AllExpressionProfiles(), PKSimBuildingBlockType.ExpressionProfile);
```
msevestre, line 112: When we create a simulation, we need to all also all expression profiles as USED building blocks so that they are serialized as well into the project. It won't be available when loading v11 but so is life. We could try to add them if the simulation is green etc.. to be disucssed

src/PKSim.Core/Services/BuildingBlockParametersToSimulationUpdater.cs
```diff
@@ -65,9 +72,28 @@ public ICommand UpdateParametersFromBuildingBlockInSimulation(IPKSimBuildingBloc
...
          return updateCommands;
       }
 
+      /// <summary>
+      ///    We need to make sure that once the simulation has been updated with the building block, depending building blocks
+      ///    are also updated
+      ///    For instance, if we update the individual in the simulation, we will also update all expression profile (since
+      ///    expression profile are linked to the individual)
+      /// </summary>
+      private void synchronizeBuildingBlocks(IPKSimBuildingBlock templateBuildingBlock, Simulation simulation)
```
msevestre, line 86: BB => simulation. We need to make sure that the expression profile in the sim matches the template AFTER THE UPDATE

src/PKSim.Core/Services/SimulationParametersToBuildingBlockUpdater.cs
```diff
@@ -126,6 +125,8 @@ private ICommand synchronizeBuildingBlocks(IPKSimBuildingBlock templateBuildingB
          //Now that our expression profile are updated, we need to trigger the synchronization in all building blocks
          expressionProfilesToUpdate.Each(x => _expressionProfileUpdater.SynchronizeAllSimulationSubjectsWithExpressionProfile(x));
 
+         //last, synchronize the expression profile in the simulation as ell
+         _expressionProfileUpdater.SynchronizeExpressionProfilesUsedInSimulationSubjectWithSimulation(simulationSubject, simulation);
```
msevestre, line 129: Same idea with the opposite directon
We do Sim=>BB => Exp Profile in Sim

src/PKSim.Infrastructure/Services/LazyLoadTask.cs
```diff
@@ -113,14 +113,21 @@ private void loadSimulations(Simulation simulation)
...
             _simulationResultsLoader.LoadResultsFor(simulation.DowncastTo<IndividualSimulation>());
 
          else if (simulation.IsAnImplementationOf<PopulationSimulation>())
             _simulationAnalysesLoader.LoadAnalysesFor(simulation.DowncastTo<PopulationSimulation>());
 
+         //make sure each individual gets the expression profile defined in the simulation)  
+         var simulationSubject = simulation.BuildingBlock<ISimulationSubject>();
+       
+         //this can happen for an imported simulation
+         if (simulationSubject != null)
```
msevestre, line 128: When we load a simulation , we add the references  (individual in the simulation used expression profile in the simulation)

src/PKSim.Presentation/Presenters/ContextMenus/UsedBuildingBlockInSimulationTreeNodeContextMenuFactory.cs
```diff
@@ -45,6 +45,19 @@ protected override IContextMenu CreateFor(Simulation simulation, UsedBuildingBlo
       }
    }
 
+
+   public class UsedExpressionProfileInSimulationTreeNodeContextMenuFactory : UsedBuildingBlockInSimulationTreeNodeContextMenuFactory<ExpressionProfile>
```
msevestre, line 49: added if for debug purposes

msevestre (commented):
Please review and let me know :)

msevestre (commented):
Please review and let me know :)

src/PKSim.Core/Model/SimulationFactory.cs
Yuri05: let's create a new Issue for this

PR #2675, opened 2023-06-24
Fixes #2493

### LINKED ISSUE #2493: Adding simulation without results to "Compare results"
When I select several simulations and try to add them via drag&drop to a results comparison chart: if **at least 1 simulation** has no results (was not run yet) - nothing happens.
I would suggest to
* add all simulations which have results AND
* show the info message with the list of simulations which must be run first before adding
grafik

msevestre:


src/PKSim.Presentation/Presenters/Charts/IndividualSimulationComparisonPresenter.cs
```diff
@@ -66,10 +78,22 @@ public bool AnyCurves()
...
-            individualSimulationsFrom(droppedNodes).Each(addSimulationToChart);
-         else
+         var messages = new List<string>();
+         if (!containsIndividualSimulationNodes(droppedNodes))
+         {
             base.OnDragDrop(sender, e);
+            return;
+         }
+
+         individualSimulationsFrom(droppedNodes).Each(s =>
```
msevestre, line 88: @Yuri05 adding it to a message and showing the error. It was their all the time but the exception was swallowed lol

src/PKSim.Presentation/Presenters/Charts/IndividualSimulationComparisonPresenter.cs
```diff
@@ -201,7 +223,14 @@ private void editChart(IndividualSimulationComparison chart)
 
       private void updateResultsInChart()
       {
-         Chart.AllSimulations.Each(s => UpdateAnalysisBasedOn(s, s.DataRepository));
+         var isEmptyChart = !Chart.Curves.Any();
+         Chart.AllSimulations.Each(s =>
+         {
+            UpdateAnalysisBasedOn(s, s.DataRepository);
+            //if there are no curves in the chart, we are probably just creating it programatically
+            if (isEmptyChart)
```
msevestre, line 231: this is the creation based on the active simulation

msevestre (commented):
it's a huge pain to allow when there are NO results because we cache according to the DataRepo instance. It does not make much sense to start a comparison without an active simulation anyways
at least the one with results are added and you get a good message

Yuri05 (approved):

PR #2677, opened 2023-06-26
Fixes #2479 clone of compare results charts

### LINKED ISSUE #2479: Clone of "Compare Results" Charts
Would be really helpful (especially because we cannot display >1 plot in the results comparison at the moment)

msevestre:


src/PKSim.Presentation/Services/SimulationComparisonTask.cs
```diff
@@ -122,6 +128,19 @@ public bool Delete(IReadOnlyList<ISimulationComparison> simulationComparisons)
          return true;
       }
 
+      public async Task<ISimulationComparison> CloneSimulationComparision(ISimulationComparison simulationComparison)
+      {
+         _executionContext.Load(simulationComparison);
```
msevestre, line 133: @rwmcintosh @Yuri05 This is the meet of the actual implementation
I spent 2 hours trying using the various Cahrt task that we have but it's always made for one simulation and breaks as soon as you have multiple simulations which we have. here we are really recreating identical

rwmcintosh (approved):

PR #2678, opened 2023-06-27
Fixes #2676 add output to tme profile: default scaling

### LINKED ISSUE #2676: Add output to TIme Profile analysis population
Shouldn't the default scaling be Log for all output of dimension concentration?

image

msevestre:


msevestre:
@Yuri05 Trivial. Not sure why we never implemented anything like this

PR #2731, opened 2023-09-05
Fixes #2717 run nightly only if changes

### LINKED ISSUE #2717: Run nightly builds only if there were any commits to the repo in the last 24 hours
Implementation suggestion from ChatGPT :)

=========================================================

To achieve your goal of running the AppVeyor project build on a schedule (every 24 hours), but only if there were any repository changes (commits) in the last 24 hours, you'll need to use a combination of scheduled builds and conditional checks in your `appveyor.yml` configuration file. Here's how you can set it up:

1. **Scheduled Build Configuration:**
   Set up a scheduled build that runs every 24 hours. You can do this by using the `schedule` configuration in your `appveyor.yml` file.

```yaml
schedule:
  - cron: "0 */24 * * *"
```

This configuration will trigger a build every 24 hours at the start of the hour (e.g., 00:00, 01:00, etc.).

2. **Conditional Check for Repository Changes:**
   To check whether there were any repository changes (commits) in the last 24 hours before triggering the build, you'll need to use the `git` command within the `before_build` or `init` section of your `appveyor.yml` file.

```yaml
init:
  - ps: |
      $lastCommitDate = git log -1 --format=%ci
      $currentDate = Get-Date -Format "yyyy-MM-dd HH:mm:ss"
      $timeDiff = New-TimeSpan -Start $lastCommitDate -End $currentDate
      if ($timeDiff.TotalHours -lt 24) {
        Write-Host "Changes detected in the last 24 hours. Proceeding with the build."
      }
      else {
        Write-Host "No changes in the last 24 hours. Skipping the build."
        Exit-AppVeyorBuild
      }
```

In this example, the script calculates the time difference between the last commit's timestamp and the current time. If the time difference is less than 24 hours, the build proceeds. Otherwise, the script exits the build.

3. **Rest of Your Build Configuration:**
   After the conditional check, you can define the rest of your build configuration as needed.

Putting it all together, your `appveyor.yml` might look something like this:

```yaml
schedule:
  - cron: "0 */24 * * *"

init:
  - ps: |
      $lastCommitDate = git log -1 --format=%ci
      $currentDate = Get-Date -Format "yyyy-MM-dd HH:mm:ss"
      $timeDiff = New-TimeSpan -Start $lastCommitDate -End $currentDate
      if ($timeDiff.TotalHours -lt 24) {
        Write-Host "Changes detected in the last 24 hours. Proceeding with the build."
      }
      else {
        Write-Host "No changes in the last 24 hours. Skipping the build."
        Exit-AppVeyorBuild
      }

# Define the rest of your build configuration here
```

Remember to adjust the paths and commands according to your project's structure and requirements.

msevestre:


msevestre:
@Yuri05 FYI: Merging so that we can see if this is working

Yuri05:
> @Yuri05 FYI: Merging so that we can see if this is working

👀

Yuri05:
@msevestre The build was not triggered, not because of the 24h-check - but because the schedule in appveyor-nigthly.yml is not properly defined.
When I tried to trigger the build manually, I got this:
grafik

Yuri05:
I think we should just define the schedule in the AppVeyor project configuration (as before) and remove the `schedule` section from **appveyor-nightly.yml**
The rest of the new code seems to work (at least when I try it locally):

grafik

PR #2736, opened 2023-09-06
2709 metabolite does not get updated

msevestre:


src/PKSim.Core/Model/Simulation.cs
```diff
@@ -169,6 +169,12 @@ public virtual void AddUsedBuildingBlock(UsedBuildingBlock usedBuildingBlock)
 
       public virtual void AddReactions(ReactionBuildingBlock reactionBuildingBlock) => _allReactions.Add(reactionBuildingBlock);
 
+      public virtual void UpdateReactions(IEnumerable<ReactionBuildingBlock> reactionBuildingBlocks)
```
msevestre, line 172: We were having an interesting side effects where we kept adding to the list of reactions.

src/PKSim.Infrastructure/InfrastructureRegister.cs
```diff
@@ -50,6 +48,10 @@ public class InfrastructureRegister : Register
    {
       public static IContainer Initialize(bool registerContainerAsStatic = true)
       {
+         Thread.CurrentThread.CurrentCulture = new CultureInfo("en-US");
```
msevestre, line 51: @Yuri05 I moved this to the code that is called also for TESTS
Hopefully this will fix your test issues

src/PKSim.UI/Mappers/SimulationToSimulationReactionDiagramDTOMapper.cs
```diff
@@ -14,19 +13,24 @@ namespace PKSim.UI.Mappers
...
+         _diagramModelFactory = diagramModelFactory;
       }
 
-      public SimulationReactionDiagramDTO MapFrom(Simulation simulation)
+      public SimulationReactionDiagramDTO MapFrom(Simulation simulation, bool recreateDiagram)
       {
-         var reactionBuildingBlockDTO = simulation.Reactions.FirstOrDefault() ?? _reactionBuildingBlockCreator.CreateFor(simulation);
+         //Note: We do not use the building block in the simulation here as it MIGHT be out of date due to 
+         //simulation being configured
+         var reactionBuildingBlock = _reactionBuildingBlockCreator.CreateFor(simulation);
```
msevestre, line 29: this was a bug here. SOmetimes the reaction building block was OFF because the configuration had changed => outdated graph

msevestre: this was another bug lol

src/PKSim.Core/Model/Simulation.cs
Yuri05: interesting. never noticed that :)

Yuri05:
looks good for me. merging

Yuri05 (approved):

src/PKSim.Core/Model/Simulation.cs
msevestre: This is new. Until now, we only had one . In v12, it can be multiple (coming from different modules)

PR #2737, opened 2023-09-07
Fixes #2691 concentration parameter must be created with can be varie…

### LINKED ISSUE #2691: `Concentration` parameter must be created with `CanBeVaried = false`
At the moment it's not the case and Concentration parameter appears e.g. in the SensitivityAnalysis in MoBi and we have other side effects (e.g. when creating model C++ code in SimModel)

grafik

https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/99ce0a39f167aaf54f2e549fd72ce8d9266830be/src/PKSim.Core/Model/ParameterFactory.cs#L239-L248

msevestre:
…d false


msevestre:
@Yuri05 Also can be varied in population :)

Yuri05 (approved):

PR #2742, opened 2023-09-07
Fixes #2687 expression profile comparison

### LINKED ISSUE #2687: Expression profiles comparison: same ontogeny shown as different
Import the project from snapshot [Expr_Digoxin_Pediatrics.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/12162082/Expr_Digoxin_Pediatrics.zip) and compare 2 expression profiles.
Ontogeny is shown as different - but it's the same
(I also compared the ontogeny name with the text search in json - it's the same)
grafik

msevestre:


msevestre:
@Yuri05 Please check it out. RIght now, the TABLE PARAMETERS for ontogeny are visible
Whcih means that when we have two different ontogeny, CYP3A4 and CYP2D6, the parameters are also compared which does not make much sense. Not sure how to prevent that (this has nothing to do with this change but with the fact that we have two parameters with different formula). 
We could hide those parameters? Thoughts ?

src/PKSim.Infrastructure/ORM/Repositories/OntogenyRepository.cs
```diff
@@ -176,8 +181,10 @@ public DistributedTableFormula DataRepositoryToDistributedTableFormula(DataRepos
 
       public DistributedTableFormula OntogenyToDistributedTableFormula(Ontogeny ontogeny, string containerName)
       {
+         //We use a clone in this case as we do want to make sure we do not have two difference two references to the same table 
+         //this would lead otherwise as the same table being referenced by a parameter and the ontogeny and an issue will arise when loading
```
Yuri05, line 185: _we do not have two difference two references to the same table_ you mean "we do not have two references to the same table"?
_this would lead otherwise as the same table being referenced by_ something is missing here.

Yuri05:
> We could hide those parameters? Thoughts ?

hmm not sure. LEt's create another issue for this.
If we don't hide them: I think I just found another bug: when changing the ontogeny table in a simulation and commiting individual to BB: ontogeny profile remains unchanged in the expression profile BB.

Yuri05 (approved):
apart from the comment: looks good for me :)

msevestre (commented):
Updated for review

PR #2743, opened 2023-09-07
Fixes #2539

### LINKED ISSUE #2539: Export to Snapshot changes the display unit of observed data
E.g. import this snapshot: [01_Before_Import.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/10714212/01_Before_Import.zip)
Then directly export to snapshot: [02_After_Export.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/10714214/02_After_Export.zip)

In the original snapshot:
grafik

In the exported snapshot:
grafik

So the unit for this (and many other) datasets changed from **µg/l** to **mg/l.**
The values were recalculated accordingly, so they are still correct. 
Thus the only problem is the errorneous change of the display unit.

msevestre:


src/PKSim.Core/Snapshots/Mappers/DataColumnMapper.cs
```diff
@@ -67,10 +67,10 @@ public override async Task<ModelDataColumn> MapToModel(SnapshotDataColumn snapsh
          var dataInfo = await _dataInfoMapper.MapToModel(snapshot.DataInfo, snapshotContext);
          var dimension = dimensionFrom(snapshot);
          var dataColumn = dataInfo.Origin == ColumnOrigins.BaseGrid ? new BaseGrid(snapshot.Name, dimension) : new ModelDataColumn(snapshot.Name, dimension, dataRepository.BaseGrid);
+         //this needs to be set after DATA Info to be sure that we are using the value from the snapshot
+         dataColumn.DataInfo = dataInfo;
          dataColumn.DisplayUnit = displayUnitFor(dimension, snapshot.Unit);
          dataColumn.Values = valuesInBaseUnits(dataColumn, snapshot.Values);
-
-         dataColumn.DataInfo = dataInfo;
```
msevestre, line 73: very interesting side effects. DisplayUnit is saved in the DataInfo and not in the column. Therefor, by setting the display unit AND THEN the dataInfo, it was reset to the default (which in our case is not serialized in the snapshot)
So it is done the way it should de in snapshot. Unit belongs with the Data and not somewhere else. (WTH !!)

Yuri05 (approved):

PR #2748, opened 2023-09-08
update core and utility

rwmcintosh:


tests/PKSim.Tests/IntegrationTests/IndividualParameterBySpeciesRepositorySpecs.cs
```diff
@@ -44,7 +44,13 @@ public void all_parameter_paths_should_start_with_organism_or_neighborhoods()
 
    public class When_testing_if_a_parameter_is_common_for_all_species : concern_for_IndividualParametersNotCommonForAllSpeciesRepository
    {
-      private readonly IFlatContainerRepository _flatContainerRepository = IoC.Resolve<IFlatContainerRepository>();
```
msevestre, line 47: @Yuri05 HA HA . It took us a while to understand what was going on here.

Yuri05: what was the problem?

PR #2762, opened 2023-09-13
Fixes #2760

### LINKED ISSUE #2760: New path keywords not resolved in individual building block?
E.g. try to start PK-Sim with this database: [PKSimDB.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/12586008/PKSimDB.zip)

Here, the paths for `ORGAN|Volume (endothelium)` were changed using the `<NBH>` keyword.
This worked when tested in MoBi:
grafik

However, when trying to create an individual in PK-Sim the error comes

```
Unable to resolve path '..|Plasma|<NBH>|..|Interstitial|<NBH>|Surface area (plasma/interstitial)' 
for entity 'Volume (endothelium)' with path 'ROOT|Organism|Bone|Volume (endothelium)'
```
grafik

msevestre:


src/PKSim.Core/Services/BuildingBlockFinalizer.cs
```diff
@@ -61,6 +61,7 @@ private void finalizeIndividual(Individual individual)
          _individualPathWithRootExpander.AddRootToPathIn(individual);
          
          _keywordReplacerTask.ReplaceIn(individual);
+         _formulaTask.ExpandDynamicReferencesIn(individual.Root);
```
msevestre, line 64: @Yuri05 Here is the change

Yuri05: It works now - cool! 😂

PR #2783, opened 2023-09-19
Fixes #2770

### LINKED ISSUE #2770: V12: Ontogeny factors AGP/Albumin wrong in SIMULATIONS
E.g. create ICRP 0-years old male and a simulation with this individual.

In the individual BB ontogeny values are correct:
grafik

But in the created simulation ontogeny values are all set to 1:
grafik

msevestre:
@Yuri05 WIP but my local does not run tests anymore and I don't know why

### DIFF
--- a/src/PKSim.Core/Model/IndividualParameterSameFormulaOrValueForAllSpecies.cs
+++ b/src/PKSim.Core/Model/IndividualParameterSameFormulaOrValueForAllSpecies.cs
@@ -5,8 +5,5 @@
       public int ContainerId { get; set; }
       public string ContainerPath { get; set; }
       public string ParameterName { get; set; }
-      public bool IsSameFormula { get; set; }
-      public bool IsSameValue => !IsSameFormula;
-      public bool IsSame => IsSameFormula || IsSameValue;
    }
 }
--- a/src/PKSim.Core/Repositories/IIndividualParametersSameFormulaOrValueForAllSpeciesRepository.cs
+++ b/src/PKSim.Core/Repositories/IIndividualParametersSameFormulaOrValueForAllSpeciesRepository.cs
@@ -6,9 +6,7 @@ namespace PKSim.Core.Repositories
 {
    public interface IIndividualParametersSameFormulaOrValueForAllSpeciesRepository : IStartableRepository<IndividualParameterSameFormulaOrValueForAllSpecies>
    {
-      bool IsSameFormula(ParameterMetaData parameterMetaData);
-      bool IsSameValue(ParameterMetaData parameterMetaData);
-      (bool isSame, bool exists) IsSameFormulaOrValue(ParameterMetaData parameterMetaData);
+      bool IsSameFormulaOrValue(ParameterMetaData parameterMetaData);
       bool IsSameFormulaOrValue(IParameter parameter);
    }
 }
--- a/src/PKSim.Core/Services/MoleculeOntogenyVariabilityUpdater.cs
+++ b/src/PKSim.Core/Services/MoleculeOntogenyVariabilityUpdater.cs
@@ -71,13 +71,15 @@ namespace PKSim.Core.Services
 
       private void updatePlasmaProteinsOntogenyFor(Individual individual)
       {
-         _ontogenyRepository.SupportedProteins.Each(x => updatePlasmaProteinOntogenyTableFor(individual,x.TableParameterName, x.Name));
+         _ontogenyRepository.SupportedProteins.Each(x => updatePlasmaProteinOntogenyTableFor(individual, x.TableParameterName, x.Name));
       }
 
       private void updatePlasmaProteinOntogenyTableFor(Individual individual, string tableParameterName, string proteinName)
       {
          var parameter = individual.Organism.Parameter(tableParameterName);
-         if (parameter == null) return;
+         if (parameter == null) 
+            return;
+
          var tableFormula = _ontogenyRepository.PlasmaProteinOntogenyTableFormula(proteinName, individual.OriginData);
 
          updateOntogenyParameterTable(parameter, tableFormula);
@@ -116,16 +118,10 @@ namespace PKSim.Core.Services
       {
          //the formula may be null if no ontogeny was selected. In this case, we simply set a default formula of 1
          parameter.Formula = tableFormula ?? _formulaFactory.ValueFor(CoreConstants.DEFAULT_ONTOGENY_FACTOR, parameter.Dimension);
-      }
 
-      private void updateOntogenyParameter(IParameter parameter, double value)
-      {
-         //this can be the case if the parameter is defined in an expression profile
-         if (parameter == null)
-            return;
-
-         parameter.DefaultValue = value;
-         parameter.Value = parameter.DefaultValue.Value;
+         //in case of a tableFormula being set, make sure we also reset the default value of the parameter so that
+         //it will not look like the value was changed by the user
+         parameter.DefaultValue = tableFormula != null ? null : CoreConstants.DEFAULT_ONTOGENY_FACTOR;
       }
 
       public void UpdateAllOntogenies(Population population)
@@ -179,7 +175,7 @@ namespace PKSim.Core.Services
       private void updatePlasmaProteinOntogenyFor(Population population, IReadOnlyList<double> allAges, IReadOnlyList<double> allGAs, string parameterName, string proteinName)
       {
          var plasmaProteinOntogenyParameter = population.Organism.Parameter(parameterName);
-         if (plasmaProteinOntogenyParameter == null) 
+         if (plasmaProteinOntogenyParameter == null)
             return;
 
          var ontogenyFactors = new ParameterValues(_entityPathResolver.PathFor(plasmaProteinOntogenyParameter));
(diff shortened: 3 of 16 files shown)


src/PKSim.Core/Model/IndividualParameterSameFormulaOrValueForAllSpecies.cs
```diff
@@ -5,8 +5,5 @@ public class IndividualParameterSameFormulaOrValueForAllSpecies
       public int ContainerId { get; set; }
       public string ContainerPath { get; set; }
       public string ParameterName { get; set; }
-      public bool IsSameFormula { get; set; }
```
msevestre, line 8: @Yuri05 I never care if this is same formula or sane value. I care if it's the same or not (I think)

src/PKSim.Core/Repositories/IIndividualParametersSameFormulaOrValueForAllSpeciesRepository.cs
```diff
@@ -6,9 +6,7 @@ namespace PKSim.Core.Repositories
 {
    public interface IIndividualParametersSameFormulaOrValueForAllSpeciesRepository : IStartableRepository<IndividualParameterSameFormulaOrValueForAllSpecies>
    {
-      bool IsSameFormula(ParameterMetaData parameterMetaData);
-      bool IsSameValue(ParameterMetaData parameterMetaData);
-      (bool isSame, bool exists) IsSameFormulaOrValue(ParameterMetaData parameterMetaData);
+      bool IsSameFormulaOrValue(ParameterMetaData parameterMetaData);
```
msevestre, line 9: It's the same of if it's present in this list

src/PKSim.Core/Services/MoleculeOntogenyVariabilityUpdater.cs
```diff
@@ -116,16 +118,10 @@ private void updateOntogenyParameterTable(IParameter parameter, TableFormula tab
...
-      {
-         //this can be the case if the parameter is defined in an expression profile
-         if (parameter == null)
-            return;
 
-         parameter.DefaultValue = value;
-         parameter.Value = parameter.DefaultValue.Value;
+         //in case of a tableFormula being set, make sure we also reset the default value of the parameter so that
+         //it will not look like the value was changed by the user
+         parameter.DefaultValue = tableFormula != null ? null : CoreConstants.DEFAULT_ONTOGENY_FACTOR;
```
msevestre, line 124: So many issues with this one issue
So here, we need to make sure that the default value is reset for tableFormula otherwise it will be set to the VALUE when exported

src/PKSim.Core/Services/ParameterContainerTask.cs
```diff
@@ -76,11 +76,10 @@ void parameterValueModifier(ParameterMetaData parameterMetaData, IParameter para
          {
             //here we know that the parameter fulfills the criteria above. However, in case of a parameter that is used by all species, we need to verify that either the formula or the value is common 
             //to all species. Otherwise, we set the value to NaN
-            var (isSame, exists) = _sameFormulaOrValueForAllSpeciesRepository.IsSameFormulaOrValue(parameterMetaData);
```
msevestre, line 79: I don't thik it could have existedc before and NOT BE THE SAME (since by construction we either had isSameFormula or isSameValue and isSame as A || B
so it could not exist and not be the same. Is this correct?

src/PKSim.Infrastructure/ORM/Repositories/IndividualParametersSameFormulaOrValueForAllSpeciesRepository.cs
```diff
@@ -42,14 +42,21 @@ protected override void DoStart()
          foreach (var individualParameter in flatIndividualParametersSameFormulaOrValue)
          {
             var (containerId, parameterName, isSameFormula, _, _) = individualParameter;
+
+            //For ontogeny table parameter, we NEVER use the same formula as they are clearly individual specific parameters but are defined as constant in the database
+            if (parameterName.IsOneOf(ONTOGENY_FACTOR_TABLE, ONTOGENY_FACTOR_GI_TABLE, ONTOGENY_FACTOR_ALBUMIN_TABLE, ONTOGENY_FACTOR_AGP_TABLE))
```
msevestre, line 47: @Yuri05 I am not sure if this is the best way to exclude them, But they are returned as the same because the value in DB is 1 but this is clearly BS

src/PKSim.Infrastructure/ORM/Repositories/IndividualParametersSameFormulaOrValueForAllSpeciesRepository.cs
```diff
@@ -62,40 +69,21 @@ public override IEnumerable<IndividualParameterSameFormulaOrValueForAllSpecies>
...
-      }
+      public bool IsSameFormulaOrValue(ParameterMetaData parameterMetaData) => isSameFormulaOrValue(parameterMetaData);
 
       public bool IsSameFormulaOrValue(IParameter parameter)
       {
          Start();
          var path = _entityPathResolver.PathFor(parameter);
          var parameterSameFormulaOrValue = _allByParameterPath[path];
-         return parameterSameFormulaOrValue?.IsSame ?? false;
+         return parameterSameFormulaOrValue != null;
```
msevestre, line 79: it's present, then it's the same

src/PKSim.Infrastructure/ORM/Repositories/IndividualParametersSameFormulaOrValueForAllSpeciesRepository.cs
```diff
@@ -62,40 +69,21 @@ public override IEnumerable<IndividualParameterSameFormulaOrValueForAllSpecies>
...
-      private (bool isSameFormula, bool isSameValue, bool exists) isSameFormulaOrValue(ParameterMetaData parameterMetaData)
+      private bool isSameFormulaOrValue(ParameterMetaData parameterMetaData)
       {
          Start();
          var parameterSameFormulaOrValue = _allByContainerIdAndParameterName[(parameterMetaData.ContainerId, parameterMetaData.ParameterName)];
-
-         return parameterSameFormulaOrValue == null
-            ? (false, false, false)
-            : (parameterSameFormulaOrValue.IsSameFormula, parameterSameFormulaOrValue.IsSameValue, true);
+         return parameterSameFormulaOrValue != null;
```
msevestre, line 86: it's present, then it's the same

src/PKSim.Core/Model/IndividualParameterSameFormulaOrValueForAllSpecies.cs
Yuri05: ok

src/PKSim.Core/Services/ParameterContainerTask.cs
Yuri05: Theoretically yes. We will see 😄

src/PKSim.Infrastructure/ORM/Repositories/IndividualParametersSameFormulaOrValueForAllSpeciesRepository.cs
Yuri05: OK for me.
Technically those parameters are the same, so we need to exclude them explicitly somewhere.

Yuri05 (approved):

Yuri05:
@msevestre Good news: simulations with ontogeny, which produced different results compared to V11 are now correct and produce exactly the same results as in V11 👍 
Bad news: 3 other simulations, which were correct before the fix (compared to V11), now produce some differences. Investigating...
grafik

Yuri05:
OK, so at least one of the problems is the solver issue.

If I reduce the maximal solver step both in V11 and V12, then I get the same results in both versions (which are different from V11-results and different from V12-results with the default solver settings).
One thing I will try is this: https://github.com/Open-Systems-Pharmacology/OSPSuite.SimModel/issues/156

It is still not clear for me though, why this PR leads to the differences between V11 and V12 - I could not find the reason neither in the project nor in generated SimModel-XMLs

V11 vs. V12
 

V11 vs. V12 with solver HMax=0.001

Yuri05:
Another problem is the solver issue as well.
Again, reducing the max step of the ODE makes the curves identical.
Here; there are no tables, but a tablet with `Use As Suspension = false` (meaning it "jumps" from one GI-Segment to the next one).
An improvement would be to use the [root finding](https://github.com/Open-Systems-Pharmacology/OSPSuite.SimModel/issues/157) of CVODES and reformulate the definition of the tablet location using events.

I don't see it as urgent; the differences in the plasma curve are negligible; there are only small differences in few GI segments (s. below).
The only problem is that it destroys the batch comparison 😄 

### V11 vs. V12 (default solver settings)
grafik

### V11 vs. V12 with `HMax=0.01`
grafik

Yuri05:
For the remaining 3rd simulation with the differences: also here reducing the solver max step brings the sim results together.
So overall: the fix is good 👍

msevestre:
Awesome
I don't understand however how the same generated simmodel file produces different results. I thought our changes were completely equivalent

Yuri05:
Good question. This wonders me as well...

PR #2791, opened 2023-09-21
2786 pk sim refuses to compare two simulations

msevestre:


src/PKSim.Core/Services/PopulationAnalysisFlatTableCreator.cs
```diff
@@ -202,7 +201,35 @@ private static void storeReferenceGroupingItemOn(PopulationAnalysisDerivedField
 
       private string getReferenceExpression(string referenceSimulationName, string referenceLabel, string elseCase)
       {
-         return $"iif([{_simulationNameColumnName}]='{referenceSimulationName}', '{referenceLabel}', {elseCase})";
+         return $"iif([{_simulationNameColumnName}]='{escapeField(referenceSimulationName)}', '{referenceLabel}', {elseCase})";
+      }
+
+      private string escapeField(string fieldToEscape)
```
msevestre, line 207: ok somehow we need to escape special chars (and more). So I got this methods from chatGPT and it works :)

tests/PKSim.Tests/Core/PopulationAnalysisFlatTableCreatorSpecs.cs
```diff
@@ -141,7 +141,8 @@ protected override void Context()
       {
          base.Context();
 
-         var referenceSimulation = A.Fake<PopulationSimulation>().WithName("Sim");
+         //adding a character that should be escaped
+         var referenceSimulation = A.Fake<PopulationSimulation>().WithName("Sim's");
```
msevestre, line 145: it crashed without the fix !!

src/PKSim.Core/Services/PopulationAnalysisFlatTableCreator.cs
Yuri05: nice :)

PR #2797, opened 2023-10-01
2796 replace keywords also in reaction tags

msevestre:


msevestre:
keywords also replaced in tags. Not sure if we really need this but at least we have it now

Yuri05 (approved):

PR #2803, opened 2023-10-04
Fixes #2800

### LINKED ISSUE #2800: Do not export initial conditions with IsPresent false
If they are not created in the IC BB, then that is already not present.

msevestre:


msevestre:
@Yuri05 Trivial Merging. Please review

PR #2810, opened 2023-10-14
Fixes #2808 crreate individual wrongly memoized

### LINKED ISSUE #2808: Create Individual: wrongly memorized age
If I create an individual and change the age: the **next time** when I create an individual, I get the **previously set age** value as the default input.
E.g. I created a 1 yo ICRP individual.
In the next call: Age is set to 1 (instead of 30).
grafik

In the example above I would get an error, because Weight/Height still correspond to the 30yo individual.
But if e.g. my first Individual was 15yo, in the next call I would get as per default 15yo Individual with 30 yo weight/height and no error - dangerous.

msevestre:


src/PKSim.Presentation/Presenters/Individuals/IndividualSettingsPresenter.cs
```diff
@@ -116,7 +119,8 @@ private IReadOnlyList<DiseaseState> allDiseaseStatesFor(SpeciesPopulation popula
 
       public void PrepareForCreating()
       {
-         var individual = _defaultIndividualRetriever.DefaultIndividual();
+         //clone individual to ensure that we do not modify the default individual cached
+         var individual = _cloner.Clone(_defaultIndividualRetriever.DefaultIndividual());
```
msevestre, line 123: cloning the individual here feels like the best way to avoid any side effects. We probably have been dealing with this bug for quite a while

Yuri05 (approved):

PR #2818, opened 2023-10-23
Fixes #2761 change transport direction

### LINKED ISSUE #2761: Changing transpor direction in the expression is not applied in the simulation
1. Add a transporter protein with direction "Efflux"
2. Add a transport process using the new transporter protein to a molecule
3. Create a simulation and send to MoBi:

image

4. In the Expression Profile, change the direction of the transporter protein to "Influx". Simulation becomes "Red" and the comparison with building block correctly shows the difference in direction:
image

5. Right click on individual BB in the simulation and select "Update from BB"
6. Send the updated simulation to MoBi -> Direction of the transport is still "Efflux":

image

msevestre:
Fixes #2761

# Description

Probem was that it was seen as a non structural change (e.g. changing the expression profile structure). now I am also checking that all used expression profile have the same structure version

## Type of change

- [ ] This change requires documentation changes (link at least one documentation issue):

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/PKSim.Core/Services/SimulationBuildingBlockUpdater.cs
+++ b/src/PKSim.Core/Services/SimulationBuildingBlockUpdater.cs
@@ -42,8 +42,10 @@ namespace PKSim.Core.Services
       ///    Returns false if a simple parameter value update cannot be performed (i.e. structural change was made)
       /// </summary>
       /// <param name="templateBuildingBlock">Template building block as defined in repository</param>
-      /// <param name="usedBuildingBlock">Used building block (the one based on the template building block at a given time)</param>
-      /// <returns></returns>
+      /// <param name="usedBuildingBlock">
+      ///    Used building block (the one based on the template building block at a given time). It
+      ///    it assumed that the object is loaded (e.g. reference to building block can be used
+      /// </param>
       bool QuickUpdatePossibleFor(IPKSimBuildingBlock templateBuildingBlock, UsedBuildingBlock usedBuildingBlock);
 
       /// <summary>
@@ -61,7 +63,7 @@ namespace PKSim.Core.Services
       bool BuildingBlockSupportsQuickUpdate(IPKSimBuildingBlock templateBuildingBlock);
 
       /// <summary>
-      /// Returns whether a building block comparison is available for the building block
+      ///    Returns whether a building block comparison is available for the building block
       /// </summary>
       bool BuildingBlockSupportComparison(IPKSimBuildingBlock templateBuildingBlock);
    }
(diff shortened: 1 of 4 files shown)


src/PKSim.Core/Services/SimulationBuildingBlockUpdater.cs
```diff
@@ -140,7 +142,33 @@ public bool QuickUpdatePossibleFor(IPKSimBuildingBlock templateBuildingBlock, Us
          if (!BuildingBlockSupportsQuickUpdate(templateBuildingBlock))
             return false;
 
-         return templateBuildingBlock.StructureVersion == usedBuildingBlock.StructureVersion;
+         //not the same structure, easy return
+         var sameStructureVersion = templateBuildingBlock.StructureVersion == usedBuildingBlock.StructureVersion;
+         if (!sameStructureVersion)
+            return false;
+
+         //For individual, there is a special handling required as we need to also check that the used expression profile have hte same structure version
```
Yuri05, line 150: typo: "hte" => the

Yuri05, line 150: Do we need a special handling for _populations_ as well?

src/PKSim.Core/Services/SimulationBuildingBlockUpdater.cs
```diff
@@ -140,7 +142,33 @@ public bool QuickUpdatePossibleFor(IPKSimBuildingBlock templateBuildingBlock, Us
...
+         //It has to be available by construction
+         var usedIndividual = usedBuildingBlock.BuildingBlock as Individual;
+         //but we return false just in case :)
+         if (usedIndividual == null)
+            return false;
+
+         //let's compare the expression profile in each individuals and see if they are comparable
+         //not the same length of expression profile, structural change
+         if (usedIndividual.AllExpressionProfiles().Count != individualTemplate.AllExpressionProfiles().Count)
+            return false;
```
Yuri05, line 163: Is it actually possible, that `sameStructureVersion` in the code above is true, but we have a different number of expression profiles?

Yuri05:
If we use a PR template, we should also tick some options when creating a PR 😉

src/PKSim.Core/Services/SimulationBuildingBlockUpdater.cs
msevestre: nope. We dont allow commit or update for population

src/PKSim.Core/Services/SimulationBuildingBlockUpdater.cs
msevestre: I don't think so. you are right this check can be removed.

msevestre:
Agreed. I forgot to check. Still have to get used to doing this here

src/PKSim.Core/Services/SimulationBuildingBlockUpdater.cs
Yuri05: ok

PR #2837, opened 2023-11-23
Fixes #2836 export to invidual crashes

### LINKED ISSUE #2836: Export to invidualBB for MoBi crashes when the unit is not defined
This happens for dimensionsless object

msevestre:
Fixes #2836

# Description

Crashing during qualification run

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/PKSim.Core/Mappers/IndividualToIndividualBuildingBlockMapper.cs
+++ b/src/PKSim.Core/Mappers/IndividualToIndividualBuildingBlockMapper.cs
@@ -149,7 +149,7 @@ namespace PKSim.Core.Mappers
       private string originDataFormattedForDisplay(OriginDataParameter parameter)
       {
          var dimension = _dimensionRepository.DimensionForUnit(parameter.Unit);
-         var unit = dimension.FindUnit(parameter.Unit);
+         var unit = dimension.UnitOrDefault(parameter.Unit);
          return _formatter.Format(dimension.BaseUnitValueToUnitValue(unit, parameter.Value));
       }
 
--- a/src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
+++ b/src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
@@ -115,7 +115,7 @@ namespace PKSim.Core.Snapshots.Mappers
          var observedData = await observedDataFrom(snapshot.ObservedData, snapshotContext);
          observedData?.Each(repository => addObservedDataToProject(project, repository));
 
-         var allSimulations = await allSimulationsFrom(project, projectContext, snapshot.Simulations, snapshotContext);
+         var allSimulations = await allSimulationsFrom(projectContext, snapshot.Simulations, snapshotContext);
          allSimulations?.Each(simulation => addSimulationToProject(project, simulation));
 
          var allSimulationComparisons = await allSimulationComparisonsFrom(snapshot.SimulationComparisons, snapshotContext);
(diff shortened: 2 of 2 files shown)


src/PKSim.Core/Mappers/IndividualToIndividualBuildingBlockMapper.cs
```diff
@@ -149,7 +149,7 @@ private void addOriginDataToBuildingBlock(IndividualBuildingBlock buildingBlock,
       private string originDataFormattedForDisplay(OriginDataParameter parameter)
       {
          var dimension = _dimensionRepository.DimensionForUnit(parameter.Unit);
-         var unit = dimension.FindUnit(parameter.Unit);
+         var unit = dimension.UnitOrDefault(parameter.Unit);
```
msevestre, line 152: FindUnit return null if not found. Not good

src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -262,7 +262,7 @@ private void addQualificationPlanToProject(ModelProject project, Model.Qualifica
          project.GetOrCreateClassifiableFor<TClassifiableWrapper, TSubject>(subject);
       }
 
-      private async Task<IEnumerable<Model.Simulation>> allSimulationsFrom(ModelProject project, ProjectContext projectContext, Simulation[] snapshots, SnapshotContext snapshotContext)
+      private async Task<IReadOnlyList<Model.Simulation>> allSimulationsFrom(ProjectContext projectContext, Simulation[] snapshots, SnapshotContext snapshotContext)
```
msevestre, line 265: @Yuri05 This returns am empty list of a simulation cannot be loaded. I am wondering why we are not throwing at the end...

src/PKSim.Core/Mappers/IndividualToIndividualBuildingBlockMapper.cs
Yuri05: I don't understand the problem.
How can it happen that `dimension.FindUnit(parameter.Unit)` returns null, when the dimension was retireved for precisely this unit via `_dimensionRepository.DimensionForUnit(parameter.Unit)`?

src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
Yuri05: looks good for me AS IS.
Why would we throw an exception?

src/PKSim.Core/Mappers/IndividualToIndividualBuildingBlockMapper.cs
msevestre: Unit empty is not saved to the snapshot presumably. It’s in fact null when looking at the loaded object. Maybe the find unit could do something special for the case of dimensionsless. But right now it doesn’t support this.

src/PKSim.Core/Snapshots/Mappers/ProjectMapper.cs
msevestre: Well this is how I lost two hours. No error is thrown => run returns success => qualification runner says everything is good > reporting engine barfs with some weird progress bar error

I only found the problem because i started the run from PK-Sim itself  with breakpoints

PR #2915, opened 2024-06-18
2909 create dependency resolver project and publish an artifact

### LINKED ISSUE #2909: Create "dependency resolver project" and publish an artifact
Implement a new resolver project in PK-Sim. It should depend on PKSim.R only.

From there we can retrieve all the dependencies required for OSPSuite-R and publish them as an artifact.

https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2905

benjaperez1983:
Fixes #2909 

# Description

This PR is for creating a dependency project and publish the artifacts needed to run the last PK-SIM

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):


appveyor.yml
```diff
@@ -44,10 +44,18 @@ before_build:
   - nuget sources add -name core -source https://ci.appveyor.com/nuget/ospsuite-core
   - nuget restore
 
+
+after_build:
+  - cd src\ConsoleApp\bin\%CONFIGURATION%
+  - 7z a package.zip *
```
rwmcintosh, line 50: This should probably be only some of the files. We don't need the PDB files or many of the XML files (we will need OSPSuite.Dimensions.xml and maybe OSPSuite.PKParameters.xml. I don't think we need .config files either if they are there. This line probably zips up the .exe file too which we don't need.

rwmcintosh: Let's name the package something like `pk-sim-r-dependencies.zip`

src/ConsoleApp/Properties/AssemblyInfo.cs
rwmcintosh, line 1: Since you created are using SolutionInfo.cs can this be deleted?

src/ConsoleApp/PKDependencyApp.csproj
rwmcintosh, line 1: Similar to the package name let's rename this to say what it is. `PKSimRDependencyResolution` or something.

src/ConsoleApp/Program.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ConsoleApp
+{
+   internal class Program
+   {
+      static void Main(string[] args)
+      {
+         Console.WriteLine("Dummy Message");
```
rwmcintosh, line 13: We might as well make the text say what the purpose of the project is. Describe that there's no functionality, the project is used to resolve dependencies for R or something.

rwmcintosh (changes requested):
Just a few comments to polish it up a bit. Looks simple and we already know it works great.

rwmcintosh:
Fill in the #2909 (without space) there so that when the PR is merged, the issue is closed. Also all linking between PR and issue is made using that #

image

benjaperez1983:
@PavelBal  @Yuri05 @msevestre

rwmcintosh:
> Shall we run the build and publish the artifact with each merge? or whall we do it on the nightly build?

We would only see new Core updates propagated to OSPSuite-R after a nightly, rather than on-demand. I think it's not bad if Core updates in OSPSuite-R are nightly.

appveyor.yml
```diff
@@ -3,6 +3,9 @@ image: Visual Studio 2022
 
 version: '{build}'
 
+environment:
+  TARGET_FRAMEWORK: net472
```
rwmcintosh, line 7: Why is this needed?

rwmcintosh (approved):
Just a comment on the environment variable being set in AppVeyor. I'm surprised by  that. Can you comment?

appveyor.yml
msevestre: same question

src/PKSimRDependencyResolution/PKSimRDependencyResolution.csproj
```diff
@@ -0,0 +1,72 @@
...
+    <PackageReference Include="OSPSuite.FuncParser" Version="4.0.0.54" />
+    <PackageReference Include="OSPSuite.Infrastructure" Version="12.0.296" />
+    <PackageReference Include="OSPSuite.Infrastructure.Autofac" Version="12.0.296" />
+    <PackageReference Include="OSPSuite.Infrastructure.Import" Version="12.0.296" />
+    <PackageReference Include="OSPSuite.R" Version="12.0.296" />
+    <PackageReference Include="OSPSuite.Serializer" Version="3.0.0.1" />
+    <PackageReference Include="OSPSuite.SimModel" Version="4.0.0.56" />
+    <PackageReference Include="OSPSuite.Utility" Version="4.1.0.5" />
+    <PackageReference Include="Portable.BouncyCastle" Version="1.8.9" />
+    <PackageReference Include="Serilog" Version="2.10.0" />
```
msevestre, line 54: Are all of these required or can we somehow reference InfrasturacutreImport that depends on these already? Why is this package depending on so many DIRECTLY?

src/PKSimRDependencyResolution/Program.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ConsoleApp
+{
+   internal class Program
+   {
+      static void Main(string[] args)
+      {
+         Console.WriteLine("This console application includes all the dependencies needed for the PK-SIM project");
```
msevestre, line 13: `PK-Sim` !

src/PKSimRDependencyResolution/filelist.txt
```diff
@@ -0,0 +1,62 @@
+Autofac.pdb
```
msevestre, line 1: I don't get waht this file is? Why do we need this?
And can we generate it automatically instead of statically?

msevestre: why are we copying all xmls?

msevestre: why are we copying all pdbs?

msevestre (commented):
some questions remain

appveyor.yml
benjaperez1983: Sure, since yesterday, we updated the PKSimRDependencyResolution.cproj file, remember?
Before this update, the folder containing the files was "src\PKSimRDependencyResolution\bin\debug\"
But now, I can see (both the pipeline and local, of course) the folder has changed to "src\PKSimRDependencyResolution\bin\debug\net472\", so this is the folder to ZIP and distribute.
as net472 is the target framework, it did nos seem like a good idea to hardcode that value, but better set it on a variable.

src/PKSimRDependencyResolution/filelist.txt
benjaperez1983: It is the opposite in fact, this file contains all the files that we want to EXCLUDE, this is the way we achieve that:
- 7z a pk-sim-r-dependencies.zip -x@filelist.txt * --> "-x" does the trick.
The thing is, we need to exclude certain files, but we do not know in advance which ones, given that we need some .xms for example.
So the approach here is that if we were to include some dependencies in the future, and forgot to update this file, the worst case scenario will be having a zip with more files than we need, and not otherwise, which could cause not to have the files needed for it to run properly

src/PKSimRDependencyResolution/filelist.txt
Yuri05, line 1: Would rename the file to e.g. _**exclude**filelist.txt_ for clarity.

appveyor.yml
rwmcintosh: ah yes, I didn't see how it was being used below. It's fine for me, and I would also have accepted it as hard-coded in the path.

In other words, I don't have a strong preference.

rwmcintosh:
@Yuri05 @msevestre @benjaperez1983 @PavelBal still a question whether we should trigger this artifact nightly, or on each PR.

I think it belongs nightly...

Yuri05:
>I think it belongs nightly...

Agree.

src/PKSimRDependencyResolution/Program.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ConsoleApp
+{
+   internal class Program
+   {
+      static void Main(string[] args)
+      {
+         Console.WriteLine("This console application includes all the dependencies needed for PK-SIM");
```
Yuri05, line 13: **PK-Sim** instead of **PK-SIM** 😉

Yuri05:
What just comes into my mind: do we need separate packages for Windows and Linux?
At the moment the package contains Windows versions of all C++ libs.
Is probably ok for now, but latest when we have migrated PK-Sim Core/Infrastructure to .net8 and going to use the package under Linux as well, we would need 2 separate package versions.

rwmcintosh:
> What just comes into my mind: do we need separate packages for Windows and Linux?

I think so, but maybe for this iteration we can consume the package as built by AppVeyor and replace the native dll as we are now.

src/PKSimRDependencyResolution/excludedFiles.txt
```diff
@@ -0,0 +1,62 @@
+Autofac.pdb
+Autofac.xml
+BouncyCastle.Crypto.xml
+ConsoleApp.exe
+ConsoleApp.exe.config
+ConsoleApp.pdb
+DevExpress.Data.v21.2.xml
+filelist.txt
```
Yuri05, line 8: This should become **excludedFiles.txt** - right? :)

Yuri05:
Apart from the missing Linux build and https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/2915/files?w=1#r1672413911 - looks good for me :)

Yuri05:
and yes, as mentioned above, it should be actually part of the **nigthly** build

rwmcintosh:
> Apart from the missing Linux build and https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/2915/files?w=1#r1672413911 - looks good for me :)

The Linux build piece will be part of the OSPSuite-R PR, though, right, as it is now?

Yuri05:
> > Apart from the missing Linux build and https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/2915/files?w=1#r1672413911 - looks good for me :)
> 
> The Linux build piece will be part of the OSPSuite-R PR, though, right, as it is now?

hmm, maybe

Yuri05 (approved):

PR #2973, opened 2024-09-20
Fixes #2957 Endothelial surface area "Blood flow" - wrong values in the simulation

### LINKED ISSUE #2957: Endothelial surface area "Blood flow" - wrong values in the simulation?
The values AND formulas differ between the Individual and the actual simulation.

- Individual with "Endothelial surface area" calculation "Organ vascularization":
image

Uses the correct equation:
image

- Individual with "Endothelial surface area" calculation "Blood flow":
 
image

Uses the correct equation, mind that the values are different (as expected):

image

- Simulation using the individual with "Organ vascularization" - correct values (as in the individual):

image

## !!!!!
Simulation using the individual with "Blood flow" - WRONG formula AND the values are different from those in the individual!!

image

rwmcintosh:
Fixes #2957

# Description
Removing the update of ModelProperties

## Type of change

# How Has This Been Tested?

- [x] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):


rwmcintosh:
It's very tough for me to tell if this is the right approach. It appears that the concept of updating calculation methods from a new MoleculeProperties and old MoleculeProperties is not needed now.

I tried using different species and individuals with different calculation methods and it seems like we just want to use the newly created default MoleculeProperties in all these cases.

As pointed out during our dev meeting by @msevestre this code pre-dates Open Sourcing, so there's not much history available on it.

src/PKSim.Presentation/Presenters/Simulations/SimulationModelSelectionPresenter.cs
```diff
@@ -59,7 +59,7 @@ public IEnumerable<ModelConfiguration> AllModels()
       public void ModelSelectionChanging(ModelConfiguration newModelConfiguration)
       {
          ModelConfiguration = newModelConfiguration;
-         _modelPropertiesToUse = _modelPropertiesTask.Update(ModelProperties, _modelPropertiesTask.DefaultFor(newModelConfiguration, _originData), _originData);
+         _modelPropertiesToUse = _modelPropertiesTask.DefaultFor(newModelConfiguration, _originData);
```
rwmcintosh, line 62: In a nutshell here - do we need to combine ModelProperties from old to new?

tests/PKSim.Tests/IntegrationTests/ModelPropertiesTaskSpecs.cs
```diff
@@ -30,7 +30,7 @@ public class When_updating_model_properties : concern_for_ModelPropertiesTask
 
       protected override void Because()
       {
-         _updatedProperties = sut.Update(_modelProperties_4Comp, _modelProperties_2Pores, _originData);
+         _updatedProperties = _modelProperties_2Pores;
```
rwmcintosh, line 33: depending the need for the concept of Update, this test should go

src/PKSim.Presentation/Presenters/Simulations/SimulationModelSelectionPresenter.cs
msevestre: this is what was us causing the problem. Creating new one from scratch should be good I believe
Every entry that was set by hand will be reset but this seems to be ok
Do we need the .Update method at all? if not, I would just remove it

rwmcintosh: yeah, it's removed in this PR. I'll remove the tests for it and make the PR ready for review

PR #3108, opened 2025-04-21
3103 update ospsuitecore using async simulation runners

### LINKED ISSUE #3102: Remove Non-string resources in PKSim.Assets.Images
Image

These resources force us to reference package `System.Resources.Extensions` which as of version 9.0.2 has a bug where it cannot be loaded when the assembly targets .NET core and the runtime is .NET framework.

This will probably disappear when we update to use a newer .NET runtime, but we can remove this dependency by handling the images differently (not as embedded resources)

### LINKED ISSUE #3103: Update OSPSuite.Core using async simulation runners

benjaperez1983:
Fixes #3103 
Fixes #3102

# Description

Allow multiple running simulations

## Type of change

# How Has This Been Tested?

- [x] Integration tests
- [x] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):


src/PKSim.Core/Services/IndividualSimulationEngine.cs
```diff
@@ -19,6 +24,8 @@ public class IndividualSimulationEngine : SimulationEngine, IIndividualSimulatio
       private IProgressUpdater _progressUpdater;
       private readonly ISimulationResultsSynchronizer _simulationResultsSynchronizer;
       private readonly ISimulationToModelCoreSimulationMapper _modelCoreSimulationMapper;
+      
+
```
rwmcintosh: spacing

src/PKSim.Core/Services/IndividualSimulationEngine.cs
```diff
@@ -1,9 +1,14 @@
...
+using System.Collections.Concurrent;
+using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
+using ICSharpCode.SharpZipLib.Zip;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Mappers;
 using OSPSuite.Core.Domain.Services;
 using OSPSuite.Core.Events;
+using OSPSuite.Core.Journal;
```
rwmcintosh: I bet some of these are not used

src/PKSim.Core/Services/IndividualSimulationEngine.cs
```diff
@@ -117,5 +116,10 @@ private void simulationProgress(object sender, SimulationProgressEventArgs simul
       {
          _progressUpdater?.ReportProgress(simulationProgressArgs.Progress, PKSimConstants.UI.Calculating);
       }
+
+      public void StopSimulation(Simulation simulation)
+      {
+      
+      }
```
rwmcintosh: that should not be needed

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
```diff
@@ -1,18 +1,25 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Events;
+using OSPSuite.Core.Journal;
```
rwmcintosh: Probably not used

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
```diff
@@ -50,31 +58,59 @@ public InteractiveSimulationRunner(
...
-         _lazyLoadTask.Load(simulation);
+         return _cancellationTokenSources.TryGetValue(simulation, out var cts) &&
+                !cts.IsCancellationRequested;
+      }
 
-         if (outputSelectionRequired(simulation, selectOutput))
+      public Task RunSimulation(Simulation simulation, bool selectOutput)
+      {
+         return runSimulationAsync(simulation, selectOutput);
+      }
```
rwmcintosh: Use an expression body for really simple ones like this.

```
public Task RunSimulation(Simulation simulation, bool selectOutput) => runSimulationAsync(simulation, selectOutput);
```

src/PKSim.Presentation/Presenters/ContextMenus/SimulationContextMenu.cs
```diff
@@ -44,8 +46,11 @@ protected virtual IEnumerable<IMenuBarItem> DebugMenuFor(TSimulation simulation)
 
       private IEnumerable<IMenuBarItem> deleteMenuFor(TSimulation simulation)
       {
-         yield return DeleteMenuFor(simulation)
-            .AsGroupStarter();
+         yield return CreateMenuButton.WithCaption(PKSimConstants.MenuNames.Delete)
+
```
rwmcintosh: space

src/PKSim.Presentation/Presenters/ContextMenus/SimulationContextMenu.cs
```diff
@@ -73,9 +78,15 @@ protected IMenuBarItem ExportODEForRMenuItem(TSimulation simulation)
 
       private IEnumerable<IMenuBarItem> commonItemsForSimulations(TSimulation simulation)
       {
+      
+
```
rwmcintosh: space

rwmcintosh (changes requested):

rwmcintosh (approved):

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
```diff
@@ -50,31 +56,55 @@ public InteractiveSimulationRunner(
...
+
+            await _simulationRunner.RunSimulation(simulation, _simulationRunOptions, cts.Token);
+
+            addAnalysableToSimulationIfRequired(simulation);
+         }
+         finally
+         {
+            if (_cancellationTokenSources.ContainsKey(simulation))
+            {
+               _cancellationTokenSources[simulation].Dispose();
```
msevestre: @benjaperez1983 @rwmcintosh I don't get this code here
You are disposing the of the token source for the simulation
then you are trying to remove ti from the list and you dispose again?

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
```diff
@@ -99,5 +134,17 @@ private void addAnalysableToSimulationIfRequired(Simulation simulation)
...
          _simulationAnalysisCreator.CreateAnalysisFor(simulation);
       }
+
+      public void StopAllSimulations()
+      {
+         foreach (var simulation in _cancellationTokenSources.Keys.ToList())
+         {
+            if (_cancellationTokenSources.TryRemove(simulation, out var cts))
+            {
+               cts.Cancel();
```
msevestre: Are we making our life difficult with all of this? Any reason why you don't clear the list instead of removing them one by one like this? 
Also no event is being thrown here to notify cancel?

src/PKSim.Core/Events/ProgressDoneWithMessageEvent.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace PKSim.Core.Events
+{
+   public class ProgressDoneWithMessageEvent
```
msevestre: Why do we need this eventt in PKsim? don't we have something generic in core for this?

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
```diff
@@ -43,6 +49,10 @@ public class StatusBarPresenter : IStatusBarPresenter
       private readonly IApplicationConfiguration _applicationConfiguration;
       private int _numberOfReportsBeingCreated;
       public event EventHandler StatusChanged = delegate { };
+      private readonly ConcurrentDictionary<string, bool> _simulations = new ConcurrentDictionary<string, bool>();
```
msevestre: this variable naming is also off. This is not a list of simulations at all. There is no way I can understand the code by just reading it

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
```diff
@@ -43,6 +49,10 @@ public class StatusBarPresenter : IStatusBarPresenter
...
+
+      private int activeSimulations => _simulations.Count(x => x.Value);
```
msevestre: rename this variable. I am expecting to get a list of active simulations from the name, not a count
numberOfActiveSimulations or activeSimulationsCount or whatever

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
msevestre: why are we using a concurrent dictionnary here? isn't it all happening on the UI thread anyways?

msevestre (changes requested):
I don't think this is ready.

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
benjaperez1983: As I remember, on MoBi you requested this to be a concurrent dictionary.
I was using a regular dictionary but in our call you suggested to use a concurrent Dictionary.
Shall I revert this then?

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
benjaperez1983: Well if we clear the whole list, we will not be disposing any of the CancellationTokenSource objects. 
That leads to a resource leak.
We need to dispose them since the CancellationTokenSource uses wait handles internally (native OS resources).
Not disposing them could cause your app to slowly consume more resources over time.

Let me know what you think

src/PKSim.Core/Events/ProgressDoneWithMessageEvent.cs
benjaperez1983: well we could take this into core, but as now we do not have an event which relates to progress and has a message.

msevestre: Considerign there is EXACTLY the same class in Core and that a ProgresMessage with Even is not linked to MoBi or PKSim, it should be in Core. 
```
  public class ProgressDoneWithMessageEvent
  {
     public string Message { get; }

     public ProgressDoneWithMessageEvent(string message)
     {
        Message = message;
     }
  }
```

this is the code in MoBi. IMO that belongs in Core

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
msevestre: why is this marked as resolved if the code does not change and no explanation is given. Please please, do not resolve conversation if we do not agree on a resolution otherwise what's the point ? @benjaperez1983

msevestre: unresolving for now

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
msevestre: I mean this.  
```
 _cancellationTokenSources.each(x=>x.Dispose)
_cancellationTokenSources.Clear()
```

msevestre (changes requested):
@benjaperez1983 See comments. ALso as discussed previously, can we try to not resolve discussion if you don't provide comments as to why it makes sense as is or change the code...

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
msevestre: sounds good. Let's make it consistant with MoBi. I think the naming in MoBi is probably off then if it's copied from MoBi

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
benjaperez1983: It has changed and pushed. 
image

In fact this comment is showing as "Outdated".
Let me know if you see it since it was resolvec

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
benjaperez1983: Alright, will create a PR in mobi for this and for the event class as well.

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
benjaperez1983: Yeah I would not advice to do it this way for many reasons.
When we perform the **Keys.ToList()** we take a snapshot of the current keys.
then **TryRemove** to safely remove only if the key still exists.

If we apply your proposed solution we need to Cancel and Dispose:
_cancellationTokenSources.Each(x => x.Value.Cancel());
_cancellationTokenSources.Each(x => x.Value.Dispose());
_cancellationTokenSources.Clear();

So this will iterate multiple times, and most important it introduces a window between the cancel and the Dispose where other threads cuold be interacting with those entreas.

Finally the CLEAR removes everything at once, even those not yet canceled or disposed.

let me know what do you think.

msevestre: >_cancellationTokenSources.Each(x => x.Value.Cancel());
_cancellationTokenSources.Each(x => x.Value.Dispose());
_cancellationTokenSources.Clear();

So this will iterate multiple times, and most important it introduces a window between the cancel and the Dispose where other threads cuold be interacting with those entreas.


na. We would of course create one method that does both instead of interating twice. 

I am ok with the solution. It just looks like a lot of repetition especially with canceling one simulation. The way it's implemented, it feels that stop all should iterate and stop(simulation) instead of duplicating the exact same code. The only difference seems to be that a very generic event is being thrown when cancellng one simulation, but not when cancelling one simulation in a loop, which also is weird to me 

So ok to keep the current strucutr.e But can you think of a way to not repate the Tryremove, x.Dispose. x.cancel? and encapuslate it?

msevestre:
Good stuff!

PR #3118, opened 2025-04-28
Fixes #3117 Serialization should ignore unknown elements

### LINKED ISSUE #3117: Serialization should ignore unknown elements

rwmcintosh:
Fixes #3117

# Description
This is a follow up to changes made in OSPSuite.Serializer to allow unknown XML nodes to not crash the deserialization

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/PKSim.Infrastructure/Serialization/Xml/Serializers/UsedBuildingBlockXmlSerializer.cs
+++ b/src/PKSim.Infrastructure/Serialization/Xml/Serializers/UsedBuildingBlockXmlSerializer.cs
@@ -2,6 +2,7 @@ using System.Linq;
 using System.Xml.Linq;
 using PKSim.Core.Model;
 using OSPSuite.Core.Serialization.Xml;
+using OSPSuite.Serializer;
 
 namespace PKSim.Infrastructure.Serialization.Xml.Serializers
 {
@@ -34,7 +35,7 @@ namespace PKSim.Infrastructure.Serialization.Xml.Serializers
          var serializer = SerializerRepository.SerializerFor(buildingBlockNode);
 
          if (serializer == null)
-            return;
+            throw new SerializerNotFoundException(buildingBlockNode.Name.LocalName);
 
          usedBuildingBlock.BuildingBlock = serializer.Deserialize<IPKSimBuildingBlock>(buildingBlockNode, context);
       }
--- a/src/PKSim.Infrastructure/Serialization/Xml/Serializers/UsedBuildingBlockXmlSerializer.cs
+++ b/src/PKSim.Infrastructure/Serialization/Xml/Serializers/UsedBuildingBlockXmlSerializer.cs
@@ -33,6 +33,9 @@ namespace PKSim.Infrastructure.Serialization.Xml.Serializers
          var buildingBlockNode = usedBuildingBlockNode.Elements().First();
          var serializer = SerializerRepository.SerializerFor(buildingBlockNode);
 
+         if (serializer == null)
+            return;
+
          usedBuildingBlock.BuildingBlock = serializer.Deserialize<IPKSimBuildingBlock>(buildingBlockNode, context);
       }
    }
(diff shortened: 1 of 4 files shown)


src/PKSim.Infrastructure/Serialization/Xml/Serializers/UsedBuildingBlockXmlSerializer.cs
```diff
@@ -33,6 +33,9 @@ protected override void TypedDeserialize(UsedBuildingBlock usedBuildingBlock, XE
          var buildingBlockNode = usedBuildingBlockNode.Elements().First();
          var serializer = SerializerRepository.SerializerFor(buildingBlockNode);
 
+         if (serializer == null)
```
msevestre, line 37: hum... that's probably some serious shit going wrong here. Or why would you accept null? If this is a valid use casse, maybe comment why?

msevestre (changes requested):
unsure about this very unique change in UsedBBXmlSerializer

src/PKSim.Infrastructure/Serialization/Xml/Serializers/UsedBuildingBlockXmlSerializer.cs
rwmcintosh: This is when you have the element and you are trying to find the serializer. If the element is not a known type, then it should not throw an exception. That change is made also in OSPSuite.Serializer.

rwmcintosh: Or maybe we should leave this one as throwing an exception?

msevestre: I am just very surprised by the `specificity` of this one.  Is this the only place where we retrieve a serializer liek this?

msevestre: because heere we would have a UsedBuildingBlock WITHOUT a building block. That feels like it should NEVER happen

rwmcintosh: Yeah, that's right. We usually don't do this Elements.First() and then try to get the serializer. This is unique.

I guess this means that the order of the XML is important. Maybe an issue should be entered for that.

rwmcintosh: Why wouldn't there be 

```
Map(x => x.BuildingBlock);
```

rwmcintosh: ah, it's because the actual building block type is unknown I guess.

msevestre: yeah exactly. That's definitely a bit weird
The other options would be to make sure the tool does not crash if the ref to the BB is not found

rwmcintosh: We can just throw the serializer not found exception as it would have done before with old ospsuite.serializer.

rwmcintosh: I changed it to throw the Serializer not found exception. That's what would have happened in previous OSPSuite.Serializer if a serializer can't be found from the element.

PR #3136, opened 2025-05-30
update core

rwmcintosh:
This removes the dependency on msbuild so that we can skip it for CI builds. It seems like the latest github hosted runners have a mismatch where there's a newer SDK installed compared to the version of Visual Studio making it impossible to build unless you specify the exact SDK to use.

But that means that devs all need to have the same version of Visual Studio at all times and this must be updated in lock-step with the CI updates. :(

### DIFF
--- a/src/PKSim.UI/PKSim.UI.csproj
+++ b/src/PKSim.UI/PKSim.UI.csproj
@@ -56,14 +56,14 @@
   <ItemGroup>
      <PackageReference Include="Microsoft.Extensions.Logging" Version="9.0.2" />
     <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="9.0.2" />
-    <PackageReference Include="OSPSuite.Assets" Version="12.1.107" />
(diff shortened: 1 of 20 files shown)


src/PKSim.UI/PKSim.UI.csproj
```diff
@@ -10,6 +10,7 @@
     <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
     <NoWarn>1591</NoWarn>
     <Version Condition="'$(Version)' == ''">12.1.0</Version>
+    <GenerateResourceUsePreserializedResources>true</GenerateResourceUsePreserializedResources>
```
rwmcintosh, line 13: This consolidates how binary resources are serialized into the DLL. If you don't have it in csproj file then msbuild uses an older method of serializing binary data into dll and dotnet build will not work at all.

That means that no matter how you build (either through msbuild or dotnet build) you get the same result.

Yuri05 (approved):

PR #3147, opened 2025-06-19
Fixes #3132 Population simulations: wrong (geo)SD calculation?

### LINKED ISSUE #3132: Population simulations: wrong (geo)SD calculation?
I think this code 
https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/62fbbdd38a897dae5a78bad8bca944525f4ebff4/src/PKSim.Core/Services/StatisticalDataCalculator.cs#L34-L39 should be replaced by
```
 case StatisticalAggregationType.ArithmeticStandardDeviation: 
    yield return calculateArithmeticStandardDeviationFor(sortedResults);
    break; 
```

Same here: this code
https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/62fbbdd38a897dae5a78bad8bca944525f4ebff4/src/PKSim.Core/Services/StatisticalDataCalculator.cs#L51-L56 should be replaced by
```
 case StatisticalAggregationType.GeometricStandardDeviation: 
    yield return  calculateGeometricStandardDeviationFor(sortedResults); 
    break; 
```

Otherwise, the value for SD is shown as Mean-SD (for arithmetic) resp. GeoMean/GeoSD (for geometric), which is totally weird.

<details><summary>Example</summary>

[Test2_4PopPK_Omeprazole-Model.zip](https://github.com/user-attachments/files/20415914/Test2_4PopPK_Omeprazole-Model.zip)

Values shown for Cmax in PK-Sim:
Image

Values calculated in Excel: from exported individuals results
Image

Correct values are calculated but not returned in favor of $Mean \pm SD$ or $GeoMean \\* GeoSD$
Image

</details>

rwmcintosh:
Fixes #3132

# Description
Now you can request the deviations in two different ways from `StatisticalDataCalculator`. Default is two values as a range around the mean. Otherwise as a single value.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/PKSim.Core/Services/StatisticalDataCalculator.cs
+++ b/src/PKSim.Core/Services/StatisticalDataCalculator.cs
@@ -1,22 +1,29 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
+﻿using OSPSuite.Core.Extensions;
+using OSPSuite.Core.Maths;
 using OSPSuite.Utility.Extensions;
 using PKSim.Core.Model;
 using PKSim.Core.Model.PopulationAnalyses;
-using OSPSuite.Core.Extensions;
-using OSPSuite.Core.Maths;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using static PKSim.Core.Services.StatisticalDataCalculator;
 
 namespace PKSim.Core.Services
 {
    public interface IStatisticalDataCalculator
    {
-      IEnumerable<float[]> StatisticalDataFor(FloatMatrix sortedResults, StatisticalAggregation statisticalAggregation);
+      IEnumerable<float[]> StatisticalDataFor(FloatMatrix sortedResults, StatisticalAggregation statisticalAggregation, DeviationModes deviationMode = DeviationModes.Range);
    }
 
    public class StatisticalDataCalculator : IStatisticalDataCalculator
    {
-      public IEnumerable<float[]> StatisticalDataFor(FloatMatrix sortedResults, StatisticalAggregation statisticalAggregation)
+      public enum DeviationModes
+      {
+         Range,
+         Value
+      }
+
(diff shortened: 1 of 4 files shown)


Yuri05 (approved):

src/PKSim.Core/Services/StatisticalDataCalculator.cs
```diff
@@ -1,22 +1,29 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using static PKSim.Core.Services.StatisticalDataCalculator;
 
 namespace PKSim.Core.Services
 {
    public interface IStatisticalDataCalculator
    {
-      IEnumerable<float[]> StatisticalDataFor(FloatMatrix sortedResults, StatisticalAggregation statisticalAggregation);
+      IEnumerable<float[]> StatisticalDataFor(FloatMatrix sortedResults, StatisticalAggregation statisticalAggregation, DeviationModes deviationMode = DeviationModes.Range);
```
msevestre: I would not make this mode optional. The results are DIFFERENT based on the mode and it should be  required

src/PKSim.Core/Services/StatisticalDataCalculator.cs
```diff
@@ -1,22 +1,29 @@
...
    }
 
    public class StatisticalDataCalculator : IStatisticalDataCalculator
    {
-      public IEnumerable<float[]> StatisticalDataFor(FloatMatrix sortedResults, StatisticalAggregation statisticalAggregation)
+      public enum DeviationModes
+      {
+         Range,
```
msevestre, line 25: I think this could use a little explanation as to what the difference is ?

msevestre (changes requested):

PR #3148, opened 2025-06-20
#3145-expression-profiles-missing-in-mobi-unless-individual-is-opened…

### LINKED ISSUE #3145: Expression Profiles Missing in MoBi Unless Individual is Opened in PK-Sim (V12)
Description:

When working with the Observer course project in PK-Sim V12, we’ve identified an issue related to expression profiles not being loaded into MoBi unless the individual is explicitly opened in PK-Sim first.

Steps to Reproduce:
1) Open the project in PK-Sim (V12).
2) Without opening the individual, send the simulation to MoBi.
3) In MoBi, observe that no expression profiles are present.
4) Now, go back to PK-Sim, double-click to open the individual.
5) Without modifying anything, send the same simulation to MoBi.
6) This time, the expression profiles appear in MoBi as expected.

The same behavior occurs in PK-Sim V11 (i.e., expression profiles are only visible after opening the individual).

However, in V11, the simulation results in MoBi remain unaffected whether or not the individual was opened beforehand.

In contrast, in V12 the presence or absence of expression profiles significantly affects the simulation outcome.

Video Link:
https://esqlabs.sharepoint.com/:v:/s/MandS/EeF1DDNXsqZNilWRmpEHAdYBOAZrIVhHlQDjra6HNABgcw?e=wvCcpC

benjaperez1983:
Fixes #3145

# Description

We’ve identified an issue related to expression profiles not being loaded into MoBi unless the individual is explicitly opened in PK-Sim first.

## Type of change

# How Has This Been Tested?

- [x] Integration tests
- [ ] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/PKSim.Presentation/Presenters/Main/BuildingBlockExplorerPresenter.cs
+++ b/src/PKSim.Presentation/Presenters/Main/BuildingBlockExplorerPresenter.cs
@@ -24,26 +24,40 @@ using PKSim.Presentation.Nodes;
 using PKSim.Presentation.Regions;
 using PKSim.Presentation.Services;
 using PKSim.Presentation.Views.Main;
+using ILazyLoadTask = PKSim.Core.Services.ILazyLoadTask;
 using ITreeNodeFactory = PKSim.Presentation.Nodes.ITreeNodeFactory;
 
 namespace PKSim.Presentation.Presenters.Main
 {
    public interface IBuildingBlockExplorerPresenter : IExplorerPresenter,
       IListener<SwapBuildingBlockEvent>,
-      IListener<RenamedEvent>
+      IListener<RenamedEvent>,
+      IListener<ProjectLoadedEvent>
    {
    }
 
    public class BuildingBlockExplorerPresenter : ExplorerPresenter<IBuildingBlockExplorerView, IBuildingBlockExplorerPresenter>, IBuildingBlockExplorerPresenter
    {
       private readonly IObservedDataInExplorerPresenter _observedDataInExplorerPresenter;
+      private readonly ILazyLoadTask _lazyLoadTask;
 
-      public BuildingBlockExplorerPresenter(IBuildingBlockExplorerView view, ITreeNodeFactory treeNodeFactory, ITreeNodeContextMenuFactory treeNodeContextMenuFactory, IMultipleTreeNodeContextMenuFactory multipleTreeNodeContextMenuFactory, IBuildingBlockIconRetriever buildingBlockIconRetriever,
-         IRegionResolver regionResolver, IBuildingBlockTask buildingBlockTask, IToolTipPartCreator toolTipPartCreator, IProjectRetriever projectRetriever, IClassificationPresenter classificationPresenter, IObservedDataInExplorerPresenter observedDataInExplorerPresenter)
+      public BuildingBlockExplorerPresenter(IBuildingBlockExplorerView view, 
+         ITreeNodeFactory treeNodeFactory, 
+         ITreeNodeContextMenuFactory treeNodeContextMenuFactory, 
+         IMultipleTreeNodeContextMenuFactory multipleTreeNodeContextMenuFactory, 
+         IBuildingBlockIconRetriever buildingBlockIconRetriever,
+         IRegionResolver regionResolver, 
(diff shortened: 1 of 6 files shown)


src/PKSim.Presentation/Presenters/Main/BuildingBlockExplorerPresenter.cs
```diff
@@ -181,5 +195,12 @@ private void handleRenamedBuildingBlock(RenamedEvent renamedEvent)
 
          RefreshTreeAfterRename();
       }
+
+      public void Handle(ProjectLoadedEvent eventToHandle)
+      {
+         //This should be here since there are too many ways to open a simulation
+         //and this should be done before any simulation is loaded.
```
rwmcintosh: I would change this comment to read something like:
```
// Ensures that all individuals are loaded before any simulations can be lazy loaded.
```

rwmcintosh: I guess we should be stating the 'why' as well here. Loading the individual ensures that the expression profiles are created.

I'm not sure if the missing expression profiles are only a problem when converting from a project version that did not support them.

Is this step needed when the project is created with V12?

Yuri05: >Is this step needed when the project is created with V12?

I don't think so. Probably only for projects created with PK-Sim versions when expressions profiles were integrated into individuals and not defined as standalone building blocks.

rwmcintosh: Then we can probably move this to some post-conversion process.

msevestre: I do not understand this at all.. Why should the presneter be repsonsibiel of loading all individuals. This makes me very scared

msevestre (changes requested):
I do not believe that business logic should ever rely on event being thrown and or presenter to react to it. this should be done in the task if at all

src/PKSim.Presentation/Presenters/Main/BuildingBlockExplorerPresenter.cs
rwmcintosh: Maybe in `ProjectMetaDataToProjectMapper`. We know the project metadata version.

rwmcintosh: Ok, not in the mapper, but maybe in `ProjectPersistor` where it's called from. We have the version there too.

src/PKSim.Presentation/Services/ProjectTask.cs
```diff
@@ -325,6 +329,14 @@ private void openProjectFromFile(string projectFile, bool shouldStartWorker = tr
          void openProject()
          {
             _workspace.OpenProject(projectFile);
+
+            if (_workspace.Project.MetaDataVersion < ProjectVersions.V11)
```
benjaperez1983: I`ve added a property to the Project and stored the version there, since we did not find an easy way to get it otherwise.

Please let me know if this is fine.

Thanks!

rwmcintosh: Also, is this the correct version? When were expressions introduced as separate building blocks?

src/PKSim.Presentation/Services/ProjectTask.cs
```diff
@@ -325,6 +329,14 @@ private void openProjectFromFile(string projectFile, bool shouldStartWorker = tr
...
+            {
+               _workspace.Project.All<Individual>().Each(x =>
+               {
+                  _lazyLoadTask.Load(x);
```
msevestre: I don't undertand why we are not doing it when we load the project instead? Can we have a chat about this

msevestre (commented):
I don't understand what's going on

src/PKSim.Presentation/Services/ProjectTask.cs
benjaperez1983: Of course. let me know when available.

PR #3243, opened 2025-09-05
pksim uistarter is not usable by mobicli

### LINKED ISSUE #3240: PKSim UIStarter is not usable by MoBi.CLI
The reason is that `MoBi.CLI` registration does not register enough components to allow successful initialization of PKSim UIStarter.

We will need a CLIStarter that can be used by CLI services like `QualificationRunner`

### LINKED ISSUE #3242: Export Input markdown while rebuilding PK-Sim modules

rwmcintosh:
Fixes #3240
Fixes #3242

# Description
When MoBi.CLI dynamically loads and access CLI only methods from a PK-Sim assembly, the registration has to be a little different.

The MoBi container will not have any Windows Forms/DevExpress components registered as well as a UserSettings that is defined in a UI project. That is only registered when dynamicaly loading PK-Sim assemblies from a MoBi GUI process.

Also, export the building block markdown where required when rebuilding a PK-Sim module.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/PKSim.CLI.Core/Services/QualificationInputTask.cs
+++ b/src/PKSim.CLI.Core/Services/QualificationInputTask.cs
@@ -1,0 +1,85 @@
+﻿using OSPSuite.Core.Qualification;
+using OSPSuite.Core.Services;
+using OSPSuite.Utility;
+using OSPSuite.Utility.Extensions;
+using PKSim.Core.Model;
+using PKSim.Core.Services;
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using static OSPSuite.Assets.Error;
+using static OSPSuite.Core.Domain.Constants.Filter;
+using Project = PKSim.Core.Snapshots.Project;
+
+namespace PKSim.CLI.Core.Services
+{
+   public interface IQualificationInputTask
+   {
+      InputMapping[] ExportInputs(PKSimProject project, QualificationConfiguration configuration);
+
+      void ValidateInputs(Project snapshotProject, QualificationConfiguration configuration);
+   }
+   
+   public class QualificationInputTask : IQualificationInputTask
+   {
+      private readonly IOSPSuiteLogger _logger;
+      private readonly IMarkdownReporterTask _markdownReporterTask;
+
+      public QualificationInputTask(IOSPSuiteLogger logger, 
+         IMarkdownReporterTask markdownReporterTask)
+      {
+         _logger = logger;
+         _markdownReporterTask = markdownReporterTask;
+      }
+
+      public InputMapping[] ExportInputs(PKSimProject project, QualificationConfiguration configuration)
+      {
+         if (configuration.Inputs == null)
+            return Array.Empty<InputMapping>();
+         //            return Task.FromResult(Array.Empty<InputMapping>());
+
+         //TODO Enable parallel runs once https://github.com/Open-Systems-Pharmacology/OSPSuite.Utility/issues/26 is fixed
+         //  return Task.WhenAll(configuration.Inputs.Select(x => exportInput(project, configuration, x)));
+
+         return configuration.Inputs.Where(x => string.Equals(x.Project, project.Name)).Select(x => exportInput(project, configuration, x)).ToArray();
+      }
+
+      public void ValidateInputs(Project snapshotProject, QualificationConfiguration configuration)
+      {
+         configuration.Inputs?.Each(x =>
+         {
+            var buildingBlock = snapshotProject.BuildingBlockByTypeAndName(x.Type, x.Name);
+            if (buildingBlock == null)
+               throw new QualificationRunException(CannotFindBuildingBlockInSnapshot(x.Type.ToString(), x.Name, snapshotProject.Name));
+         });
+      }
+
+      private InputMapping exportInput(PKSimProject project, QualificationConfiguration configuration, Input input)
+      {
+         var buildingBlock = project.BuildingBlockByName(input.Name, input.Type);
+
+         var inputsFolder = configuration.InputsFolder;
+         var projectName = FileHelper.RemoveIllegalCharactersFrom(project.Name);
+         var buildingBlockName = FileHelper.RemoveIllegalCharactersFrom(input.Name);
+         var targetFolder = Path.Combine(inputsFolder, projectName, input.Type.ToString());
+         DirectoryHelper.CreateDirectory(targetFolder);
+
+         var fileFullPath = Path.Combine(targetFolder, $"{buildingBlockName}{MARKDOWN_EXTENSION}");
+
+         exportToMarkdown(buildingBlock, fileFullPath, input.SectionLevel).Wait();
+         _logger.AddDebug($"Input data for {input.Type} '{input.Name}' exported to '{fileFullPath}'", project.Name);
+
+         return new InputMapping
+         {
+            SectionId = input.SectionId,
+            SectionReference = input.SectionReference,
+            Path = relativePath(fileFullPath, configuration.OutputFolder)
+         };
+      }
+
+      private Task exportToMarkdown(object buildingBlock, string fileFullPath, int? inputSectionLevel) => _markdownReporterTask.ExportToMarkdown(buildingBlock, fileFullPath, inputSectionLevel);
+
+      private string relativePath(string path, string relativeTo) => FileHelper.CreateRelativePath(path, relativeTo, useUnixPathSeparator: true);
+   }
+}
--- a/src/PKSim.CLI.Core/Services/QualificationRunner.cs
+++ b/src/PKSim.CLI.Core/Services/QualificationRunner.cs
@@ -1,7 +1,4 @@
-﻿using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
-using OSPSuite.CLI.Core.RunOptions;
+﻿using OSPSuite.CLI.Core.RunOptions;
 using OSPSuite.CLI.Core.Services;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Services;
@@ -12,8 +9,10 @@ using PKSim.Core;
 using PKSim.Core.Model;
 using PKSim.Core.Services;
 using PKSim.Core.Snapshots.Services;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
 using static OSPSuite.Assets.Error;
-using static PKSim.Assets.PKSimConstants.Error;
 using Project = PKSim.Core.Snapshots.Project;
 using Simulation = PKSim.Core.Snapshots.Simulation;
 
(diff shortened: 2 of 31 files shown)


rwmcintosh:
I'd like to keep both open until they are both ready. They are linked
https://github.com/Open-Systems-Pharmacology/MoBi/pull/2089

src/PKSim.CLI.Core/Services/QualificationInputTask.cs
rwmcintosh, line 1: This whole file is just refactored from `QualificationRunner` so it can be used by the `PKSim.Starter`

src/PKSim.UI.Starter/CLIApplicationStartup.cs
rwmcintosh, line 1: When a PK-Sim service is exposed to MoBi that does not have any UI requirements, we have to register differently. In the UI starter, we register `IShell` and other Windows Forms objects that will not be resolvable when accessing a service from a CLI context.

src/PKSim.UI.Starter/CLIApplicationStartup.cs
```diff
@@ -0,0 +1,80 @@
...
+using PKSim.Presentation;
+using System;
+using System.Threading;
+using OSPSuite.Presentation;
+
+namespace PKSim.UI.Starter;
+
+public class CLIApplicationStartup : ApplicationStartup
+{
+   static IContainer _container;
```
rwmcintosh, line 23: What we are doing is building a separate container for resolving PK-Sim services in a CLI context. This container does not have any UI elements of course and does not require them from MoBi

src/PKSim.UI.Starter/UIApplicationStartup.cs
```diff
@@ -21,11 +19,13 @@
...
+using System;
+using System.Threading;
 using PresenterRegister = OSPSuite.Presentation.PresenterRegister;
 
 namespace PKSim.UI.Starter
 {
-   public class ApplicationStartup : Core.ApplicationStartup
+   public class UIApplicationStartup : ApplicationStartup
    {
       static IContainer _container;
```
rwmcintosh, line 30: Compared here, the container will have all the UI objects

src/PKSim.CLI.Core/Services/QualificationRunner.cs
```diff
@@ -24,36 +23,36 @@ public class QualificationRunner : QualificationRunner<Project, PKSimProject>
...
 
       protected override void SaveProjectContext(string projectFile) => _workspacePersistor.SaveSession(_workspace, projectFile);
 
       protected override void LoadProjectContext(PKSimProject project) => _workspace.Project = project;
 
-      protected override void ValidateInputs(Project snapshotProject, QualificationConfiguration configuration)
+      protected override void ValidateInputs(Project snapshotProject, QualificationConfiguration configuration) => 
+         _qualificationInputTask.ValidateInputs(snapshotProject, configuration);
+
+      protected override async Task<(PKSimProject, InputMapping[])> LoadProjectAndExportInputs(QualificationRunOptions runOptions, Project snapshot, QualificationConfiguration config)
```
msevestre, line 51: This is what I wanted to comment in the upstream PR.  `Export` feels weird in naming because I am expectign a side effects such as inputs being saved as a file
Also loading a project and returning some inputs are two complete different concerns. Why do we put them in one method? (just curious, no need to change)

src/PKSim.CLI.Core/Services/QualificationRunner.cs
```diff
@@ -125,10 +124,5 @@ private Simulation simulationFrom(Project snapshotProject, string simulationName
          var referenceSimulation = snapshotProject.Simulations?.FindByName(simulationName);
          return referenceSimulation ?? throw new QualificationRunException(CannotFindSimulationInSnapshot(simulationName, snapshotProject.Name));
       }
-
-      protected override Task ExportToMarkdown(object buildingBlock, string fileFullPath, int? inputSectionLevel)
```
msevestre, line 129: before this naming made sense. We are exporting to a file and return void (Task). No surprise here

src/PKSim.Core/Services/ProjectSnapshotToSimulationTransferMapper.cs
```diff
@@ -11,15 +11,15 @@
...
+public interface IProjectSnapshotToSimulationTransferMapper : IMapper<string, (SimulationTransfer transfer, PKSimProject project)>;
 
 public class ProjectSnapshotToSimulationTransferMapper(
    IJsonSerializer jsonSerializer,
    ISnapshotMapper snapshotMapper,
    ISimulationConfigurationTask simulationConfigurationTask,
    ISimulationToModelCoreSimulationMapper simulationMapper) : IProjectSnapshotToSimulationTransferMapper
 {
-   public SimulationTransfer MapFrom(string snapshotString)
+   public (SimulationTransfer transfer, PKSimProject project) MapFrom(string snapshotString)
```
msevestre, line 22: unexpected return type here for a mapper

src/PKSim.UI.Starter/CLIApplicationStartup.cs
msevestre, line 23: why static?

src/PKSim.CLI.Core/Services/QualificationRunner.cs
rwmcintosh: This is actually the case. The project will be created as before, and the input will be processed. That means markdown will be exported to the file system. 

The reason they are together now is that for MoBi, when we load a project from a snapshot, part of the process is recreating the PK-Sim project to export the PK-Sim module. The inputs specified in the qualification are relative to the PK-Sim project, so we have to process the markdown at that time.

At no point does the PK-Sim project exist beyond the duration of the module export so that's why the steps are now together.

src/PKSim.CLI.Core/Services/QualificationRunner.cs
rwmcintosh: We are still going to export the markdown in the `LoadProjectAndExportInputs`. Do you mean we should rename it `LoadProjectAndExportMarkdown` instead?

src/PKSim.UI.Starter/CLIApplicationStartup.cs
rwmcintosh: All the `ApplicationStartup` classes are the same. They subclass from a core class and implement static methods for bootstrapping. The static methods can only access static members.

src/PKSim.Core/Services/ProjectSnapshotToSimulationTransferMapper.cs
rwmcintosh: We can make a class to hold them, but it would be single-use, I think.

src/PKSim.CLI.Core/Services/QualificationRunner.cs
msevestre: ok so 
Project is loaded
Inputs are exported 
Project and inputs are returned. 
I think I am used to seeing a "file" somewhere when we export but I am assuming not that this will be into some config folder somewhere?

src/PKSim.CLI.Core/Services/QualificationRunner.cs
msevestre: I guess what's unexpected is "export" that does not take a file name but returns something and in fact it does export to file and still returns. This is not a typical signature that's why it tripped me

src/PKSim.Core/Services/ProjectSnapshotToSimulationTransferMapper.cs
msevestre: arguably a smulation transfer could also old a ref to the project it's coming from but would not be serialized. I am find with the tuple as return type but I see it in many of the code changes lately and I don't think (personal preference) that it helps with understanding what's going on, either as output or as input. I like it for `private` stuff because the context is clear but for public API, 🤢

src/PKSim.UI.Starter/SimulationTransferConstructor.cs
```diff
@@ -1,17 +1,29 @@
...
-      public static object CreateSimulationTransfer(string projectSnapshot)
-      {
-         var container = ApplicationStartup.Initialize();
-         var projectSnapshotToSimulationTransferMapper = container.Resolve<IProjectSnapshotToSimulationTransferMapper>();
-         var moBiExportTask = container.Resolve<IMoBiExportTask>();
+      var container = CLIApplicationStartup.Initialize();
+      var projectSnapshotToSimulationTransferMapper = container.Resolve<IProjectSnapshotToSimulationTransferMapper>();
+      var qualificationInputTask = container.Resolve<IQualificationInputTask>();
+
+      var (simulationTransfer, project) = projectSnapshotToSimulationTransferMapper.MapFrom(projectSnapshot);
```
msevestre, line 23: in a project snapshot there is only ever one simulation?

src/PKSim.UI.Starter/SimulationTransferConstructor.cs
```diff
@@ -1,17 +1,29 @@
...
 
-         var simulationTransfer = projectSnapshotToSimulationTransferMapper.MapFrom(projectSnapshot);
-         return moBiExportTask.Serialize(simulationTransfer);
-      }
+      var inputMappings = qualificationInputTask.ExportInputs(project, qualificationConfiguration);
+      
+      return (simulationTransfer, inputMappings);
```
msevestre, line 27: this also exports all the input that may or may not be used by the simulation. Why?

src/PKSim.CLI.Core/Services/QualificationRunner.cs
rwmcintosh: The `QualificationConfiguration` contains the output path arguments and that combines with the project name to give the full output path for the input markdown.

src/PKSim.UI.Starter/SimulationTransferConstructor.cs
rwmcintosh: In this case yes, because it's coming from a Module. PK-Sim builds the module for `SimulationTransfer` - only one simulation.

src/PKSim.UI.Starter/SimulationTransferConstructor.cs
rwmcintosh: It will only export those that are tied to the project by name. That's handled in `QualificationInputTask`, although we need to provide the configuration anyway because it contains the components of the path. We could also build the base path first here.

rwmcintosh: And the project only has the bb's required for the simulation. The qualification configuration could contain any inputs, but if they don't match the project name, they won't be exported.

PR #3248, opened 2025-09-09
Fixes #3246

### LINKED ISSUE #3163: 12.1 Confusing LockUI/ProgressBar behaviour when running simulations
In 12.1, when I start running an individual or population simulation, the rest of the UI is not locked anymore.
Now, if I open another simulation which **was not loaded before**, the calculation progress bar disappears.
The only possibility to reanimate it seems to start another simulation run in parallel.

This is new in 12.1 (in 12.0, the rest of the UI was locked - except the STOP button - till the calculation was completed).

### LINKED ISSUE #3246: 12.1 Simulation run fails => Run Buttons remain inactive
[simple_ErrorRun.zip](https://github.com/user-attachments/files/22237685/simple_ErrorRun.zip)
If a simulation run exists with error - Run buttons remain inactive (until the active simulation is changed)

Image

msevestre:
Fixes #3246
Fixes #3163

# Description
Running a simulation=> we cannot run in again unless we activate it again by switchign tabs for instance.
Now we have a way to superseed the STATE by using the parameter given

Also fixes #3163 by simply removing the progress bar when lazy loading an object. This does not make sense any more as ONLY one progress bar can be active at a time

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
+++ b/src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
@@ -276,18 +276,21 @@ namespace PKSim.Presentation.Presenters.Main
          updateSimulationItemsAccordingToSimulationState();
       }
 
-      private void updateSimulationItemsAccordingToSimulationState(bool isRunning = false)
+      /// <summary>
+      ///    Updates the state of all menu items related to simulation according to the current state of the simulation
+      /// </summary>
+      /// <param name="isSimulationRunning">If defined, super seeds the simulationState </param>
+      private void updateSimulationItemsAccordingToSimulationState(bool? isSimulationRunning = null)
       {
-         bool enabled = _simulationState.IsActivated && _enabled;
-         bool resultsEnabled = enabled && _simulationState.HasResult;
-         bool enablePopSimulationItems = enabled && !_simulationState.IsIndividual;
-         bool enableIndividualSimulationItems = enabled && _simulationState.IsIndividual;
-         bool enabledPKSimSimulationOnlyItems = enabled && !_simulationState.IsImported;
+         var enabled = _simulationState.IsActivated && _enabled;
+         var resultsEnabled = enabled && _simulationState.HasResult;
+         var enablePopSimulationItems = enabled && !_simulationState.IsIndividual;
+         var enableIndividualSimulationItems = enabled && _simulationState.IsIndividual;
+         var enabledPKSimSimulationOnlyItems = enabled && !_simulationState.IsImported;
+         var isRunning = enabled && (isSimulationRunning ?? _simulationState.IsRunning);
 
-         var isSimulationAlreadyRunning = (isRunning || _simulationState.IsRunning);
-         _menuBarItemRepository[MenuBarItemIds.Run].Enabled = enabled && !isSimulationAlreadyRunning;
-         _menuBarItemRepository[MenuBarItemIds.RunWithSettings].Enabled = enabled && !isSimulationAlreadyRunning; ;
-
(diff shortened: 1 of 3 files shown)


src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -276,18 +276,21 @@ private void disableSimulationItems()
          updateSimulationItemsAccordingToSimulationState();
       }
 
-      private void updateSimulationItemsAccordingToSimulationState(bool isRunning = false)
```
msevestre, line 279: this default value was not correct. because it looks like by default it's never running and since we were combining it with the state coming from the ISRunning extension method, it would alwyas default to this value which was wrong since it's updated AFTER the event are raised

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -276,18 +276,21 @@ private void disableSimulationItems()
...
+      /// </summary>
+      /// <param name="isSimulationRunning">If defined, super seeds the simulationState </param>
+      private void updateSimulationItemsAccordingToSimulationState(bool? isSimulationRunning = null)
+      {
+         var enabled = _simulationState.IsActivated && _enabled;
+         var resultsEnabled = enabled && _simulationState.HasResult;
+         var enablePopSimulationItems = enabled && !_simulationState.IsIndividual;
+         var enableIndividualSimulationItems = enabled && _simulationState.IsIndividual;
+         var enabledPKSimSimulationOnlyItems = enabled && !_simulationState.IsImported;
+         var isRunning = enabled && (isSimulationRunning ?? _simulationState.IsRunning);
```
msevestre, line 290: either we pass a flag and we use it if defined, OTHERWUSE we use the state. We do not combine these two flags because this will almost always be wrong

src/PKSim.Presentation/Presenters/Main/MenuAndToolBarPresenter.cs
```diff
@@ -336,11 +337,11 @@ public void Handle(SimulationResultsUpdatedEvent eventToHandle)
          updateSimulationItemsFor(eventToHandle.Simulation as Simulation);
       }
 
-      private void updateSimulationItemsFor(Simulation simulation, bool isRunning = false)
+      private void updateSimulationItemsFor(Simulation simulation, bool? isRunning = null)
       {
          var activeSimulation = _activeSubjectRetriever.Active<Simulation>();
-         bool simIsActive = (activeSimulation != null) && (activeSimulation == simulation);
-         updateSimulationStateFrom(simulation, isActiveSimulation: simIsActive);
+         var isActiveSimulation = Equals(activeSimulation, simulation);
```
msevestre, line 343: resharper suggestions is accurate. We should not dismiss them

msevestre (commented):
This fixes the bug but I feel that we have a bigger issue. The simulationState was meant to be used when we had ONE simulation running. Now we can have multiple so quick switching between tabs and running stopping simulation could potentiall;y lead to weird behavior. Maybe mostly theoretical but we should look at this eventually

Yuri05 (approved):

Yuri05:
Approved... but meanwhile I'm lost in the logic. Just hope it works...

msevestre:
Yeah the implementation does not make sense for multiple runs. I don’t this it really can lead to non sense anyways. It’s all a matter of timing and the way we check that a simulation is running is off

PR #3275, opened 2025-09-30
3249 centralize on the simulation runner all the inofrmation

### LINKED ISSUE #3249: Centralize on the simulation runner all the inofrmation.
As now the simulation runner is being consumed for "IsRunning" but as the event indicating the simultion has finished running is being fired on a different place, then we can`t fully rely on it.

The simulation runner should be the one in charge to fire this evet and also to store the values from the % of completion for each simulation running.

This way we will be setting the ground for having a reliable indication of multiple simulations running.

benjaperez1983:
Fixes #3249

# Description

Centralize all the infromation for running simulations on just one class.

## Type of change

- [x] Algorithm update - updates algorithm documentation/questions/answers etc.

# How Has This Been Tested?

- [x] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/PKSim.Core/Services/InteractiveSimulationRunner.cs
+++ b/src/PKSim.Core/Services/InteractiveSimulationRunner.cs
@@ -1,11 +1,15 @@
-﻿using System.Collections.Concurrent;
+﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Drawing;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Events;
 using OSPSuite.Core.Serialization.SimModel.Services;
+using OSPSuite.Utility;
+using OSPSuite.Utility.Events;
 using OSPSuite.Utility.Extensions;
 using PKSim.Core.Events;
 using PKSim.Core.Model;
@@ -30,8 +34,9 @@ namespace PKSim.Core.Services
       private readonly ILazyLoadTask _lazyLoadTask;
       private readonly IExecutionContext _executionContext;
       private readonly ConcurrentDictionary<Simulation, CancellationTokenSource> _cancellationTokenSources = new ConcurrentDictionary<Simulation, CancellationTokenSource>();
-
+      private readonly IEventPublisher _eventPublisher;
       private readonly SimulationRunOptions _simulationRunOptions;
+      protected bool _shouldRaiseEvents;
 
       public InteractiveSimulationRunner(
          ISimulationSettingsRetriever simulationSettingsRetriever,
@@ -39,7 +44,8 @@ namespace PKSim.Core.Services
          ICloner cloner,
          ISimulationAnalysisCreator simulationAnalysisCreator,
          ILazyLoadTask lazyLoadTask,
-         IExecutionContext executionContext)
+         IExecutionContext executionContext,
+         IEventPublisher eventPublisher)
       {
          _simulationSettingsRetriever = simulationSettingsRetriever;
          _simulationRunner = simulationRunner;
@@ -47,11 +53,12 @@ namespace PKSim.Core.Services
          _simulationAnalysisCreator = simulationAnalysisCreator;
          _lazyLoadTask = lazyLoadTask;
          _executionContext = executionContext;
+         _eventPublisher = eventPublisher;
 
          _simulationRunOptions = new SimulationRunOptions
          {
             CheckForNegativeValues = true,
-            RaiseEvents = true,
+            RaiseEvents = true, //This is always true, but still worth checking before rising an event since it can change.
             RunForAllOutputs = false,
             SimModelExportMode = SimModelExportMode.Optimized
          };
@@ -65,7 +72,9 @@ namespace PKSim.Core.Services
 
       private async Task runSimulationAsync(Simulation simulation, bool selectOutput)
       {
+         _shouldRaiseEvents = _simulationRunOptions.RaiseEvents;
          var cts = new CancellationTokenSource();
+         var begin = new DateTime();
          if (!_cancellationTokenSources.TryAdd(simulation, cts)) //this will prevent from running one that is already running
             return;
          try
@@ -83,7 +92,8 @@ namespace PKSim.Core.Services
 
                _executionContext.PublishEvent(new SimulationOutputSelectionsChangedEvent(simulation));
             }
-
+            begin = SystemTime.UtcNow();
+            _eventPublisher.PublishEvent(new SimulationRunStartedEvent(simulation));
             await _simulationRunner.RunSimulation(simulation, _simulationRunOptions, cts.Token);
 
             addAnalysableToSimulationIfRequired(simulation);
@@ -92,10 +102,13 @@ namespace PKSim.Core.Services
          {
             if (_cancellationTokenSources.ContainsKey(simulation))
             {
+               var end = SystemTime.UtcNow();
+               var timeSpent = end - begin;
                if (_cancellationTokenSources.TryRemove(simulation, out var ctsToDispose))
                {
                   ctsToDispose.Dispose();
                }
+               _eventPublisher.PublishEvent(new SimulationRunFinishedEvent(simulation, timeSpent));
             }
          }
       }
@@ -122,6 +135,11 @@ namespace PKSim.Core.Services
          if (simulation == null || !simulation.HasResults) return;
          if (simulation.Analyses.Count() != 0) return;
          _simulationAnalysisCreator.CreateAnalysisFor(simulation);
+      }
+      private void raiseEvent<T>(T eventToPublish)
+      {
+         if (_shouldRaiseEvents)
+            _eventPublisher.PublishEvent(eventToPublish);
       }
 
       public void StopSimulation(Simulation simulation)
--- a/src/PKSim.Core/Services/InteractiveSimulationRunner.cs
+++ b/src/PKSim.Core/Services/InteractiveSimulationRunner.cs
@@ -79,11 +79,10 @@ namespace PKSim.Core.Services
       {
          _shouldRaiseEvents = _simulationRunOptions.RaiseEvents;
          var cts = new CancellationTokenSource();
-         var begin = new DateTime();
-
-         if (!_cancellationTokenSources.TryAdd(simulation, cts))
+         if (!_cancellationTokenSources.TryAdd(simulation, cts)) //this will prevent from running one that is already running
             return;
 
+         var begin = new DateTime();
          try
          {
             _lazyLoadTask.Load(simulation);
@@ -97,10 +96,10 @@ namespace PKSim.Core.Services
                simulation.OutputSelections.UpdatePropertiesFrom(outputSelections, _cloner);
                mappingsNotSelected(simulation).Each(simulation.OutputMappings.Remove);
 
-               _executionContext.PublishEvent(new SimulationOutputSelectionsChangedEvent(simulation));
+               raiseEvent(new SimulationOutputSelectionsChangedEvent(simulation));
             }
             begin = SystemTime.UtcNow();
-            _executionContext.PublishEvent(new SimulationRunStartedEvent(simulation));
+            raiseEvent(new SimulationRunStartedEvent(simulation));
             await _simulationRunner.RunSimulation(simulation, _simulationRunOptions, cts.Token);
 
             addAnalysableToSimulationIfRequired(simulation);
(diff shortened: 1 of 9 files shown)


src/PKSim.Core/Services/InteractiveSimulationRunner.cs
```diff
@@ -65,9 +77,13 @@ public Task RunSimulation(Simulation simulation, bool selectOutput) =>
 
       private async Task runSimulationAsync(Simulation simulation, bool selectOutput)
       {
+         _shouldRaiseEvents = _simulationRunOptions.RaiseEvents;
          var cts = new CancellationTokenSource();
-         if (!_cancellationTokenSources.TryAdd(simulation, cts)) //this will prevent from running one that is already running
+         var begin = new DateTime();
+
+         if (!_cancellationTokenSources.TryAdd(simulation, cts))
```
msevestre: somehow you removed the comment ? 
Also probalby the beging should be AFTER this line otherwise it's set but never used?

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
```diff
@@ -83,7 +99,8 @@ private async Task runSimulationAsync(Simulation simulation, bool selectOutput)
 
                _executionContext.PublishEvent(new SimulationOutputSelectionsChangedEvent(simulation));
             }
-
+            begin = SystemTime.UtcNow();
```
msevestre, line 99: why are we setting the time again ?

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
```diff
@@ -92,10 +109,11 @@ private async Task runSimulationAsync(Simulation simulation, bool selectOutput)
          {
             if (_cancellationTokenSources.ContainsKey(simulation))
             {
+               var end = SystemTime.UtcNow();
+               var timeSpent = end - begin;
                if (_cancellationTokenSources.TryRemove(simulation, out var ctsToDispose))
-               {
                   ctsToDispose.Dispose();
-               }
+               _executionContext.PublishEvent(new SimulationRunFinishedEvent(simulation, timeSpent));
```
msevestre: I am not sure about this. Let say it was cancelled. Is this possible that you do not throw the even then? (Becaise you only do it in the if

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
```diff
@@ -104,7 +122,7 @@ private static IReadOnlyList<OutputMapping> mappingsNotSelected(Simulation simul
          simulation.OutputMappings.Where(outputMapping => !outputMappingIsSelected(simulation, outputMapping)).ToList();
 
       private static bool outputMappingIsSelected(Simulation simulation, OutputMapping outputMapping) =>
-         simulation.OutputSelections.AllOutputs.Any(quantitySelection => Equals(quantitySelection.Path, outputMapping.OutputPath));
+         simulation.OutputSelections.AllOutputs.Any(q => Equals(q.Path, outputMapping.OutputPath));
```
msevestre, line 122: if renaming, x maybe as a varialbe? q is not our default one letter. 
and with q I am assuming it's a quantity which it's not. (p, parameter etc.). x=>I have no idea

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
msevestre: also should it use raiseEvent?

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
```diff
@@ -30,8 +37,13 @@ public class InteractiveSimulationRunner : IInteractiveSimulationRunner
       private readonly ILazyLoadTask _lazyLoadTask;
       private readonly IExecutionContext _executionContext;
       private readonly ConcurrentDictionary<Simulation, CancellationTokenSource> _cancellationTokenSources = new ConcurrentDictionary<Simulation, CancellationTokenSource>();
-
+      private readonly IEventPublisher _eventPublisher;
       private readonly SimulationRunOptions _simulationRunOptions;
+      protected bool _shouldRaiseEvents;
```
msevestre: in factt why do we have a variable. Should it be _simulationRunOptions.RaiseEvents?

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
```diff
@@ -123,6 +141,11 @@ private void addAnalysableToSimulationIfRequired(Simulation simulation)
          if (simulation.Analyses.Count() != 0) return;
          _simulationAnalysisCreator.CreateAnalysisFor(simulation);
       }
+      private void raiseEvent<T>(T eventToPublish)
+      {
+         if (_shouldRaiseEvents)
```
msevestre: somewhere I saw that this is always true? why

msevestre (commented):
is there any place where the simulation enginess are used outside the InteractiveSimulationRuner (for instance a SimulationRunner?) in this case, it would break the beahvior of the events were used.

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
benjaperez1983: Yes, the code was already in place like this, so this is always set to true on the constructor, and never changed.
I believe this is a pattern related to other places where we do the same, so it's worth keeping it as this if you ask me.

benjaperez1983:
> InteractiveSimulationRuner

That`s a great question that I tried to cover.

I`ve checked all the events that are being published here:
SimulationOutputSelectionsChangedEvent
SimulationRunStartedEvent
SimulationRunFinishedEvent
SimulationRunCanceledEvent

Those 4 events are only being fired here, and not anywhere else.

Would you think that is sufficient to cover your question?

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
msevestre: Why is this comment solved without any explanation or code change?

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
msevestre: same here. 
@benjaperez1983 let's make sure you address my comments before resolving them because otherwise I can't trust that my comments are looked at and I have to open them all again one by one (which I have to do for this PR now)

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
```diff
@@ -237,9 +241,9 @@ private void updateJournalInfo(string name, string path, bool enabled)
 
       public void Handle(ProgressDoneWithMessageEvent eventToHandle)
       {
-         var message = $"{activeSimulationsCount} {eventToHandle.Message}";
+         var message = $"{_interactiveSimulationRunner.ActiveSimulationsCount} {eventToHandle.Message}";
```
msevestre: that;s not correct. I will still say 1 if there is only one and we had said we did not want to show it like this right? Did it make it like so in 12.1?

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
```diff
@@ -142,26 +149,26 @@ public void Handle(ProgressInitEvent eventToHandle)
...
 
       public void Handle(ProgressingEvent eventToHandle)
       {
+         var message = _interactiveSimulationRunner.ActiveSimulationsCount == 0 ? string.Empty : _interactiveSimulationRunner.ActiveSimulationsCount.ToString();
          update(StatusBarElements.ProgressBar)
             .WithValue(eventToHandle.ProgressPercent);
 
          update(StatusBarElements.ProgressStatus)
-            .WithCaption($"{activeSimulationsCount} {eventToHandle.Message}");
+            .WithCaption($"{message} {eventToHandle.Message}");
```
msevestre, line 169: same issue here. It will show 1 and a message if only one simulation. Or do I see this wrong?

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
benjaperez1983: Sorry I made a comment but must have not submitted it.

The thing here is we have 2 lines:
         var begin = new DateTime(); -> Which declares the variable.
begin = SystemTime.UtcNow(); -> Which sets the value right before starting the simulation.
So as the value is needed outside the scope of the try block, I declare the value first, and set it within the try. 
Let me know what you think about it.

msevestre (changes requested):
I think this is ok. I had some questions that were simpy ignored and I think we should try not avoid this if possible. 
1-Comment on the question and leave open for me to close. And it's ok to say that the suggestion does not make sense if there is a suggestion made

2-change code to accomodate suggestion

~3- closing without commenting~ => nope

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
benjaperez1983: Sorry about that, you’re absolutely right.
Not that I haven`t addressed it, the reason I didn’t change it as you suggested is that I’m trying to follow the same approach used by both the individual and population simulation runner engines.

Both inherit from:
public abstract class SimulationEngine
{
   protected bool _shouldRaiseEvents;
   ...
   protected void RaiseEvent<T>(T eventToPublish)
   {
      if (_shouldRaiseEvents)
         _eventPublisher.PublishEvent(eventToPublish);
   }
}

I understand that the _simulationRunOptions is not a class level variable there, and while we could use _simulationRunOptions.RaiseEvents directly, I chose to keep the _shouldRaiseEvents variable for consistency and symmetry with the existing SimulationEngine structure.

If you still think this is not a good Idea, I can change it. But if it is up to me, I believe this should stay this way.

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
benjaperez1983: I believe we did not want to show "0" if no simulations are running. If we have 1 running, then the message should be "1 Simulation running" WITH progressbar, and if we have >1 then no progress bar,, just the message.
You can see this implementation in setProgressBarVisibility method.
But let me know if I`m mistaken about this.

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
msevestre: Hard to read the code like this. Let's use ```code``` to make it easier

InteractiveSimulationRunner does not inherit from SimulationEngine

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
msevestre: oh I see. Typically I would assume you would Declare outside and of the try `DateTime begin`;  but that works too

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
msevestre: @benjaperez1983 still open

src/PKSim.Core/Services/InteractiveSimulationRunner.cs
msevestre: agreed. I still don't undertand why we have this varialbe however. the method is priovate, using a protected field etc.. feels wrong to me but I may be wrong too

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
msevestre: the 1 does not make sense. It should have been removed since it's only showed when there is 1 anyways so it's redundant

src/PKSim.Presentation/Presenters/Main/StatusBarPresenter.cs
msevestre: message shoukd be moved line 251 since it's not sued otherwise.makes it clear that it's not an oversight

msevestre (approved):
Looks good to me
In a few files howver, I feel that there are unuused namesspaces 
for instnace systems drawing in [InteractiveSimulationRunner.cs‎](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3275/files#diff-f92fc32edc04b009379280c2837797fc1d3c34a5aa9ed532b07744af5fa437d4) etc

Can you review the file that you changes by running resharper ? 
Approved

Yuri05:
@benjaperez1983 Please check the Copilot suggestions. Apart from that - looks good for me!

PR #3299, opened 2025-10-17
Fix: Stack overflow in SQLite on macOS when querying VIEW_INDIVIDUAL_…

Felixmil:
…PARAMETER_NOT_FOR_ALL_SPECIES

- Adds macOS-specific workaround in FlatIndividualParametersNotCommonForAllSpeciesRepository
- Queries base tables directly instead of using the complex VIEW with UNION
- Avoids stack overflow in sqlite3_prepare_interop() on ARM64 macOS
- Maintains full functionality by filtering and computing in C# instead of SQL

The VIEW_INDIVIDUAL_PARAMETER_NOT_FOR_ALL_SPECIES references ContainerParameters_Species which has complex UNION operations that cause deep recursion in SQLite's query compiler. This only affects macOS (both ARM64 and x86_64) with System.Data.SQLite.

Windows and Linux continue to use the original view-based query.

### DIFF
--- a/.github/workflows/build-and-test-pr_12.2.yml
+++ b/.github/workflows/build-and-test-pr_12.2.yml
@@ -33,6 +33,19 @@ jobs:
       - name: Build
         run: dotnet build PKSim.sln /p:Version=12.2.9999
 
+      - name: Create R dependency artifact
+        run: |
+          pushd .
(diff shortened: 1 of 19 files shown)


rwmcintosh:
@Felixmil closing in preference to another PR.

Felixmil:
[magic happening here](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1633).

PR #3301, opened 2025-10-20
remove access database

rwmcintosh:


Yuri05 (approved):

Yuri05:
@msevestre Do you remember why we kept it after switching to sqlite?

msevestre:
As an archive mostly. It should not be in the setup anyways so it's not used

PR #3336, opened 2025-11-19
Fixes #2900 "Effective molecular weight" can become negative.

### LINKED ISSUE #2900: "Effective molecular weight" can become negative.
[Generic_PBPK_bug.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/15347972/Generic_PBPK_bug.zip)

After I have changed some parameters in the "Advanced Parameters -> Particle Dissolution" section, creation of a simulation fails:

image

Even after I have reversed the parameter values to defaults.

rwmcintosh:
Fixes #2900

# Description
Added validation and dialog errors where necessary when updating mol weight

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

Validation in-grid for value edits
image

Validation in-grid for halogens
image

# Questions (if appropriate):


src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -207,6 +207,23 @@ public static void ScaleDistributionBasedOn(this IDistributedParameter parameter
...
+         var oldValue = parameter.Value;
+         try
+         {
+            parameter.Value = newValueInBaseUnit;
+            return effectiveMolWeightParameter.Value > 0;
+         }
+         finally
+         {
+            parameter.Value = oldValue;
+         }
```
rwmcintosh: Checking this requires evaluation of the effective mol weight formula. In this code I'm editing it in-place then reverting the old value in the finally.

src/PKSim.Core/Model/ParameterExtensions.cs
```diff
@@ -207,6 +207,23 @@ public static void ScaleDistributionBasedOn(this IDistributedParameter parameter
...
+      /// <summary>
+      /// Returns false if the proposed <paramref name="newValueInBaseUnit"/> will make the value of <paramref name="effectiveMolWeightParameter"/> negative
+      /// </summary>
+      public static bool IsEffectiveMolWeightPositive(this IParameter parameter, double newValueInBaseUnit, IParameter effectiveMolWeightParameter)
+      {
+         var oldValue = parameter.Value;
+         try
+         {
+            parameter.Value = newValueInBaseUnit;
+            return effectiveMolWeightParameter.Value > 0;
```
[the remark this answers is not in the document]
msevestre: why is this an extension method on PArameter? Isn't this so spcecific that is should be localized where it's used?

src/PKSim.Presentation/DTO/Compounds/MolWeightParameterDTO.cs
```diff
@@ -0,0 +1,33 @@
...
+      {
+         _effectiveMolWeightParameter = effectiveMolWeightParameter;
+         Rules.AddRange(AllRules.All());
+      }
+
+      private static class AllRules
+      {
+         private static IBusinessRule effectiveMolWeightPositive { get; } = CreateRule.For<MolWeightParameterDTO>()
+            .Property(item => item.Value)
+            .WithRule((dto, folder) => dto.Parameter.IsEffectiveMolWeightPositive(dto.Parameter.ConvertToBaseUnit(folder), dto._effectiveMolWeightParameter))
```
msevestre: e.g. here. As a local method from the MolWeightParmaeterDTO?

src/PKSim.Presentation/DTO/Mappers/ParameterToMolWeightParameterDTOMapper.cs
```diff
@@ -0,0 +1,29 @@
...
+      MolWeightParameterDTO MapFrom(IParameter molWeightParameter, IParameter effectiveMolWeightParameter);
+   }
+   
+   public class ParameterToMolWeightParameterDTOMapper : ParameterToParameterDTOMapper, IParameterToMolWeightParameterDTOMapper
+   {
+      public ParameterToMolWeightParameterDTOMapper(IRepresentationInfoRepository representationInfoRepository, IFormulaToFormulaTypeMapper formulaTypeMapper, IPathToPathElementsMapper parameterDisplayPathMapper, IFavoriteRepository favoriteRepository, IEntityPathResolver entityPathResolver, IParameterListOfValuesRetriever parameterListOfValuesRetriever) : base(representationInfoRepository, formulaTypeMapper, parameterDisplayPathMapper, favoriteRepository, entityPathResolver, parameterListOfValuesRetriever)
+      {
+      }
+
+      public MolWeightParameterDTO MapFrom(IParameter molWeightParameter, IParameter effectiveMolWeightParameter)
```
msevestre: what about it's a method of ParameterToParameterDTO instead? MapMolWeightFrom ?
Just a question as opposed to a suggestion

src/PKSim.Presentation/Presenters/Compounds/HalogensPresenter.cs
```diff
@@ -0,0 +1,59 @@
...
+         base(view, scaleParametersPresenter, editParameterPresenterTask, parameterTask, parameterDTOMapper, contextMenuFactory)
+      {
+         _dialogCreator = dialogCreator;
+      }
+
+      public override void SetParameterValue(IParameterDTO parameterDTO, double valueInDisplayUnits)
+      {
+         var parameter = parameterDTO.Parameter;
+
+         if (!parameter.IsEffectiveMolWeightPositive(parameter.ConvertToBaseUnit(valueInDisplayUnits), _effectiveMolWeight))
```
msevestre: hum/ Wou;dn't the validation using DTO be better in this case with the red cross?

rwmcintosh: Not so easy in the case of these combobox editors. I haven't seen this in another view.

rwmcintosh: Maybe I'm wrong, does it work in this case?

msevestre: oh it's a combo box. I did not see it. I think it works if you are bidning to a gridview where the cell is a combo box

msevestre:
image

oh but that's interesting. why do we have an error at the molweight? I guess why not .but then we do not need the message box?

rwmcintosh:
The effective mol weight can become negative either through setting the mol weight, or by editing the halogens. Both need a validation.

src/PKSim.Presentation/Presenters/Compounds/HalogensPresenter.cs
rwmcintosh: Yeah, if I do it through DTO validation then the Rule does not execute until after the editor closes. So you get this effect in the UI after the edtior closes and you open it again, but the parameter has already been updated.

image

rwmcintosh: Yeah, it's better if the validation prevents the invalid state. Is there a way to change that?

rwmcintosh: I think I know why it's not working quite right. I'll change and describe in PR.

msevestre:
if you have a rule, it should prevent you from RUNNING it. The rule should be defined in the actual parameter however .Can we have sthg like this> I forgot

src/PKSim.UI/Views/Parameters/ParameterSetView.cs
```diff
@@ -89,7 +89,6 @@ public void Initialize(UxGridView gridView)
          InitializeWithGrid(_gridView);
          _gridViewBinder = new GridViewBinder<ParameterDTO>(_gridView)
          {
-            ValidationMode = ValidationMode.LeavingRow,
```
rwmcintosh, line 92: That's why it was not working satisfactorily before. Without this, the validation shows up right when you try to add a halogen that makes effective weight below 0

src/PKSim.Core/Model/ParameterExtensions.cs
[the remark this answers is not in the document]
rwmcintosh: extension method removed

src/PKSim.Presentation/DTO/Compounds/MolWeightParameterDTO.cs
rwmcintosh: yeah, with the validation working on the dto, that code is now part of the dto rules.

src/PKSim.Presentation/DTO/Mappers/ParameterToMolWeightParameterDTOMapper.cs
rwmcintosh: yep

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -401,6 +401,7 @@ public static class Error
          public const string UnknownObserverBuilderType = "Observer builer type unknown.";
          public const string ImporterConfigurationNotFoundInProject = "Importer configuration for this dataset was not found in this project.\n\nThe data cannot be reloaded.";
          public const string SimulationCannotShareNamesWithCompounds = "Simulation cannot share names with compounds";
+         public const string EffectiveMolWeightCannotBeNegative = "Effective mol weight cannot be negative";
```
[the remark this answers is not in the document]
rwmcintosh: allowing 0

src/PKSim.Presentation/DTO/Compounds/MolWeightParameterDTO.cs
```diff
@@ -0,0 +1,45 @@
...
+      {
+         private static IBusinessRule effectiveMolWeightPositive { get; } = CreateRule.For<MolWeightParameterDTO>()
+            .Property(item => item.Value)
+            .WithRule((dto, valueInDisplayUnit) =>
+            {
+               var oldValue = dto.Parameter.Value;
+               try
+               {
+                  dto.Parameter.Value = dto.Parameter.ConvertToBaseUnit(valueInDisplayUnit);
+                  return dto._effectiveMolWeightParameter.Value >= 0;
```
Yuri05: For the mol weight parameter, we define the value as **>= 15 g/mol**:



I think for the effective mol weight, the lower bound should be the same (@PavelBal Agree?). 

E.g. you could pass the lower bound of the "Molecular Weight" parameter as an additional argument to the dto:
`
public MolWeightParameterDTO(IParameter molWeightSubParameter, IParameter effectiveMolWeightParameter, double molWeightLowerBound) : base(molWeightSubParameter)` and then check for

`return dto._effectiveMolWeightParameter.Value >= molWeightLowerBound;`

src/PKSim.Presentation/DTO/Compounds/MolWeightParameterDTO.cs
```diff
@@ -0,0 +1,45 @@
...
+using PKSim.Assets;
+using PKSim.Presentation.DTO.Parameters;
+
+namespace PKSim.Presentation.DTO.Compounds
+{
+   public class MolWeightParameterDTO : ParameterDTO
+   {
+      private readonly IParameter _effectiveMolWeightParameter;
+
+      public MolWeightParameterDTO(IParameter molWeightParameter, IParameter effectiveMolWeightParameter) : base(molWeightParameter)
```
Yuri05: Here the `molWeightParameter` argument can be one of the Parameters "Molecular weight", "Br, "Cl", ... correct? 
I would then rename it to something like `molWeightSubParameter` or (find a better name :)).
Because when i see `molWeightParameter` I think it's "Molecular weight" parameter (but not one of the halogens).

Yuri05 (changes requested):
s. comments

src/PKSim.Presentation/DTO/Compounds/MolWeightParameterDTO.cs
PavelBal: Agree.

PavelBal: Well, wait... In theory, if someone would like to model the distribution of hydrogen :D the MW would be 1. So I do not see a reason to limit it to 15 (oxygen?).

Yuri05: I will not find the original discussion for sure, but with very low MW other model parameters became negative and simulations were not runnable anyways.

Yuri05: Btw: if now you will modify your mol weight and halogens parameters so that Effective mol weight becomes < 15 g/Mol: the simulation will be not runnable



So in fact the min value of the Effective molecular weight parameter is already set to 15 g/Mol.
It's just not evaluated directly when modifying mol weight/halogens parameters (probably because the efective mol weight parameter is not editable).

rwmcintosh: Yes, we can use the built-in min for this validation

src/PKSim.Presentation/DTO/Compounds/MolWeightParameterDTO.cs
rwmcintosh: done

Yuri05 (approved):

PR #3339, opened 2025-11-20
update core

rwmcintosh:
This branch doesn't build after updating core version due to changes with `CheckForNegativeValues`


Yuri05:
> This branch doesn't build after updating core version due to changes with `CheckForNegativeValues`

Remove the line 
https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/ffbff57e91e2faf2bd11b287bf0b94465adb64f6/src/PKSim.Core/Services/InteractiveSimulationRunner.cs#L63

PR #3362, opened 2025-12-22
#3361 improve and change info message

### LINKED ISSUE #3361: Improve (and change?) the info message when some individuals in popsim failed
Add **individuals** to the message: _N out of M **individuals** were ..._

I also wonder if we should show the number of **failed** individuals instead.
`N out of M individuals failed for the simulation "Tralala" (X.YZ% success rate)`

And: should we also display the IDs of failed individuals? This would allow extracting them as Individuals-BB for the further investigations.
(At the moment I can figure out those IDs by e.g. exporting the results to CSV and checking for the missing IDs, but showing them directly would be more user-friendly).
```
N out of M individuals failed for the simulation "Tralala" (X.YZ% success rate)
Ids of failed individuals: 1,2,10,11,20
```

Image

benjaperez1983:
Fixes #3361 

# Description

Improve the message for population results when individuals are failing

## Type of change

# How Has This Been Tested?

- [x] Integration tests
- [ ] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

# Questions (if appropriate):

### DIFF
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2602,7 +2602,7 @@ namespace PKSim.Assets
 
                var idsSuffix = failedCount > MaxFailedIndexesToShow ? $" of {failedCount}" : string.Empty;
 
-               message += $"{Environment.NewLine}Failing individual numbers (first {indexesToShow.Count}{idsSuffix}) {string.Join(", ", indexesToShow)}";
+               message += $"{Environment.NewLine}Failing individual ids (first {indexesToShow.Count}{idsSuffix}) {string.Join(", ", indexesToShow)}";
             }
 
             return message;
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2582,7 +2582,7 @@ namespace PKSim.Assets
          public static string LinkedExpressionProfileIs(string expressionProfileName) => $"Using expression profile <b>{expressionProfileName}</b>";
 
          public static string ChildPughScoreFor(string score) => $"Child-Pugh {score}";
-
+         private const int MaxFailedIdsToShow = 20;
          public static string PopulationSimulationFailed(IReadOnlyCollection<int> failedIds, int totalCount, string simulationName)
          {
             if (totalCount == 0)
@@ -2597,12 +2597,12 @@ namespace PKSim.Assets
             {
                var idsToShow = failedIds
                   .OrderBy(id => id)
-                  .Take(20)
+                  .Take(MaxFailedIdsToShow)
                   .ToList();
 
                message +=
-                  $"{Environment.NewLine}Failing individual IDs (first {idsToShow.Count}" +
-                  $"{(failedCount > 20 ? " of " + failedCount : string.Empty)}): " +
+                  $"{Environment.NewLine}Failing individual Indexes (first {idsToShow.Count}" +
+                  $"{(failedCount > MaxFailedIdsToShow ? " of " + failedCount : string.Empty)}): " +
                   string.Join(", ", idsToShow);
             }
 
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -2600,7 +2600,7 @@ namespace PKSim.Assets
                   .Take(MaxFailedIndexesToShow)
                   .ToList();
 
-               var idsSuffix = failedCount > 20 ? $" of {failedCount}" : string.Empty;
+               var idsSuffix = failedCount > MaxFailedIndexesToShow ? $" of {failedCount}" : string.Empty;
 
                message += $"{Environment.NewLine}Failing individual numbers (first {indexesToShow.Count}{idsSuffix}) {string.Join(", ", indexesToShow)}";
             }
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -4,7 +4,6 @@ using System.Linq;
 using System.Text;
 using OSPSuite.Assets;
 using OSPSuite.Assets.Extensions;
-using OSPSuite.Core.Domain;
 using OSPSuite.Utility.Extensions;
 
 namespace PKSim.Assets
@@ -925,7 +924,7 @@ namespace PKSim.Assets
          public static string SnapshotDuplicateEntryByName(string name, string type) =>
             $"Another {type} named '{name}' already exists in the project. Snapshot file is corrupted.";
 
-         public static string EffectiveMolWeightMustBeGreaterThan(double valueInDisplayUnit, string displayUnit) => 
+         public static string EffectiveMolWeightMustBeGreaterThan(double valueInDisplayUnit, string displayUnit) =>
             $"Effective mol weight must be greater than or equal to {valueInDisplayUnit} {displayUnit}";
       }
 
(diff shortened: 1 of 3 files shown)


src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -2584,16 +2583,31 @@ public static string UserTemplateDatabaseDatabaseUsedOldFormatAndCannotBeLoaded(
...
+            if (failedCount > 0)
+            {
+               var idsToShow = failedIds
+                  .OrderBy(id => id)
+                  .Take(20)
+                  .ToList();
+
+               message +=
+                  $"{Environment.NewLine}Failing individual IDs (first {idsToShow.Count}" +
+                  $"{(failedCount > 20 ? " of " + failedCount : string.Empty)}): " +
```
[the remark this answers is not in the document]
msevestre: I agree with this comment. Let's create a const for this

src/PKSim.Core/Services/PopulationSimulationEngine.cs
```diff
@@ -103,9 +104,22 @@ private void evaluateResultsAndShowMessage(PopulationSimulation populationSimula
          if (!failingSimulations.Any())
             return;
 
+         var failingInfoSet = new HashSet<IndividualRunInfo>(failingSimulations);
+         var failingIds = new List<int>();
+         var totalIndividuals = populationRunResults.IndividualRunInfos.Count();
+
+         for (int id = 0; id < totalIndividuals; id++)
```
msevestre: This is confusing to me here
`id` is not an id is jsut a count. 
also it looks liek it coule be written in a linq expression (where. select?). Feels old school to me

tests/PKSim.Tests/Core/PopulationSimulationEngineSpecs.cs
```diff
@@ -153,7 +150,50 @@ protected override Task Because()
...
+   {
+      private string _message;
+
+      protected override async Task Context()
+      {
+         await base.Context();
+
+         var individualResult1 = new IndividualResults { Id = 0 ,IndividualId = 0};
+         var individualResult2 = new IndividualResults { Id = 1 ,IndividualId = 1};
+         var individualResult3 = new IndividualResults { Id = 2 ,IndividualId = 2};
```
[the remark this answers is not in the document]
msevestre: yes let's run resharper

tests/PKSim.Tests/Core/PopulationSimulationEngineSpecs.cs
```diff
@@ -153,7 +150,50 @@ protected override Task Because()
...
+            .Returns(_runResults);
+      }
+
+      protected override Task Because()
+      {
+         return sut.RunAsync(_populationSimulation, _simulationRunOptions);
+      }
+
+      [Observation]
+      public void should_list_failing_simulation_ids_in_message()
```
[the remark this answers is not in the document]
msevestre: also correct

msevestre (changes requested):
copilot is right and one funcion could be rewritten using more modern syntax
At least rename the variable

src/PKSim.Core/Services/PopulationSimulationEngine.cs
rwmcintosh: Yes, the `ids` are _generally_ organized just using indexes, but using for loop iterator as an id we are leaking that implementation detail. I would do something like this instead so that if the implementation of the `id` changes then the code won't break.

```
      private void evaluateResultsAndShowMessage(PopulationSimulation populationSimulation, PopulationRunResults populationRunResults)
      {
         var failingIndividuals = populationRunResults.IndividualRunInfos.Where(x => !x.Success).Select(runInfo => populationRunResults.IndividualRunInfos.IndexOf(runInfo)).ToList();

         var message = PKSimConstants.UI.PopulationSimulationFailed(failingIndividuals, populationRunResults.IndividualRunInfos.Count(), populationSimulation.Name);

         _dialogCreator.MessageBoxInfo(message);
      }
```

I would also be careful about how the message is worded because this code still uses an index to create the warning message. Either make it clear that it's an index in the message, or you might have to create a method in `PopulationRunResults` that returns the `id` from the `IndividualRunInfo`. Eg `populationRunResults.GetIdFor(individualRunInfo)`

If you do either the message change, or the `id` getter, then changing the implementation of the individual `id` won't break this code.

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -2583,17 +2582,32 @@ public static string UserTemplateDatabaseDatabaseUsedOldFormatAndCannotBeLoaded(
...
+
+            if (failedCount > 0)
+            {
+               var idsToShow = failedIds
+                  .OrderBy(id => id)
+                  .Take(MaxFailedIdsToShow)
+                  .ToList();
+
+               message +=
+                  $"{Environment.NewLine}Failing individual Indexes (first {idsToShow.Count}" +
```
rwmcintosh: I would prefer `Failing individual numbers` instead of `indexes`. Either way it should not be capitalized.

rwmcintosh (changes requested):
Reconsider the wording of the message and address all the CoPilot items. They are very good suggestions.

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -2583,17 +2582,31 @@ public static string UserTemplateDatabaseDatabaseUsedOldFormatAndCannotBeLoaded(
          public static string LinkedExpressionProfileIs(string expressionProfileName) => $"Using expression profile <b>{expressionProfileName}</b>";
 
          public static string ChildPughScoreFor(string score) => $"Child-Pugh {score}";
-
-         public static string PopulationSimulationSuccessful(int successCount, int totalCount, string simulationName)
+         private const int MaxFailedIdsToShow = 20;
+         public static string PopulationSimulationFailed(IReadOnlyCollection<int> failedIds, int totalCount, string simulationName)
```
rwmcintosh: I would just go back over this PR and evaluate in each case where you use `Id` whether or not it's really the `Id`. I think in most cases we are using `index` and `id` interchangably and we should not.

Even though coincidentally they are the same value.

rwmcintosh (changes requested):

rwmcintosh (approved):

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -2581,19 +2580,33 @@ public static string UserTemplateDatabaseDatabaseUsedOldFormatAndCannotBeLoaded(
...
+            var message = $"{failedCount} out of {totalCount} individuals failed for simulation {simulationName} ({successRate:F1}% success rate).";
+
+            if (failedCount > 0)
+            {
+               var indexesToShow = failedIndexes
+                  .OrderBy(index => index)
+                  .Take(MaxFailedIndexesToShow)
+                  .ToList();
 
+               var idsSuffix = failedCount > 20 ? $" of {failedCount}" : string.Empty;
```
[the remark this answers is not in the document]
Yuri05: @benjaperez1983 makes sense

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -2581,19 +2580,33 @@ public static string UserTemplateDatabaseDatabaseUsedOldFormatAndCannotBeLoaded(
...
+
+               message += $"{Environment.NewLine}Failing individual numbers (first {indexesToShow.Count}{idsSuffix}) {string.Join(", ", indexesToShow)}";
```
Yuri05: I think we should use `Ids` instead of `numbers` in the message, because we use it in other cases in the UI, e.g. like below. 
In the rest of the code, I would keep it AS IS (index / indexes), so it's only about the shown message.

Image

Yuri05: Btw., I was not sure how **imported** populations are handled (because in a population CSV files user can provide **arbitrary** individual IDs). But apparently when importing from CSV, those individual IDs are automatically replaced by 0,1, ... N-1 - so there is no difference between ID and Index of an individual, again.

rwmcintosh: Shoot, I didn't realize we do this elsewhere in the UI (use index as Id) because it's coincidental that the Id is an index.

rwmcintosh: I think we should change this. Not in this PR and not in 12.2 but in the future we should not assume that an iterator index is the Id.

Yuri05: > I think we should change this. Not in this PR and not in 12.2 but in the future we should not assume that an iterator index is the Id.

Agree

Yuri05 (changes requested):
s. the comments 
https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3362/changes#r2661640910
and
https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3362/changes#r2664757679
Once done, can be merged as for me.

Yuri05 (approved):

PR #3401, opened 2026-02-13
Fixes #3400 create an individualbuildingblock from individualcharacteristics and expose for mobi dynamic loading

### LINKED ISSUE #3400: Create an `IndividualBuildingBlock` from `IndividualCharacteristics` and expose for MoBi dynamic loading

rwmcintosh:
Fixes #3400 create an individualbuildingblock from individualcharacteristics and expose for mobi dynamic loading

# Description
A PKSimStarter style access for IndividualBuildingBlock that does not use Presentation or UI and re-uses existing `IndividualCharacteristics`

## Type of change

# How Has This Been Tested?

- [x] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

## Summary by CodeRabbit

* **Chores**
  * Updated OSPSuite dependencies to version 13.0.85 across projects and tests.

* **Refactor**
  * Consolidated snapshot/data model handling to rely on shared external types for improved consistency.
  * Streamlined internal domain classes to reduce duplication.

* **New Features**
  * Added building-block creation capability to the R API for easier integration.

* **Tests**
  * Added an integration test verifying building-block creation.


src/PKSim.R/Exchange/BuildingBlockCreator.cs
```diff
@@ -0,0 +1,27 @@
...
+         var individualFactory = Api.ResolveTask<IIndividualFactory>();
+         var originDataMapper = Api.ResolveTask<OriginDataMapper>();
+         var serializer = Api.ResolveTask<IPKMLPersistor>();
+         var mapper = Api.ResolveTask<IIndividualToIndividualBuildingBlockMapper>();
+
+         var originData = originDataMapper.MapToModel(individualCharacteristics, new SnapshotContext(new PKSimProject(), SnapshotVersions.Current)).Result;
+
+         var buildingBlock = individualFactory.CreateAndOptimizeFor(originData, individualCharacteristics.Seed);
+
+         return serializer.Serialize(mapper.MapFrom(buildingBlock));
```
rwmcintosh: The meat and potatoes

src/PKSim.R/Exchange/BuildingBlockCreator.cs
```diff
@@ -0,0 +1,27 @@
...
+      }
```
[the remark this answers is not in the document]
rwmcintosh: We'll keep it with the exception

src/PKSim.R/Exchange/BuildingBlockCreator.cs
```diff
@@ -0,0 +1,27 @@
...
+using PKSim.Core.Model;
+using PKSim.Core.Snapshots.Mappers;
+
+namespace PKSim.R.Exchange
+{
+   public class BuildingBlockCreator
+   {
+      public static string CreateIndividual(IndividualCharacteristics individualCharacteristics)
+      {
+         var individualFactory = Api.ResolveTask<IIndividualFactory>();
```
msevestre: why can't we exposte a method in the Individual factory doing all of this? The OridinDataMApper, etc all of this feels like some internals of the Individual Factory to me

msevestre (commented):
I am wondering why the Individual Factory can't create an individual from the IndividualCharactristic

src/PKSim.R/Exchange/BuildingBlockCreator.cs
rwmcintosh: I did it this way so as not to expose a method to R that should not be used. Maybe if we use 'internal' scope it won't be accessible? What do you think?

msevestre: I don't get it, Naybe we are looking at two differnet IndividualFactory. I can't tell from the diff

but we have thsi

```
public CreateIndividualResults CreateIndividual(IndividualCharacteristics individualCharacteristics)
``` 
in the IndividualFactory (R) that does almost every steps that is below (I think) . Why can't we reuse this ? or create the method that returns the BuildingBlock ? 
Also I don't get why we have all the API.Resolve everywhere. Why aren't this in the constrcutor like any other services. 

OVerall this just looks weird to me lol

msevestre: but there is probably reason. We can hop on a call

rwmcintosh: yes, we can have a call.

The reason not to create another method in the R `IndividualFactory` is because that becomes exposed to R since `IndividualFactory` is a task consumed by R. But this new method should not be used directly by R, only through MoBi exchange with PK-Sim. 

That's because there is a separate dimension system in PK-Sim vs. MoBi/OSPSuite-R. The `IndividualBuildingBlock` needs to be created in the correct dimension system. We do this through the PK-Sim Starter already by serializing in PK-Sim, transferring PKML, then deserializing in MoBi.

The reason we don't have DI in the constructor is because we do the exchange with MoBi through static class and methods.

msevestre (approved):

PR #3447, opened 2026-03-16
Merge branch 'develop' into V13

rwmcintosh:
merge commit

## Summary by CodeRabbit

* **Chores**
  * Discontinued automatic nightly build schedule; builds now available on manual trigger only
  * Updated build infrastructure tooling

* **Tests**
  * Expanded API integration test coverage for core factories and runner functionality

### DIFF
--- a/.github/workflows/build-nightly_v13.yml
+++ b/.github/workflows/build-nightly_v13.yml
@@ -3,6 +3,8 @@ run-name: Version 13.0.${{ github.run_number }}
 
 on:
   workflow_dispatch:
+  schedule:
+    - cron: '0 2 * * *'
 
 env:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-nightly_v13.yml
rwmcintosh, line 1: @Yuri05 do we want these changes in V13 also? I removed the schedule for now

Yuri05: Don't think it's required. 
As long as the workflow **build-nightly_v13.yml‎** exists only in develop and in V13 (and starting from develop is prevented by the check there) everything should be fine.

PR #3481, opened 2026-04-02
Add OverwriteParameterSet domain model with selections and serialization

### LINKED ISSUE #3427: Domain model for OverwriteParameterSet
## Parent PRD

#3422

## What to build

Create the core domain model for OverwriteParameterSet in PKSim.Core. This is the foundation slice that all other slices depend on.

**OverwriteParameterSet class:**
- Properties: Name, IsDefault
- ExtendedProperties for metadata (Species, Disease State stored here, extensible for future needs)
- Contains `IReadOnlyList<ParameterValue>` (reusing existing ParameterValue class from OSPSuite.Core)
- ParameterValues are unique by path within a set
- Defined in PKSim.Core, not OSPSuite.Core

**Compound changes:**
- Add `IReadOnlyList<OverwriteParameterSet>` property to Compound
- Starts as empty list (no default OverwriteParameterSet created automatically)
- Internal collection management exposed through service/task layer

**SimulationConfiguration changes:**
- Add `OverwriteParameterSetSelections` object to store which OverwriteParameterSet is selected per compound
- Encapsulated with behavior methods, not exposed as primitive dictionary

**Path matching logic:**
- Dedicated, testable method to determine if a simulation parameter belongs to a compound used as a building block in the simulation configuration
- Note: compounds created in the simulation but not used as building blocks (like unnamed metabolites) should not be matched
- Logic: parameter path contains compound name as one of its elements
- Returns compound name if matched, null if no match
- Designed for extension if more sophisticated matching needed later

## Acceptance criteria

- [ ] `OverwriteParameterSet` class exists in PKSim.Core with Name, IsDefault, ExtendedProperties, and ParameterValues collection
- [ ] ParameterValues are unique by path within a set (enforced)
- [ ] Compound has `IReadOnlyList<OverwriteParameterSet>` property (starts empty)
- [ ] `OverwriteParameterSetSelections` class exists for SimulationConfiguration
- [ ] Path matching logic implemented and tested (including unnamed metabolite exclusion)
- [ ] Unit tests for all domain classes

## Blocked by

None - can start immediately

## User stories addressed

Foundation for all user stories in the PRD.

rwmcintosh:
## Summary
Domain for overwrite parameter sets

Closes #3427

### DIFF
--- a/src/PKSim.Core/Model/OverwriteParameterSet.cs
+++ b/src/PKSim.Core/Model/OverwriteParameterSet.cs
@@ -1,0 +1,43 @@
+using System.Collections.Generic;
+using System.Linq;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Core.Domain.Services;
+using OSPSuite.Utility.Collections;
+using OSPSuite.Utility.Extensions;
+
+namespace PKSim.Core.Model
+{
+   public class OverwriteParameterSet : ObjectBase
+   {
+      private readonly ICache<ObjectPath, ParameterValue> _parameterValues = new Cache<ObjectPath, ParameterValue>(pv => pv.Path, x => null);
+
+      public bool IsDefault { get; set; }
+
+      public ExtendedProperties ExtendedProperties { get; } = new();
+
+      public IReadOnlyList<ParameterValue> ParameterValues => _parameterValues.ToList();
+
+      public void Add(ParameterValue parameterValue) => 
+         _parameterValues[parameterValue.Path] = parameterValue;
+
+      public void Remove(ParameterValue parameterValue) =>
+         _parameterValues.Remove(parameterValue.Path);
+
+      public ParameterValue ParameterValueByPath(string parameterPath) =>
+         _parameterValues[parameterPath.ToObjectPath()];
+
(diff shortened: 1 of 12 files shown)


src/PKSim.Core/Model/OverwriteParameterSet.cs
```diff
@@ -0,0 +1,43 @@
...
+         base.UpdatePropertiesFrom(source, cloneManager);
+
+         if (!(source is OverwriteParameterSet sourceSet))
+            return;
+
+         IsDefault = sourceSet.IsDefault;
+         ExtendedProperties.UpdateFrom(sourceSet.ExtendedProperties);
+         _parameterValues.Clear();
+         sourceSet.ParameterValues.Each(pv => Add(cloneManager.Clone(pv)));
+      }
```
Yuri05, line 41: I think we will also need something like 
`ChangeCompoundName(string oldCompoundName, string newCompoundName)` which would replace the compound name in all parameter paths

Yuri05 (approved):

src/PKSim.Core/Model/OverwriteParameterSet.cs
rwmcintosh: And in the OverwriteSelection object.

msevestre:
What about you give me a chance to review the code before merging ?

PR #3483, opened 2026-04-02
Remove SQLite.Interop.dll.dylib from dependency package

### LINKED ISSUE #3372: Remove SQLite.Interop.dll.dylib from the dependency package
That file only comes in Intel format and we are no longer building any MacOS packages on Intel.

rwmcintosh:
## Summary
- Removes the macOS Intel (`osx-x64`) SQLite interop library from the R dependency package
- This file only comes in Intel format and macOS packages are no longer built on Intel

Fixes #3372

## Summary by CodeRabbit

* **Chores**
  * Removed macOS-specific SQLite interop artifact handling from the build process; Windows and Linux functionality unaffected.


msevestre:
what about all the R experiemnting stuff that was implemented a while back?

rwmcintosh:
> what about all the R experiemnting stuff that was implemented a while back?

We're removing this because there aren't any more GitHub runners that run on Mac with Intel hardware. So in 12.3 we have to remove Mac support.

Mac ARM is added in V13

Issue #3492, opened 2026-04-04
Description of the Event Tab in simulation


msevestre:
Now that we have an event protocol, the other tab will become obsolete (or to add single events)
<img width="656" height="751" alt="Image" src="https://github.com/user-attachments/assets/07276395-be33-48f2-b8d3-972b71f19e73" />

I think we should add a description there or maybe even only show it on configure and clone when it's already there. Otherwise it will just confuse more than help

we probably alos need some info here too

<img width="644" height="368" alt="Image" src="https://github.com/user-attachments/assets/58b3baf7-c04c-4452-aa7e-1b1c0d1b79b4" />

Yuri05:
>I think we should add a description there or maybe even only show it on configure and clone when it's already there. Otherwise it will just confuse more than help

Agree. Would show the Events-Tab only for clone/configure of old simulations with "single" events.
When showing, would remove (or disable) the "Add Event" buttons and add a tooltip (_for new events, go to adm. protocols_)
<img width="640" height="216" alt="Image" src="https://github.com/user-attachments/assets/0df3d0c8-c9cf-49f5-bec7-437a1a749917" />

Yuri05:
> > I think we should add a description there or maybe even only show it on configure and clone when it's already there. Otherwise it will just confuse more than help
> 
[...]

PR #3494, opened 2026-04-04
Create event groups for protocol event schema items (Fixes #3493)

### LINKED ISSUE #3493: Create Event in the simulation.
## Context

In #3461, event schema items were filtered out in `EventBuildingBlockCreator.addProtocol` to prevent a `NullReferenceException` — event schema items were being passed to `ApplicationFactory.CreateFor`, which only handles drug administration entries (it expects dose parameters that events don't have).

The filter is at `src/PKSim.Core/Services/EventBuildingBlockCreator.cs` line 153:
```csharp
_schemaItemsMapper.MapFrom(protocol).Where(item => !item.IsEvent).ToList().Each(...)
```

## What needs to be done

After filtering out event schema items from application creation, we need to **create event groups** for those filtered-out event entries. This should happen in the same `addProtocol` method, after application creation.

For each event schema item in the protocol:
1. Look up the `EventPlaceholderMapping` from `compoundProperties.ProtocolProperties.EventMappingWith(schemaItem.EventKey)`
2. Resolve the `PKSimEvent` building block from the mapping (via `UsedBuildingBlockByTemplateId`)
3. Clone the event group template from `IEventGroupRepository.FindByName(pkSimEvent.TemplateName)`
4. Set the start time from the schema item's `StartTime` parameter
5. Set event group parameters from the PKSimEvent building block
6. Add the event group to the protocol's event group builder

### Reference pattern

Follow the existing `createNonApplicationEvents` method (lines 82-133) which does the same thing for standalone events from `EventProperties.EventMappings`. The key difference is:
- **Standalone events**: start time comes from `EventMapping.StartTime`
- **Protocol events**: start time comes from the schema item's `StartTime` parameter

### Files to modify
- `src/PKSim.Core/Services/EventBuildingBlockCreator.cs` — add event group creation for protocol event entries
- `tests/PKSim.Tests/IntegrationTests/SimulationsWithEventsSpecs.cs` — extend the integration test to verify event groups are actually created in the model

msevestre:
## Summary
- In `EventBuildingBlockCreator.addProtocol`, event schema items (previously filtered out with a TODO) now generate proper event groups
- Groups event schema items by `EventKey` — same key gets one parent event group with sub-containers per start time, different keys get separate event groups even if they reference the same PKSimEvent template
- Follows the same pattern as `createNonApplicationEvents` for standalone events

## Test plan
- [x] Unit tests for `EventBuildingBlockCreator`: same key grouping, missing mapping skip, different keys with same template
- [x] Integration test verifying event group appears in the simulation model
- [x] All 10 new tests pass, build succeeds

Closes #3493

## New Features
* Protocol events are now properly created and configured from protocol specifications with customized naming and start-time parameters
* Multiple event instances per protocol are supported with automatic deduplication of duplicate start times
* Event parameter handling and structure mapping have been enhanced for improved accuracy in simulations

### DIFF
--- a/src/PKSim.Core/Services/EventBuildingBlockCreator.cs
+++ b/src/PKSim.Core/Services/EventBuildingBlockCreator.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using OSPSuite.Assets;
@@ -5,10 +6,12 @@ using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Domain.Descriptors;
 using OSPSuite.Core.Domain.Services;
+using OSPSuite.Utility.Collections;
 using OSPSuite.Utility.Extensions;
 using PKSim.Core.Mappers;
 using PKSim.Core.Model;
 using PKSim.Core.Repositories;
+using static OSPSuite.Core.Domain.Constants.Parameters;
 
 namespace PKSim.Core.Services
 {
@@ -66,7 +69,7 @@ namespace PKSim.Core.Services
 
             createApplications(_simulation.CompoundPropertiesList);
 
-            createNonApplicationEvents();
+            createEvents();
 
             _parameterDefaultStateUpdater.UpdateDefaultFor(_eventGroupBuildingBlock);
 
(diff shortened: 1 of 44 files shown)


src/PKSim.Core/Services/EventBuildingBlockCreator.cs
```diff
@@ -160,6 +161,66 @@ private void addProtocol(CompoundProperties compoundProperties)
...
+
+            // create subcontainers for each schema item sharing this key
+            int eventIndex = 0;
+
+            foreach (var schemaItem in group.OrderBy(item => item.StartTime.Value))
+            {
+               var mainSubContainer = _cloneManagerForBuildingBlock.Clone(templateEventGroup.MainSubContainer());
+
+               eventIndex += 1;
+               mainSubContainer.Name = $"{pkSimEvent.Name}_{eventIndex}";
```
msevestre: @Yuri05 I am not so sure about this. Also how does it work with single events added also. This is getting complicated ha ha

msevestre: @Yuri05 Does this make any sense?
image

Should we prevent this from happening (e.g., mapping TWO different placeholders to the same event)

msevestre: I guess we are allowing it for Formulation..hum

msevestre: ok my idea is to group all by event and insure we only have unique start time. I can't see how this makes sense to strt the same event at multiple time (or this is maybe even an errror?)

msevestre:
@coderabbitai full review

src/PKSim.Core/Services/EventBuildingBlockCreator.cs
```diff
@@ -79,59 +82,97 @@ public EventGroupBuildingBlock CreateFor(Simulation simulation)
...
-            _parameterIdUpdater.UpdateBuildingBlockId(eventGroup, eventTemplate);
+         _eventGroupBuildingBlock.Add(eventGroup);
+      }
 
-            _eventGroupBuildingBlock.Add(eventGroup);
+      /// <summary>
+      ///    Collects unique start times grouped by PKSimEvent from both standalone event mappings
+      ///    and protocol event placeholder mappings. Duplicate start time values are removed.
+      /// </summary>
+      private Cache<string, EventStartTimeCollection> collectEventStartTimes()
```
msevestre, line 128: Combining the standalone evnet and the protocal event. Uniquenes by event name and by start time withing the grou[p. Right now two event starting a tthe same time will result in one event being dropped. Easy to make it an error

src/PKSim.Core/Services/EventBuildingBlockCreator.cs
```diff
@@ -242,5 +282,26 @@ private static void setParticleRadiusDistributionParametersToLockedAndInvisible(
...
+         public IReadOnlyList<IParameter> UniqueStartTimes => _startTimes;
+
+         public EventStartTimeCollection(PKSimEvent pkSimEvent)
+         {
+            Event = pkSimEvent;
+         }
+
+         public void AddStartTime(IParameter startTime)
+         {
+            if (_seenValues.Add(startTime.Value))
```
msevestre, line 302: that's the place where I avoid duplicate by time

src/PKSim.Core/Services/EventBuildingBlockCreator.cs
Yuri05: Make it an error please!

Yuri05: Actually, there are some situations where we could have 2 different events at the same time (we have not only **meal** events - compare https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3492#issuecomment-4198176932). 
So dropping an event if we have >1 events with the same start time might be a problem.

**For now, I would say: throw an error, if any 2 events start at the same time.**

For later: I would extend the event definition with the event **type**. And then:
* 2 events of the same type with the same start time => Error
* 2 events with the different type and the same start time => both events must be created

Yuri05: > For later: I would extend the event definition with the event **type**. And then:
> 
>     * 2 events of the same type with the same start time => Error
> 
>     * 2 events with the different type and the same start time => both events must be created

Issue created: https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3497

msevestre: I don’t understand this.  I am grouping by event name so wr can’t have the same event twice at the same time but we can have many different event at the same time. When two meals event. We can’t have two breakfast at the same time. Why would we want this ?

Yuri05: >When two meals event. We can’t have two breakfast at the same time. Why would we want this ?

We don't. Could happen due to the user settings though (the simplest case - like below) - then it should produce an error.
grafik

msevestre: Well right now it won’t. It will be swallowed lol

msevestre: >* 2 events with the different type and the same start time => both events must be created

that's already the case. I think we are talking about sthg else or e are just saying the same in circle. Are you an AI @Yuri05

msevestre:
@yuri05 Can we merge this one and we will throw on same time in 3497?

Yuri05:
> @Yuri05 Can we merge this one and we will throw on same time in 3497?

sure

Yuri05 (approved):

Issue #3422, opened 2026-03-02
Feature: Compound Overwrite Parameter Sets


msevestre:
# Compound Overwrite Parameter Sets

## Problem Statement

Users working with PK-Sim simulations frequently override **compound-dependent simulation parameters** (BuildingBlockType = Simulation) that are originally formula-based with constant values.

_Compound-dependent_ means in this context: the initial formula of the simulation parameter depends on a compound which is used in the simulation configuration as a building block.

These customized parameter values are valuable and represent validated configurations, but currently they exist only within the simulation context and cannot be:

- Saved and reused across different projects
- Shared with other team members
- Applied consistently when creating new simulations with the same compound
- Tracked as part of the compound's validated configuration

This leads to repeated manual work, risk of inconsistency, and loss of institutional knowledge when simulation-specific parameter overrides cannot be persisted back to the compound building block.

## Solution

Introduce **Overwrite Parameter Sets** as a new concept attached to Compounds. An Overwrite Parameter Set is a named collection of parameter values that can be:

1. **Created** by committing changed (compound-dependent) simulation parameters back to their respective compounds
2. **Selected** when configuring a simulation (one active Overwrite Parameter Set per compound, optional)
3. **Applied** automatically during simulation model construction
4. **Edited** within the Compound's new Overwrite Parameter Sets tab (no new entries should be created from the tab)

The system will track uncommitted (compound-dependent) simulation parameter changes with a visual indicator (orange status) to help users identify parameters that could be committed to Overwrite Parameter Sets.

## User Stories

1. As a user, I want to save my optimized simulation parameters back to the compound, so that I can reuse them in future simulations without manual re-entry.

2. As a user, I want to see a visual indicator when my simulation has uncommitted compound-related parameter changes, so that I know there are values I might want to persist.

3. As a user, I want to commit all my changed (compound-dependent) simulation parameters to compounds in a single action, so that I don't have to save them one by one.

4. As a user, I want to review which parameters will be committed before confirming, so that I can verify the changes are correct. I should be able to deselect some parameters from the commit process.

5. As a user, I want to choose whether to create a new Overwrite Parameter Set or update an existing one when committing, so that I can organize my parameter configurations appropriately. The choice is per compound, not global.

6. As a user, I want to exclude specific compounds from the commit action, so that I only save parameters for the compounds I intend to update.

[...]

PR #3495, opened 2026-04-05
Add SimulationParameterChangeTracker for compound parameter tracking

### LINKED ISSUE #3430: Change tracking for simulation parameters
## Parent PRD

#3422

## What to build

Implement tracking of uncommitted compound-dependent simulation parameter changes. This enables the visual indicator and commit workflow.

**SimulationParameterChangeTracker class:**
- Maintains list of ObjectPath for changed simulation parameters
- `HasUncommittedChanges` property (true when list is non-empty)
- Methods: Track(path), Untrack(path), Clear(), GetChangedPaths()

**Integration with simulation:**
- Add `SimulationParameterChangeTracker` to Simulation
- Persisted with simulation (survives save/load)

**Command integration:**
- Commands that modify compound-dependent simulation parameters add paths to tracker
- Uses path matching logic from #3427 to identify compound-dependent parameters
- Reset parameter command removes path from tracker

**Path matching behavior:**
- Only track parameters that belong to compounds used as building blocks
- Do not track parameters for unnamed metabolites or other simulation-created compounds

## Acceptance criteria

- [ ] `SimulationParameterChangeTracker` class exists with Track/Untrack/Clear/GetChangedPaths methods
- [ ] `HasUncommittedChanges` property returns true when there are tracked changes
- [ ] Tracker is attached to Simulation
- [ ] Tracker persists with project save/load
- [ ] Modifying a compound-dependent simulation parameter adds it to tracker
- [ ] Resetting a parameter removes it from tracker
- [ ] Parameters for unnamed metabolites are not tracked
- [ ] Unit tests for tracker behavior

## Blocked by

- #3427

## User stories addressed

- User story 2: visual indicator for uncommitted changes (foundation)
- User story 19: reset parameter removes from uncommitted changes list
- User story 20: uncommitted changes persisted when saving project

msevestre:
## Summary
- Adds `SimulationParameterChangeTracker` class to track uncommitted compound-dependent simulation parameter changes
- Includes BDD unit tests covering: track, untrack, clear, dedup, and empty state scenarios

## References
- Fixes #3430
- Parent PRD: #3422
- Depends on: #3427

## Test plan
- [ ] CI tests pass for `SimulationParameterChangeTrackerSpecs`
- [ ] T2-T4 (Simulation integration, command integration) to follow in subsequent commits

## Release Notes

* **New Features**
  * Added parameter change tracking to simulations, enabling identification of modified parameters awaiting commit.
  * Parameter change data is preserved across simulation clones and serialization operations.

* **Tests**
  * Comprehensive test coverage for parameter change tracking functionality, including edge cases for undo/redo operations.

### DIFF
--- a/src/PKSim.Core/Commands/EditParameterCommand.cs
+++ b/src/PKSim.Core/Commands/EditParameterCommand.cs
@@ -43,7 +43,7 @@ namespace PKSim.Core.Commands
          ExecuteUpdateParameter(_parameter, context);
 
          //Track compound-dependent simulation parameter changes
-         trackParameterChange(_parameter, context);
+         trackCompoundParameterChange(_parameter, context);
 
          //Once values have been updated, update dependent objects
          UpdateDependenciesOnParameter(_parameter, context);
@@ -113,7 +113,7 @@ namespace PKSim.Core.Commands
          }
       }
 
-      private void trackParameterChange(IParameter parameter, IExecutionContext context)
+      private void trackCompoundParameterChange(IParameter parameter, IExecutionContext context)
       {
          if (parameter.BuildingBlockType != PKSimBuildingBlockType.Simulation)
             return;
--- a/src/PKSim.Core/Commands/EditParameterCommand.cs
+++ b/src/PKSim.Core/Commands/EditParameterCommand.cs
@@ -12,6 +12,7 @@ namespace PKSim.Core.Commands
 
       public string ParameterId { get; }
       private bool _alteredOn;
+      private bool _trackedByThisCommand;
 
       public string SimulationId { get; protected set; }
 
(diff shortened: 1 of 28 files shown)


msevestre:
@coderabbitai full review

msevestre:
@Yuri05 @rwmcintosh Building upon the existing code. This is the code that is tracking changes in the simulation. I am not sure at this stage if "path" is enough or if we need to also store the compound name. I opted for not storing the component for now as it's a simple quick lookup.

src/PKSim.Core/Commands/EditParameterCommand.cs
```diff
@@ -132,6 +182,7 @@ public override void UpdateInternalFrom(IBuildingBlockChangeCommand originalComm
          base.UpdateInternalFrom(originalCommand);
          var editChangeCommand = originalCommand.DowncastTo<EditParameterCommand>();
          _alteredOn = editChangeCommand._alteredOn;
+         _trackedByThisCommand = editChangeCommand._trackedByThisCommand;
```
msevestre, line 185: @rwmcintosh @Yuri05 This is the trick to be able to undo the command properly

rwmcintosh:
> @Yuri05 @rwmcintosh Building upon the existing code. This is the code that is tracking changes in the simulation. I am not sure at this stage if "path" is enough or if we need to also store the compound name. I opted for not storing the component for now as it's a simple quick lookup.

You left a TODO for that in the code but if we don't need it do you think the TODO will be revisited again? Otherwise maybe remove the TODO and if we need it we'll find out then. Just to avoid a TODO comment that doesn't get cleaned up.

rwmcintosh (approved):

src/PKSim.Core/Snapshots/Mappers/SimulationMapper.cs
```diff
@@ -122,6 +122,7 @@ public override async Task<SnapshotSimulation> MapToSnapshot(ModelSimulation sim
          snapshot.HasResults = simulation.HasResults;
          snapshot.AlteredBuildingBlocks = alteredBuildingBlocksIn(simulation);
          snapshot.OverwriteParameterSetSelections = await _overwriteParameterSetSelectionMapper.MapToSnapshots(simulation.OverwriteParameterSetSelections.Selections, project);
+         snapshot.ChangedParameterPaths = changedParameterPathsFrom(simulation);
```
Yuri05: Do we need to save these parameter paths separately in the snapshot?

This would be a subset of parameter paths for `snapshot.Parameters` (L117), correct? 
So when loading a simulation from snapshot, we could reanalyze each of them for the tracker (for the older snapshots created with ≤v12, this needs to be done anyway).

Advantage of the current solution: no need to reanalyze parameters changed in the simulation (the time gain during the simulation creation is probably not that high though). 
Disadvantages: duplication of the information in the snapshot; direct snapshot manipulation (e.g. via [osp.snapshots](https://github.com/esqLABS/osp.snapshots) etc. becomes tricky and error-prone).

msevestre: Hum I did not think about this. So you are suggesting to not saved this information at all and rather reconstruct? What would it be? we load all the changes parameters in the simulation, see if they are "Compound" matchng (e.g SImulation type and belong to compound) and add them ?

Yuri05: > Hum I did not think about this. So you are suggesting to not saved this information at all and rather reconstruct?

exactly.

> What would it be? we load all the changes parameters in the simulation, see if they are "Compound" matchng (e.g SImulation type and belong to compound) and add them ?

Yes, something like this.

PR #3498, opened 2026-04-07
Fixes #1296 Add DTO layer to UserSettingsView

### LINKED ISSUE #1296: Reduce max. number of decimal places allowed for display
Rather unsurprisingly, a very reliable way to make PK-Sim crash is to set the number of decimal places in the general options to the maximum value allowed: 999999999 (!)
More surprisingly, even if you type in this value and then click on Cancel, PK-Sim crashes.

rwmcintosh:
## Summary
- Adds `UserSettingsDTO` so that edits in the User Settings dialog are buffered and only applied when the user clicks OK. Cancel now correctly discards changes.
- Extracts mapping logic into `UserSettingsToUserSettingsDTOMapper` following existing mapper conventions.
- Moves validation rules from `UserSettings` to the DTO layer, and adds a new rule limiting decimal places to 0–15.

Fixes #1296

## Test plan
- [x] Open User Settings, change a setting (e.g. skin, color), click Cancel — verify the change is not applied
- [x] Open User Settings, change a setting, click OK — verify the change is applied and persisted
- [x] Verify validation: decimal places > 15 prevents OK, empty default names prevent OK, core count out of range prevents OK

## Summary by CodeRabbit

* **New Features**
  * Settings UI now uses an intermediate settings DTO for editing, improving edit/save flow and validation feedback.
* **Bug Fixes**
  * Added validation enforcing decimal places between 0 and 15 and ensuring maximum cores is within allowed range.
* **Tests**
  * Added automated tests covering settings validation and presenter save/edit behavior.
* **Chores**
  * Refactored user settings infrastructure and validation wiring for clearer configuration handling.

### DIFF
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -1397,6 +1397,7 @@ namespace PKSim.Assets
 
          public static readonly string AtLeastTwoBinsRequired = "At least two bins are required";
          public static readonly string ExceededMaximumOfBins = "Maximum of bins exceeded";
+         public static readonly string DecimalPlaceMustBeBetween0And15 = "Number of decimal places must be between 0 and 15";
       }
 
       public static class UI
--- a/src/PKSim.Presentation/Presenters/UserSettingsPresenter.cs
+++ b/src/PKSim.Presentation/Presenters/UserSettingsPresenter.cs
@@ -1,17 +1,19 @@
 using System.Collections.Generic;
 using System.Linq;
-using PKSim.Assets;
-using PKSim.Core;
-using PKSim.Core.Model;
-using PKSim.Core.Repositories;
-using PKSim.Presentation.Services;
-using PKSim.Presentation.Views;
+using OSPSuite.Assets;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Services;
 using OSPSuite.Presentation.Presenters;
 using OSPSuite.Presentation.Services;
-using OSPSuite.Assets;
 using OSPSuite.Utility;
+using PKSim.Assets;
+using PKSim.Core;
+using PKSim.Core.Model;
+using PKSim.Core.Repositories;
+using PKSim.Presentation.DTO;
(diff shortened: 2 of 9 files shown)


src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -1397,6 +1397,7 @@ public static string PatternShouldContainOneIterationPattern(string iterationPat
 
          public static readonly string AtLeastTwoBinsRequired = "At least two bins are required";
          public static readonly string ExceededMaximumOfBins = "Maximum of bins exceeded";
+         public static readonly string DecimalPlaceMustBeBetween0And15 = "Number of decimal places must be between 0 and 15";
```
rwmcintosh, line 1400: Is 15 enough? Default is 2.

src/PKSim.Presentation/Presenters/UserSettingsPresenter.cs
```diff
@@ -206,6 +212,7 @@ public IReadOnlyList<Scalings> AllScalings()
 
       public void SaveSettings()
       {
+         _mapper.UpdateUserSettingsFrom(_userSettingsDTO, _userSettings);
```
rwmcintosh, line 215: We only update the real `UserSettings` if the user selects 'OK'. Otherwise changes are discarded.

msevestre (commented):
Why do we need a DTO? Can't we juste create a clone of the USerSEttings?

rwmcintosh:
> Why do we need a DTO? Can't we juste create a clone of the USerSEttings?

yeah we could do that instead.

rwmcintosh:
> > Why do we need a DTO? Can't we juste create a clone of the USerSEttings?
> 
> yeah we could do that instead.

I think the best implementation is some kind of data only object at the end of the day. In PKSim UserSettings is backed by a few injected services like DockManager, RibbonManager, SkinManager. Those are DevExpress objects and while we can just create backing fields for this when working with the clone, I think it's less clear than what is in this PR now.

msevestre:
ok. feels like a lot of code for such a small gain.

PR #3499, opened 2026-04-09
remove reporting

rwmcintosh:
I think we just remove everything now right?


Yuri05 (approved):

Yuri05:
> I think we just remove everything now right?

I think so

PR #3500, opened 2026-04-10
update core and add convenience script

rwmcintosh:
## Summary by CodeRabbit

* **New Features**
  * Added automation script for local NuGet package packing and integration with R workflows.

* **Chores**
  * Updated NuGet package dependencies to version 13.0.94 across all projects.
  * Removed deprecated reporting presenter class.
  * Cleaned up test infrastructure.


Yuri05 (approved):

msevestre:
what's the convenience script about? how do we use it? maybe a quick explanation would be good

msevestre:
I meant specifically in the readme
Why would we use this ? What does it do compared to what we have already

msevestre:
I think for next time we could create an issue to describe the problem fixed. We don’t do it for core update (ok) but this is more than this

rwmcintosh:
> I meant specifically in the readme Why would we use this ? What does it do compared to what we have already

The readme is not the place for a description of this, it's only for developers. The only comparable is a script in Core that does the same thing - updates mobi and pksim from a local nuget.

That is also not documented in the readme. But it might be useful as an [update to dev documentation](https://dev.open-systems-pharmacology.org/.net-code-specifics/debugging#creating-local-nuget-packages-from-core)

PR #3526, opened 2026-04-30
Fixes #3436 Delete OverwriteParameterSet from compound

### LINKED ISSUE #3436: Delete OverwriteParameterSet and set default
## Parent PRD

#3422

## What to build

Enable deleting entire OverwriteParameterSets and marking one as default for a compound.

**Delete entire set:**
- Context menu or action to delete an entire OverwriteParameterSet
- **Cannot delete an OverwriteParameterSet that is currently used in any simulation** (show error message)
- Must check all simulations in the project before allowing deletion

**Set as default:**
- Context menu: "Set as default Overwrite Parameter Set"
- A compound does not require a default (can have none)
- Only one set can be default per compound
- Setting a new default clears the previous default

**Commands:**
- `RemoveOverwriteParameterSetFromCompoundCommand`
- `SetDefaultOverwriteParameterSetCommand`

**Undo/redo:**
- Full undo/redo support for both operations

## Acceptance criteria

- [ ] Delete action available for entire OverwriteParameterSet
- [ ] Cannot delete a set that is used in any simulation (error message shown)
- [ ] "Set as default" action available
- [ ] Only one set can be default per compound
- [ ] Setting new default clears previous default
- [ ] Compound can have no default (clearing default works)
- [ ] `RemoveOverwriteParameterSetFromCompoundCommand` implemented with undo/redo
- [ ] `SetDefaultOverwriteParameterSetCommand` implemented with undo/redo
- [ ] Changes are persisted when project is saved

## Blocked by

- #3429

## User stories addressed

- User story 10: delete entire OverwriteParameterSets
- User story 11: mark one OverwriteParameterSet as default for a compound

rwmcintosh:
Some of the intended functionality from here had already been implemented alongside previous commits - such as marking a set as default, so those changes don't appear here.

## Summary
- Adds a per-row delete button on the overwrite parameter sets grid in the compound editor.
- Blocks deletion (with a clear message listing the offending simulations) when any simulation has the set selected; otherwise prompts for confirmation and runs `RemoveOverwriteParameterSetFromCompoundCommand` (with full undo/redo via the existing inverse `Add…` command).
- Sets are matched by name within the compound — simulations hold cloned compounds (`UsedBuildingBlock.BuildingBlock = cloneManager.Clone(...)`), so id/reference equality misses; name + compound name is the durable identifier.
- Simulations are lazy-loaded before their `OverwriteParameterSetSelections` are inspected, since the property is empty until content is hydrated.

## Test plan
- [x] Open a compound with one or more overwrite parameter sets — each row has a delete button.
- [x] Click delete on a set that is **not** referenced by any simulation → confirmation dialog → set is removed.
- [x] Undo → set comes back; redo → set goes again.
- [x] Click delete on a set that **is** selected by a simulation (loaded or unloaded) → error dialog naming the blocking simulation(s); set remains.
- [x] When the message lists multiple simulations they appear as a `-` bullet list with blank lines around it; a single simulation appears inline in single quotes.
- [x] A same-named set in a different compound does **not** block deletion.

image

## Summary by CodeRabbit

* **New Features**
  * Added the ability to delete overwrite parameter sets from the UI, with a confirmation prompt.
* **Bug Fixes / Safety**
  * Prevents deletion of parameter sets that are referenced by simulations and shows an informative error listing those simulations.
* **Tests**
  * Added coverage for delete scenarios, confirmations, and blocking behavior when referenced by simulations.

### DIFF
--- a/src/PKSim.Core/Services/OverwriteParameterSetTask.cs
+++ b/src/PKSim.Core/Services/OverwriteParameterSetTask.cs
@@ -5,7 +5,6 @@ using OSPSuite.Core.Domain.Services;
 using OSPSuite.Utility.Extensions;
 using PKSim.Core.Commands;
 using PKSim.Core.Model;
-using PKSim.Core.Repositories;
 
 namespace PKSim.Core.Services;
 
@@ -31,16 +30,16 @@ public interface IOverwriteParameterSetTask
 public class OverwriteParameterSetTask : IOverwriteParameterSetTask
 {
    private readonly IExecutionContext _executionContext;
-   private readonly IBuildingBlockRepository _buildingBlockRepository;
+   private readonly IBuildingBlockInProjectManager _buildingBlockInProjectManager;
    private readonly ILazyLoadTask _lazyLoadTask;
 
    public OverwriteParameterSetTask(
-      IExecutionContext executionContext, 
-      IBuildingBlockRepository buildingBlockRepository, 
+      IExecutionContext executionContext,
+      IBuildingBlockInProjectManager buildingBlockInProjectManager,
       ILazyLoadTask lazyLoadTask)
    {
       _executionContext = executionContext;
-      _buildingBlockRepository = buildingBlockRepository;
+      _buildingBlockInProjectManager = buildingBlockInProjectManager;
       _lazyLoadTask = lazyLoadTask;
    }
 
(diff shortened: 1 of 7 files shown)


src/PKSim.Core/Services/OverwriteParameterSetTask.cs
```diff
@@ -58,4 +71,24 @@ public ICommand SetIsDefault(OverwriteParameterSet overwriteParameterSet, Compou
...
+   
+   private IReadOnlyList<Simulation> simulationsUsing(OverwriteParameterSet overwriteParameterSet, Compound compound)
+   {
+      var simulations = _buildingBlockRepository.All<Simulation>();
+      simulations.Each(_lazyLoadTask.Load);
+      return simulations
+         .Where(simulation => simulation.OverwriteParameterSetSelections.Selections.Any(x =>
+            string.Equals(x.CompoundName, compound.Name) &&
+            string.Equals(x.OverwriteParameterSet?.Name, overwriteParameterSet.Name)))
+         .ToList();
```
Yuri05: I'm not sure if it's possible to estimate if an `OverwriteParameterSet `is used in a simulation without LazyLoading the simulation. 
But it is at least possible for the **compound-BB**.

So I would first get only the simulations which are using the `compound` and only after this LazyLoad all of them.

Yuri05 (changes requested):
s. https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3526/changes#r3168408084

Yuri05 (approved):

PR #3553, opened 2026-05-22
Fixes #3096 Automatically add observed data to simulation comparisons

### LINKED ISSUE #3096: Automatically add observed data for comparisons
It would be nice that when we compare two simulations (or more) in PK-Sim, we automatically get the observed data displayed as well (same as in the simulations), without having to drag and drop all the data from the observations BB again

rwmcintosh:
## Summary

When a simulation is added to an individual or population simulation comparison, the comparison now automatically displays the observed data the simulation references — instead of requiring the user to re-drag every observed data set from the building block tree.

Curve styling for observed data is also templated from each simulation's own time-profile chart, so colors/line styles/symbols carry over.

## Implementation

Centralized in a new `IObservedDataInComparisonTask` (PKSim.Core) so every entry point goes through the same logic:

| Entry point | What triggers the auto-add |
|---|---|
| `SimulationComparisonTask.CreateIndividualSimulationComparison(sim)` | resolves simulation's observed data and registers it on the new chart |
| `IndividualSimulationComparisonPresenter` drag-drop / open | registers on chart + creates editor curves + templates source curve options |
| `SimulationComparisonTask.CreatePopulationSimulationComparison` (via the creator) | default time-profile analysis gets data + templated `ObservedDataCurveOptions` |
| `SimulationAnalysisCreator.CreatePopulationAnalysisFor(comparison, TimeProfile)` | every later "Add time profile analysis" on an existing comparison gets the same treatment |
| `SimulationComparisonTask.ConfigurePopulationSimulationComparison` | re-templates existing time-profile analyses after a sim-selection change |

Other population analysis types (BoxWhisker, Scatter, Range) are deliberately left alone — they don't render observed data.

The first source curve encountered wins on color (first-sim-wins for shared observed data). User customizations are preserved.

## Test plan

- [x] Unit tests pass (`ObservedDataInComparisonTaskSpecs`, `SimulationAnalysisCreatorSpecs`, `SimulationComparisonTaskSpecs`, `IndividualSimulationComparisonPresenterSpecs`)
- [x] Manual: drag an individual sim with observed data into a new comparison — observed data curves appear with sim's source styling
- [x] Manual: create an individual comparison from a sim that uses observed data — data shown when comparison opens
- [x] Manual: create a new population comparison with sims that use observed data — default time-profile analysis includes the data
- [x] Manual: add a second time-profile analysis to an existing population comparison — observed data added there too
- [x] Manual: reconfigure a population comparison to include a new sim — existing time-profile analyses pick up its observed data

## Summary by CodeRabbit

* **New Features**
  * Observed data is now automatically resolved and included in individual and population simulation comparisons and time‑profile analyses.
  * Curve styling from source analyses is preserved and templated onto matching observation curves in comparison charts.

* **Tests**
  * Added comprehensive specs covering observed‑data registration, curve templating, and applying source curve options in comparisons.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Open-Systems-Pharmacology/PK-Sim/pull/3553?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/PKSim.Core/Services/SimulationAnalysisCreator.cs
+++ b/src/PKSim.Core/Services/SimulationAnalysisCreator.cs
@@ -5,6 +5,7 @@ using OSPSuite.Utility.Extensions;
 using OSPSuite.Utility.Visitor;
 using PKSim.Core.Chart;
 using PKSim.Core.Model;
+using PKSim.Core.Model.PopulationAnalyses;
 
 namespace PKSim.Core.Services
 {
@@ -29,16 +30,24 @@ namespace PKSim.Core.Services
       private readonly IPKSimChartFactory _chartFactory;
       private readonly ICoreUserSettings _userSettings;
       private readonly ICloner _cloner;
+      private readonly IObservedDataInComparisonTask _observedDataInComparisonTask;
       private ISimulationAnalysis _simulationAnalysis;
 
-      public SimulationAnalysisCreator(IPopulationSimulationAnalysisStarter populationSimulationAnalysisStarter,
-         IExecutionContext executionContext, IContainerTask containerTask, IPKSimChartFactory chartFactory,
-         ICoreUserSettings userSettings, ICloner cloner) : base(containerTask, executionContext)
+      public SimulationAnalysisCreator(
+         IPopulationSimulationAnalysisStarter populationSimulationAnalysisStarter,
+         IExecutionContext executionContext, 
+         IContainerTask containerTask, 
+         IPKSimChartFactory chartFactory,
+         ICoreUserSettings userSettings, 
+         ICloner cloner, 
+         IObservedDataInComparisonTask observedDataInComparisonTask) : 
+         base(containerTask, executionContext)
       {
          _populationSimulationAnalysisStarter = populationSimulationAnalysisStarter;
          _chartFactory = chartFactory;
          _userSettings = userSettings;
          _cloner = cloner;
+         _observedDataInComparisonTask = observedDataInComparisonTask;
       }
 
       public ISimulationAnalysis CreateAnalysisFor(Simulation simulation)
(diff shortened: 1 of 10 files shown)


src/PKSim.Core/Services/SimulationAnalysisCreator.cs
```diff
@@ -80,6 +89,10 @@ public ISimulationAnalysis CreatePopulationAnalysisFor(IPopulationDataCollector
       {
          var populationSimulationAnalysis =
             _populationSimulationAnalysisStarter.CreateAnalysisForPopulationSimulation(populationDataCollector, populationAnalysisType);
+
+         if (populationDataCollector is PopulationSimulationComparison comparison && populationSimulationAnalysis is TimeProfileAnalysisChart timeProfileChart)
+            _observedDataInComparisonTask.AddObservedDataToTimeProfile(comparison, timeProfileChart);
+
```
Yuri05, line 95: Why is this asymmetric: for pop simulations, `observedDataInComparisonTask.AddObservedDataToTimeProfile()` is called here, but for the individual simulations in SimulationComparisonTask.cs below?

rwmcintosh: That's because of the differences in these comparisons. When you start a pop compare you have to pick the simulations ahead of time and the pop sim is created after. 

image

Then you can add another chart later, it will use the existing simulations. A pop comparison can contain multiple charts, 
image


but an individual comparison is **a** curvechart. You add simulations to it by drag drop.

image

It has the curve editor and data browser all within, not so for the population comparison.

I wanted them to be as symmetrical as possible, but in the end they are pretty different and are created and manipulated differently

Yuri05: oki, thx

Yuri05 (approved):

PR #3558, opened 2026-05-25
Fixes #3556 Load simulations from a PK-Sim snapshot

### LINKED ISSUE #3556: Load simulations from a PK-Sim snapshot
In PKSim.R, add an API entry for managing snapshots.

Given a snapshot, create Simulations and return then as Simulation objects.

Either all simulations (corresponds to opening a snapshot in PK-Sim), or only for a given simulation name, if possible.

We want to unburden the caller from creating the `SnapshotRunOptions` object and instead use method parameter for each.

Only InputFolder and OutputFolder are required in the API. The others have good defaults in the SnapshotRunOptions construction.

```
      public bool RunSimulations { get; set; } = true;
      public SnapshotExportMode ExportMode { get; set; } = SnapshotExportMode.Snapshot;
      public NotificationType NotificationType { get; set; } = NotificationType.Info | NotificationType.Error;

      public string InputFolder
      {
         get => _inputFolder;
         set => SetProperty(ref _inputFolder, value);
      }

      public string OutputFolder
      {
         get => _outputFolder;
         set => SetProperty(ref _outputFolder, value);
      }

      public IReadOnlyList<string> Folders { get; set; }
```

We should also offer a more convenient `RunSnapshot` with the same parameters. We will need to keep the existing `RunSnapshot` for backward compatibility

rwmcintosh:
Fixes #3556

## Summary
- Adds a new `ISnapshotTask` to PKSim.R with `LoadSimulationsFromSnapshot(snapshotFile, params string[] simulationNames)` — returns every simulation in the snapshot by default, or just the ones matching the provided names (equivalent to opening the snapshot in PK-Sim).
- Adds a more convenient `RunSnapshot(inputFolder, outputFolder, runSimulations=true, exportMode=Snapshot, params string[] folders)` overload that builds the `SnapshotRunOptions` for the caller. The existing `Api.RunSnapshot(SnapshotRunOptions)` stays for backward compatibility.
- Exposes the new task via `Api.GetSnapshotTask()` (resolved through the existing R container).
- Test fixture `tests/PKSim.R.Tests/Data/Atazanavir-Model_1Sim.json` was re-exported from PK-Sim with one simulation (`test`) so the new specs exercise both the "load all" and "filter by name" paths.

## Test plan
- [x] `dotnet test tests/PKSim.R.Tests/PKSim.R.Tests.csproj` — 36 / 36 pass locally (5 new SnapshotTask specs + the pre-existing suite).
- [x] `dotnet build PKSim.sln` clean.

## Summary by CodeRabbit

* **New Features**
  * Added a snapshot task API for programmatic snapshot execution.
  * Load simulations from snapshot files, with optional name-based filtering.
  * Run snapshot workflows with configurable export modes and folder-based processing.

* **Tests**
  * Added integration tests validating loading (all, filtered, none) and export behavior.
  * Updated snapshot test data to cover expanded model/compound/protocol scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Open-Systems-Pharmacology/PK-Sim/pull/3558?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/PKSim.R/Services/SnapshotTask.cs
+++ b/src/PKSim.R/Services/SnapshotTask.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Linq;
 using OSPSuite.CLI.Core.RunOptions;
 using OSPSuite.CLI.Core.Services;
@@ -40,6 +41,9 @@ namespace PKSim.R.Services
       public Simulation[] LoadSimulationsFromSnapshot(string snapshotFile, params string[] simulationNames)
       {
          var project = _snapshotTask.LoadProjectFromSnapshotFileAsync(snapshotFile).Result;
+         if (project == null)
+            return Array.Empty<Simulation>();
+
          var allSimulations = project.All<CoreSimulation>();
 
          var matched = simulationNames.Length == 0
--- a/src/PKSim.R/Services/SnapshotTask.cs
+++ b/src/PKSim.R/Services/SnapshotTask.cs
@@ -1,0 +1,65 @@
+using System.Linq;
+using OSPSuite.CLI.Core.RunOptions;
+using OSPSuite.CLI.Core.Services;
+using PKSim.Core.Model;
+using CoreSnapshotTask = PKSim.Core.Snapshots.Services.ISnapshotTask;
+
+namespace PKSim.R.Services
+{
+   public interface ISnapshotTask
+   {
+      /// <summary>
+      ///    Loads the simulations stored in the snapshot file at <paramref name="snapshotFile" />. If
+      ///    <paramref name="simulationNames" /> is provided, only the simulations whose names match are returned.
+      ///    When no name is supplied, every simulation contained in the snapshot is returned (equivalent to opening the
+      ///    snapshot in PK-Sim).
+      /// </summary>
+      Simulation[] LoadSimulationsFromSnapshot(string snapshotFile, params string[] simulationNames);
+
+      /// <summary>
+      ///    Runs the snapshot workflow for the given <paramref name="inputFolder" /> and <paramref name="outputFolder" />
+      ///    using sensible defaults for the remaining options. Optional <paramref name="folders" /> can be provided to
+      ///    process each folder pair instead of the single input/output pair.
+      /// </summary>
+      void RunSnapshot(string inputFolder, string outputFolder, bool runSimulations = true,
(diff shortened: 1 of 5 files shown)


src/PKSim.R/Services/SnapshotTask.cs
```diff
@@ -0,0 +1,65 @@
...
+
+      public Simulation[] LoadSimulationsFromSnapshot(string snapshotFile, params string[] simulationNames)
+      {
+         var project = _snapshotTask.LoadProjectFromSnapshotFileAsync(snapshotFile).Result;
+         var allSimulations = project.All<Simulation>();
+
+         if (simulationNames.Length == 0)
+            return allSimulations.ToArray();
+
+         return allSimulations.Where(x => simulationNames.Contains(x.Name)).ToArray();
```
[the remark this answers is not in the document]
rwmcintosh: keeping simulationsNames unguarded.

src/PKSim.R/Services/SnapshotTask.cs
```diff
@@ -0,0 +1,72 @@
...
+         {
+            InputFolder = inputFolder,
+            OutputFolder = outputFolder,
+            RunSimulations = runSimulations,
+            ExportMode = exportMode,
+            Folders = folders
+         };
+
+         _snapshotRunner.RunBatchAsync(runOptions).GetAwaiter().GetResult();
+      }
```
Yuri05, line 70: If I understand it correctly, `inputFolder` and `outputFolder` are ignored in case `folders` are not empty?
If so, please check that in such a case both `inputFolder` and `outputFolder` are empty and produce a warning (or error?) if not

Yuri05: > If I understand it correctly, `inputFolder` and `outputFolder` are ignored in case `folders` are not empty? 

Checked it in PK-Sim code: yep, it's the case.

rwmcintosh: I didn't look at how it works, this is just a convenient way to construct the object for R workflows. Processing is unchanged under the hood

Yuri05: if it's convenient for R, than fine for me

Yuri05 (approved):

PR #3561, opened 2026-05-27
Fixes #3530 Modified distributed parameters not exported to snapshot

### LINKED ISSUE #3530: Modified parameters not exported to the snapshot.
[Individual_Volumes_orig.zip](https://github.com/user-attachments/files/27479071/Individual_Volumes_orig.zip)
* Open the only simulation in the project attached (The project was recreated from Snapshot in 12.2 or 12.3) and scale all volumes by the factor 2. Which results in the view below:
   

* Commit the modified Individual from the simulation to the building block. The result in the individual looks like below. 
  Which is already strange (e.g. the Reset button is available only for **some** of the modified parameters). All values/percentiles are correct though.
   

* Export the project to snapshot.
* Save the project as e.g. **Individual_Volumes_modified** (here the saved project [Individual_Volumes_modified.zip](https://github.com/user-attachments/files/27479600/Individual_Volumes_modified.zip))

* Reload the project from Snapshot
  * Problem 1: warnings during Import from Snapshot
     ```
     Information: Loading Project from snapshot file 'C:\Individual_Volumes_orig.json'
     Warning: Snapshot parameter 'CYP3A4|Ontogeny factor GI table' was not found in 'Nyberg 1980 0.5m infant  (mean)'.
     Warning: Snapshot parameter 'CYP3A4|Ontogeny factor table' was not found in 'Nyberg 1980 0.5m infant  (mean)'.
     Warning: Snapshot parameter 'P-gp|Ontogeny factor GI table' was not found in 'Nyberg 1980 0.5m infant  (mean)'.
     Warning: Snapshot parameter 'P-gp|Ontogeny factor table' was not found in 'Nyberg 1980 0.5m infant  (mean)'.
     Information: Loading simulation 'Nyberg 1980 0.5m infants (mean)' (1/1)...
     Information: Project loaded from snapshot
     ```
  * Problem 2: in the inidividual, **some** of the modified volume values (along with their percentiles) are totally off (e.g. **Fat**, **Heart**, ...). seems to be exactly the organs which did not have a Reset button already in the 2nd step (commit individual to BB)
    

  * Problem 3: in the simulation, percentiles for the correctly imported values disappear

msevestre:
## Summary
- Reorder `ParameterSetUpdater.UpdateValues` to iterate distributed parameters **before** their non-distributed children. Writing a distributed parameter's value cascades to its percentile child via the `DistributedParameter.Value` setter, so the child compare-equal afterward and is skipped — preserving the parent's own `IsFixedValue=true` and `IsDefault=false` flags. The previous child-first ordering caused the parent to compare equal (the percentile had already been written, so its computed value matched the source) and skip its own update, leaving `IsDefault=true` and dropping the modified parameter from snapshot export.
- In `ParameterUpdater.setParameterValue`, the explicit value-change branch now passes `shouldUpdateDefaultStateAndValueOriginForDefaultParameter: true` instead of `!sourceParameter.IsDefault`. When the source had retained `IsDefault=true` (which happens whenever the value was set via the property setter rather than through a parameter task command), the previous code prevented the target from being marked non-default. The early-return path for equal values keeps the original `!source.IsDefault` semantics.
- Added an integration test (`When_updating_a_distributed_organ_volume_in_a_simulation_in_the_template_building_block`) that triples Fat Volume in a simulation, commits to the template, and asserts the template's `IsFixedValue`, `IsDefault`, and `ShouldExportToSnapshot()` all reflect the change.

## Test plan
- [x] New integration test passes (4/4 observations)
- [x] Full PKSim.Tests suite passes (5229 passed, 0 failed, 3 skipped)
- [ ] Manually verify in PK-Sim UI that scaling distributed organ volumes in a simulation and committing to the individual now leaves the parent Volume parameter visibly marked as changed (reset button visible)
- [ ] Manually verify that saving the individual to snapshot includes the modified Volume parameters

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected parameter update ordering and default-state handling so distributed parameters (and their percentile children) update reliably; display unit and default-value state are now handled consistently during updates.
* **Tests**
  * Added an integration test verifying distributed organ-volume updates propagate from simulations to template building blocks and adjust parameter state accordingly.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Open-Systems-Pharmacology/PK-Sim/pull/3561?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/PKSim.Core/Services/ParameterSetUpdater.cs
+++ b/src/PKSim.Core/Services/ParameterSetUpdater.cs
@@ -117,8 +117,12 @@ namespace PKSim.Core.Services
       {
          var updateCommands = new PKSimMacroCommand {CommandType = PKSimConstants.Command.CommandTypeEdit, ObjectType = PKSimConstants.ObjectTypes.Parameter};
 
-         //should update non distributed parameters first and then distributed parameter
-         foreach (var sourceParameter in sourceParameters.KeyValues.OrderBy(x => x.Value.IsDistributed()))
+         //Process distributed parameters first. Writing a distributed parameter's value implicitly updates its
+         //percentile child (DistributedParameter.Value setter), so the child compare-equal afterwards and gets
+         //skipped. If we processed the child first instead, writing the percentile would change the parent's
+         //computed Value, the parent would then compare-equal and skip its own update — leaving the parent
+         //with IsDefault=true, which keeps it out of snapshot export (see #3530).
+         foreach (var sourceParameter in sourceParameters.KeyValues.OrderByDescending(x => x.Value.IsDistributed()))
          {
             var targetParameter = targetParameters[sourceParameter.Key];
             if (targetParameter == null)
--- a/src/PKSim.Core/Services/ParameterUpdater.cs
+++ b/src/PKSim.Core/Services/ParameterUpdater.cs
@@ -100,9 +100,11 @@ namespace PKSim.Core.Services
 
       private ICommand setDistributionValue(IParameter sourceParameter, IDistributedParameter targetParameter)
       {
-         //distributed to distributed
+         //distributed to distributed: force the update when the source is fixed, otherwise the parent parameter's
+         //areValuesEqual check short-circuits after the percentile child has already been written and the parent
+         //never gets its IsFixedValue/IsDefault flags updated (see #3530)
          if (sourceParameter is IDistributedParameter sourceDistributedParameter)
-            return setParameterValue(sourceDistributedParameter, targetParameter, false);
+            return setParameterValue(sourceDistributedParameter, targetParameter, forceUpdate: sourceDistributedParameter.IsFixedValue);
 
          //source parameter it not distributed. Is the formula a table parameter? In that case, we are most likely
          //updating from simulation table parameter to distributed parameter in individual
--- a/src/PKSim.Core/Services/ParameterUpdater.cs
+++ b/src/PKSim.Core/Services/ParameterUpdater.cs
@@ -100,11 +100,9 @@ namespace PKSim.Core.Services
 
       private ICommand setDistributionValue(IParameter sourceParameter, IDistributedParameter targetParameter)
       {
-         //distributed to distributed: force the update when the source is fixed, otherwise the parent parameter's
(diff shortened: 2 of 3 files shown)


src/PKSim.Core/Services/ParameterSetUpdater.cs
```diff
@@ -117,8 +117,12 @@ public IPKSimMacroCommand UpdateValues(PathCache<IParameter> sourceParameters, P
       {
          var updateCommands = new PKSimMacroCommand {CommandType = PKSimConstants.Command.CommandTypeEdit, ObjectType = PKSimConstants.ObjectTypes.Parameter};
 
-         //should update non distributed parameters first and then distributed parameter
-         foreach (var sourceParameter in sourceParameters.KeyValues.OrderBy(x => x.Value.IsDistributed()))
+         //Process distributed parameters first. Writing a distributed parameter's value implicitly updates its
+         //percentile child (DistributedParameter.Value setter), so the child compare-equal afterwards and gets
```
msevestre, line 121: It's funny that I actually had made a comment before just to start with non-distributed parameters first, but this is a source of the issue. If we do this, then uh it doesn't look like the uh parameter has changed and therefore it is not update. This is a whole reason.

msevestre (commented):
I checked locally all again and it works. This was ANOTHER issue specifically noticeable when the percentile was not 100% 
Anyways please check it out again @Yuri05

msevestre:
@Yuri05 Also there is this fix made by copilot and I am wondering if this is was part of the problem
https://github.com/Open-Systems-Pharmacology/PK-Sim/commit/d56fda6d

msevestre:
@Yuri05 My individual after commit

image

Yuri05:
> @Yuri05 Also there is this fix made by copilot and I am wondering if this is was part of the problem [d56fda6d](https://github.com/Open-Systems-Pharmacology/PK-Sim/commit/d56fda6d)

yeah. not sure why the corresponding PR https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3335 was closed without merging...

Yuri05 (approved):

PR #3564, opened 2026-05-27
Fixes #3562 Reset of distributed individual parameters in a simulation does not work

### LINKED ISSUE #3562: Reset of distributed individual parameters in a simulation does not work
12.2 and 12.3
* Create an arbitrary simulation (e.g. default individual, any compound, IV).
* Go to the organ volumes
  
* Overwrite the value of e.g. venous blood volume
  
* Click on the Reset button: the value is resetted to a wrong one; the percentile remains unchanged

msevestre:
## Issue

Fixes #3562 — Resetting a distributed individual parameter (e.g. an organ volume such as venous blood volume) in a simulation reset the value to a wrong one and left the percentile unchanged.

## Root cause

The parameters edited/reset in the simulation UI live in `Simulation.Model.Root`. For distributed parameters, every model parameter is built with `DefaultValue == null`:

- A distributed organ volume's `DefaultValue` is computed **per-individual** by `CreateIndividualAlgorithm` and exists only on the individual instance.
- The model build works from static definitions, so it cannot reproduce that value — the model parameter ends up with `DefaultValue == null` (the `Simulation.Individual` synchronization copy still has it).

`DistributedParameter.ResetToDefault()` restores the percentile only through the `if (DefaultValue != null) Value = DefaultValue.Value;` branch (writing the default value re-derives the default percentile). With `DefaultValue` null, that branch is skipped, so the percentile child's constant formula keeps the overwritten percentile and a wrong value is recomputed.

This is distributed-specific in terms of what's user-reachable: `ValueDiffersFromDefault` returns `IsFixedValue` for distributed parameters (so the Reset button is always offered), whereas constant parameters generally carry their `DefaultValue` from static definitions and reset correctly.

## Fix

`SimulationModelCreator.updateSimulationAfterModelCreation` now propagates the individual's `DefaultValue` onto the matching distributed model parameter after model creation — mirroring how molecule parameters are already reconciled there.

## Tests

`SimulationParameterResetSpecs` edits the venous-blood-volume distributed parameter in `Model.Root`, clicks reset (via `IParameterTask`), and asserts the value, percentile, fixed flag, and default flag are all restored. Fails before the fix, passes after.

### DIFF
--- a/tests/PKSim.Tests/IntegrationTests/SimulationParameterResetSpecs.cs
+++ b/tests/PKSim.Tests/IntegrationTests/SimulationParameterResetSpecs.cs
@@ -1,0 +1,87 @@
+using System.Linq;
+using OSPSuite.BDDHelper;
+using OSPSuite.BDDHelper.Extensions;
+using OSPSuite.Core.Domain;
+using OSPSuite.Utility.Container;
+using OSPSuite.Utility.Extensions;
+using PKSim.Core;
+using PKSim.Core.Model;
+using PKSim.Core.Services;
+using PKSim.Infrastructure;
+using static PKSim.Core.CoreConstants.Organ;
+using static OSPSuite.Core.Domain.Constants.Parameters;
+
+namespace PKSim.IntegrationTests
+{
+   public abstract class concern_for_SimulationParameterReset : ContextForSimulationIntegration<IParameterTask>
+   {
+      protected Individual _individual;
+      protected IDistributedParameter _modelVolume;
+      protected double _originalValue;
+      protected double _originalPercentile;
+      protected double _newValue;
+
+      public override void GlobalContext()
+      {
+         base.GlobalContext();
+         _individual = DomainFactoryForSpecs.CreateStandardIndividual();
+         var compound = DomainFactoryForSpecs.CreateStandardCompound();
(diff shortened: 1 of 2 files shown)


msevestre:
@Yuri05 @rwmcintosh Note: it won't work for existing simulation. The default value is gone ok for me

tests/PKSim.Tests/IntegrationTests/SimulationParameterResetSpecs.cs
```diff
@@ -0,0 +1,87 @@
...
+         var project = new PKSimProject();
+         project.AddBuildingBlock(compound);
+         project.AddBuildingBlock(protocol);
+         project.AddBuildingBlock(_simulation);
+         project.AddBuildingBlock(_individual);
+         var workspace = IoC.Resolve<ICoreWorkspace>();
+         workspace.Project = project;
+
+         //The UI edits/resets the parameters in the simulation MODEL (Model.Root). The Simulation.Individual is only
+         //a synchronization copy. The distributed parameters in the model are the ones affected by #3562.
```
Yuri05, line 42: having comments like this is a really good thing 👍

Yuri05 (approved):

PR #3587, opened 2026-06-26
Fixes #3434 Apply overwrite parameter set during model construction

### LINKED ISSUE #3434: Apply OverwriteParameterSet during model construction
## Parent PRD

#3422

## What to build

Integrate OverwriteParameterSet application into the simulation creation/configuration pipeline.

**Model construction integration:**
- OverwriteParameterSet values applied during simulation creation/configuration
- Happens after keyword replacement in the model construction pipeline
- For each selected OverwriteParameterSet:
  - Find matching parameters in simulation by path
  - Apply values from OverwriteParameterSet

**Parameter behavior after application:**
- Affected parameters become BuildingBlockType = Compound
- Subsequent changes to these parameters follow normal compound parameter logic
- Changes trigger standard red status indicator on compound

**Error handling:**
- Invalid/unresolved paths: **Error - simulation cannot be created**
- Clear error message indicating which paths could not be resolved

## Acceptance criteria

- [ ] Selected OverwriteParameterSet values are applied during simulation creation
- [ ] Parameters are correctly matched by path
- [ ] Applied parameters have BuildingBlockType = Compound
- [ ] Subsequent changes to applied parameters show red indicator (normal compound behavior)
- [ ] Simulation creation fails with clear error if paths cannot be resolved
- [ ] Error message identifies which specific paths are invalid
- [ ] Works correctly when "None" is selected (no values applied)
- [ ] Works correctly with multiple compounds each having different OverwriteParameterSet selections

## Blocked by

- #3433

## User stories addressed

- User story 15: parameters applied from OverwriteParameterSet behave as compound parameters
- User story 16: error when paths cannot be resolved, simulation cannot be created

rwmcintosh:
## What

Integrates `OverwriteParameterSet` application into the simulation model construction pipeline (#3434).

After the model is constructed (i.e. after keyword replacement), `SimulationModelCreator.CreateModelFor` applies each compound's selected `OverwriteParameterSet` to the matching simulation parameters:

- Parameters are matched **by path** against the constructed model (same `IContainerTask.CacheAllChildren` path keying used by the commit flow, so paths are portable across simulations).
- Applied parameters become `BuildingBlockType = Compound` and their `Origin.BuilingBlockId` points at the compound building block, so subsequent edits follow normal compound logic (red "altered" indicator, no orange tracking).
- `"None"` selections are skipped.
- If **any** path cannot be resolved, creation aborts with `CannotApplyOverwriteParameterSetException` **before** mutating any parameter, and the message lists every unresolved path (bulleted for multiple, inline for one).

### New / changed
- `OverwriteParameterSetApplier` (`IOverwriteParameterSetApplier`) — new service, injected into `SimulationModelCreator`.
- `CannotApplyOverwriteParameterSetException` + `PKSimConstants.Error.CannotApplyOverwriteParameterSetUnresolvedPaths`.
- `SimulationModelCreator.CreateModelFor` calls the applier as the final construction step.

No OSPSuite.Core changes were required.

## Tests

- **Unit** (`OverwriteParameterSetApplierSpecs`): apply matching paths; `"None"` leaves params untouched; unresolved path throws + applies nothing (transactional) + message names the path; multiple compounds each apply to their own parameter/origin.
- **Integration** (`SimulationModelCreatorSpecs`): on a real constructed model, the selected set's value is applied and the parameter is re-tagged `Compound`; an unresolvable path aborts creation.

## Manual verification

All 8 acceptance criteria from #3434 were verified manually in the running app (values applied on creation; cross-simulation path matching; `BuildingBlockType = Compound` confirmed via the dev Parameter-Ids table; red indicator on subsequent edits; graceful error dialog that lists each unresolved path — bulleted for multiple, inline for one; `"None"` no-op; multiple compounds with distinct selections).

## Summary by CodeRabbit

* **New Features**
  * Simulation creation now automatically applies selected overwrite parameter sets.
  * Overwritten parameter values are applied to the correct compound parameters and marked accordingly.
* **Bug Fixes**
  * Improved validation for overwrite parameter paths—simulation creation fails with a clear error if any paths can’t be resolved, and no values are applied.
* **Tests**
  * Added integration and core specifications covering successful application, no-op when nothing is selected, and failure on unresolved paths.

### DIFF
--- a/src/PKSim.Core/Services/OverwriteParameterSetApplicationTask.cs
+++ b/src/PKSim.Core/Services/OverwriteParameterSetApplicationTask.cs
@@ -1,0 +1,85 @@
+using System.Collections.Generic;
+using System.Linq;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Core.Domain.Services;
+using OSPSuite.Utility.Extensions;
+using PKSim.Assets;
+using PKSim.Core.Model;
+
+namespace PKSim.Core.Services
+{
+   public interface IOverwriteParameterSetApplicationTask
+   {
+      /// <summary>
+      ///    Applies the values of the <see cref="OverwriteParameterSet" /> selected for each compound in the
+      ///    <paramref name="simulation" /> to the matching simulation parameters (matched by path). Affected parameters
+      ///    become <see cref="PKSimBuildingBlockType.Compound" /> so that subsequent changes follow normal compound
+      ///    parameter logic, and their value origin is set to identify the originating overwrite parameter set.
+      ///    Compounds whose selection is "None" are skipped.
+      ///    Throws a <see cref="CannotApplyOverwriteParameterSetException" /> if any path cannot be resolved in the
+      ///    simulation, in which case no value is applied.
+      /// </summary>
+      void ApplyOverwriteParameterSetsTo(Simulation simulation);
+   }
+
+   public class OverwriteParameterSetApplicationTask : IOverwriteParameterSetApplicationTask
+   {
+      private readonly IContainerTask _containerTask;
+
+      public OverwriteParameterSetApplicationTask(IContainerTask containerTask)
+      {
+         _containerTask = containerTask;
+      }
+
+      public void ApplyOverwriteParameterSetsTo(Simulation simulation)
+      {
+         var selections = simulation.OverwriteParameterSetSelections.Selections
+            .Where(x => x.OverwriteParameterSet != null)
+            .ToList();
+
+         if (!selections.Any())
+            return;
+
+         var parameterCache = _containerTask.CacheAllChildren<IParameter>(simulation.Model.Root);
+         var unresolvedPaths = new List<string>();
+         var resolvedValues = new List<(IParameter parameter, ParameterValue parameterValue, Compound compound, OverwriteParameterSet overwriteParameterSet)>();
+
+         selections.Each(selection => resolveSelection(selection, simulation, parameterCache, resolvedValues, unresolvedPaths));
+
+         if (unresolvedPaths.Any())
+            throw new CannotApplyOverwriteParameterSetException(unresolvedPaths);
+
+         resolvedValues.Each(x => applyValue(x.parameter, x.parameterValue, x.compound, x.overwriteParameterSet));
+      }
+
+      private void resolveSelection(OverwriteParameterSetSelection selection, Simulation simulation, PathCache<IParameter> parameterCache,
+         List<(IParameter parameter, ParameterValue parameterValue, Compound compound, OverwriteParameterSet overwriteParameterSet)> resolvedValues, List<string> unresolvedPaths)
+      {
+         var compound = simulation.Compounds.FindByName(selection.CompoundName);
+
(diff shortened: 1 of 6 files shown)


src/PKSim.Core/Services/OverwriteParameterSetApplicationTask.cs
```diff
@@ -0,0 +1,75 @@
...
+            throw new CannotApplyOverwriteParameterSetException(unresolvedPaths);
+
+         resolvedValues.Each(x => applyValue(x.parameter, x.parameterValue, x.compound));
+      }
+
+      private static void applyValue(IParameter parameter, ParameterValue parameterValue, Compound compound)
+      {
+         parameter.Value = parameterValue.Value.GetValueOrDefault();
+         parameter.BuildingBlockType = PKSimBuildingBlockType.Compound;
+         parameter.Origin.BuilingBlockId = compound.Id;
```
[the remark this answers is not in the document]
rwmcintosh: The compound and its overwrite parameter set are assumed to be consistent here: a selection is only ever created for a compound that is present in the simulation — `Simulation.AddOverwriteParameterSetSelection` stores a selection only when the compound name is one of the simulation's compounds. So `FindByName(selection.CompoundName)` is expected to resolve, and we intentionally don't guard it, in keeping with the convention of trusting building-block invariants rather than adding defensive null checks.

src/PKSim.Core/Services/OverwriteParameterSetApplicationTask.cs
```diff
@@ -0,0 +1,81 @@
...
+         var parameter = parameterCache[path];
+         if (parameter == null)
+            unresolvedPaths.Add(path);
+         else
+            resolvedValues.Add((parameter, parameterValue, compound));
+      }
+
+      private static void applyValue(IParameter parameter, ParameterValue parameterValue, Compound compound)
+      {
+         parameter.Value = parameterValue.Value.GetValueOrDefault();
```
Yuri05: why `GetValueOrDefault()`? I think it should be just `parameter.Value = parameterValue.Value`, no?

src/PKSim.Core/Services/OverwriteParameterSetApplicationTask.cs
```diff
@@ -0,0 +1,81 @@
...
+         parameter.BuildingBlockType = PKSimBuildingBlockType.Compound;
+         parameter.Origin.BuilingBlockId = compound.Id;
+      }
```
Yuri05: what about the **ValueOrigin** of the target parameter? It should be overwritten as well.
I think we did not specify in Detail how exactly, did we?
I would set the value origin of all simulation parameters which come from the same OPS to the same value.
Something like 
- **Source** = `Other`
- **Value**: `Overwrite Parameter Set <Name or Description of the OPS>`

src/PKSim.Core/Services/OverwriteParameterSetApplicationTask.cs
rwmcintosh: done

src/PKSim.Core/Services/OverwriteParameterSetApplicationTask.cs
rwmcintosh: done

Yuri05 (approved):

Issue #2897, opened 2024-05-02
PK- Analysis report not loading to Excel


siddhivernekar:
Application:
PK-Sim® 11.3.208

Object reference not set to an instance of an object.

Stack trace:
```
at OSPSuite.UI.Mappers.PivotGridCellsToDataTableMapper.<>c__DisplayClass14_0.<getColumnName>b__0(PivotGridField field)
   at OSPSuite.Utility.Extensions.EnumerableExtensions.Each[T](IEnumerable`1 list, Action`1 action)
   at OSPSuite.UI.Mappers.PivotGridCellsToDataTableMapper.getColumnName(PivotGridCells input, Int32 columnIndex)
   at OSPSuite.UI.Mappers.PivotGridCellsToDataTableMapper.configureDataTable(PivotGridCells input, DataTable dt, Func`2 useColumn)
   at OSPSuite.UI.Mappers.PivotGridCellsToDataTableMapper.map(PivotGridCells pivotGridCells, Func`2 displayNameRetreiver, Func`2 valueToDisplayConverter, Func`2 shouldUseRow, Func`2 shouldUseColumn)
   at OSPSuite.UI.Mappers.PivotGridCellsToDataTableMapper.MapFrom(PivotGridCells pivotGridCells, Func`2 displayNameRetreiver, Func`2 valueToDisplayConverter)
   at OSPSuite.UI.Controls.UxPivotGrid.GetCellsSummary()
   at PKSim.UI.Views.Simulations.PKAnalysisPivotView.GetSummaryData()
   at PKSim.UI.Views.Simulations.IndividualPKAnalysisView.GetSummaryData()
   at PKSim.Presentation.Presenters.Simulations.IndividualPKAnalysisPresenter.ExportToExcel()
   at OSPSuite.UI.Controls.BaseUserControl.<>c__DisplayClass23_0.<OnEvent>b__0()
   at OSPSuite.Utility.Extensions.ExceptionExtensions.DoWithinExceptionHandler(Object callerObject, Action actionToExecute)
```

PR #3595, opened 2026-07-03
Return population simulations from LoadSimulationsFromSnapshot

### LINKED ISSUE #3592: `LoadSimulationsFromSnapshot()` does not return population simulations
If a pop sim is defined in a snapshot, the `loadSimulationsFromSnapshot()` function will return a simulations object that does not include population simulations.

To be expected because populations are supposed to be passed to the `populations` argument of `runSimulations` but this should be explicitly described somewhere (or improved ?).

rwmcintosh:
Fixes #3592.

`LoadSimulationsFromSnapshot()` now returns population simulations. `SnapshotTask` carries each population simulation's `IndividualValuesCache` (base population values + simulation-level advanced parameters) and `AgingData` onto the R `Simulation`, via two mappers (`PopulationSimulationToIndividualValuesCacheMapper`, `AgingDataMapper`). Individual simulations are unchanged.

Also bumps OSPSuite.* to 13.0.131, which carries the Core wrapper properties + `RunSimulations` this depends on (#2896 / #2897).

## Summary by CodeRabbit

* **New Features**
  * Added support for exporting population simulation details, including per-individual values and aging data, when loading from snapshots.
  * Improved R data export so aging information is flattened into a format that’s easier to consume.

* **Bug Fixes**
  * Population simulations now retain complete individual-level data after snapshot import, while individual simulations remain unchanged.

### DIFF
--- a/src/PKSim.R/Services/SnapshotTask.cs
+++ b/src/PKSim.R/Services/SnapshotTask.cs
@@ -6,8 +6,11 @@ using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Mappers;
 using OSPSuite.R.Domain;
 using PKSim.Core.Services;
+using PKSim.R.Mappers;
 using CoreSnapshotTask = PKSim.Core.Snapshots.Services.ISnapshotTask;
 using CoreSimulation = PKSim.Core.Model.Simulation;
+using CoreAgingData = PKSim.Core.Model.AgingData;
+using PopulationSimulation = PKSim.Core.Model.PopulationSimulation;
 
 namespace PKSim.R.Services
 {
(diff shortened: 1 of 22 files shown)


src/PKSim.R/Services/SnapshotTask.cs
```diff
@@ -61,7 +70,22 @@ public Simulation[] LoadSimulationsFromSnapshot(string snapshotFile, params stri
...
+      {
+         var rSimulation = new Simulation(coreSimulationFrom(simulation));
+
+         if (simulation is PopulationSimulation populationSimulation)
+         {
+            rSimulation.IndividualValuesCache = _populationCacheMapper.MapFrom(populationSimulation);
+
+            if (shouldExportAgingData(populationSimulation.AgingData))
+               rSimulation.AgingData = _agingDataMapper.MapFrom(populationSimulation.AgingData);
+         }
```
rwmcintosh, line 86: We just want to put the aging data and values cache in the R simulation domain object. This ends up indicating it's a population simulation.

Yuri05 (approved):

Issue #2902, opened 2024-05-19
Input parameters


siddhivernekar:
How can all the input parameters for the Model be exported from the PK-Sim software?

PR #3596, opened 2026-07-04
Reduce PK-Sim startup time

### LINKED ISSUE #1472: Investigate PK-Sim start time... as it is too slow
Implementing the PKSim-R API made me realize that we have some serious stuff going on during PKSim startup that may not be required
In fact, loading of the CORE of PKSim is done in R when one called the `initPKSim` method.
It takes about 3sec. then The first individual takes about 3 sec, After that, under 1sec

In PKSim, it takes about 20 sec to start. I'd like to investigate where we spend so much time. Ideally, <10 sec would be ideal

rwmcintosh:
Fixes #1472

Reduces PK-Sim startup time. Four independent changes (~5 s combined, cold Debug):

| Fix | In this PR | Saving |
|---|---|---|
| `ValueOriginRepository.Start` guard (stop re-mapping value origins on every lookup) | yes | ~1.0 s |
| Background (parallel) repository warm-up overlapping window construction | yes | ~1.0 s |
| Journal page editor built on first use instead of at startup | yes | ~1.9 s |
| Skin gallery populated on demand (via OSPSuite.Core 13.0.132, #2902) | package bump | ~1.4 s |

Draft — opened alongside the MoBi counterpart (Open-Systems-Pharmacology/MoBi#2430) and depends on the published OSPSuite.Core 13.0.132.

## Summary by CodeRabbit

* **New Features**
  * Application startup now warms repositories concurrently in the background after settings are loaded, improving initial responsiveness.
  * Launch now waits for the warm-up to complete before the first database access.
  * Journal editor UI is created on first use and activated via the edit-start event.
* **Bug Fixes**
  * Prevents repeated initialization of the value-origin cache.
  * Journal editor menu command is now created lazily.
* **Tests**
  * Added/updated coverage for journal editor activation and warm-up/presenter startup behavior.
* **Chores**
  * Updated OSPSuite dependencies to version 13.0.132.

### DIFF
--- a/src/PKSim.Infrastructure/ORM/Repositories/ValueOriginRepository.cs
+++ b/src/PKSim.Infrastructure/ORM/Repositories/ValueOriginRepository.cs
@@ -19,6 +19,7 @@ namespace PKSim.Infrastructure.ORM.Repositories
       };
 
       private readonly Cache<int, ValueOrigin> _allValueOrigins = new Cache<int, ValueOrigin>(onMissingKey: x => _defaultValueOrigin);
+      private bool _started;
 
       public ValueOriginRepository(IFlatValueOriginRepository flatValueOriginRepository, IFlatValueOriginToValueOriginMapper valueOriginMapper)
       {
@@ -34,7 +35,9 @@ namespace PKSim.Infrastructure.ORM.Repositories
 
       public void Start()
       {
+         if (_started) return;
          _flatValueOriginRepository.All().Each(x => { _allValueOrigins[x.Id] = _valueOriginMapper.MapFrom(x); });
+         _started = true;
       }
 
       public ValueOrigin ValueOriginFor(IParameter parameter)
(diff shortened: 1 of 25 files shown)


rwmcintosh:
@coderabbitai full review please

src/PKSim.Infrastructure/ORM/Repositories/ValueOriginRepository.cs
```diff
@@ -34,7 +35,9 @@ public IEnumerable<ValueOrigin> All()
 
       public void Start()
       {
+         if (_started) return;
          _flatValueOriginRepository.All().Each(x => { _allValueOrigins[x.Id] = _valueOriginMapper.MapFrom(x); });
+         _started = true;
       }
```
Yuri05: any reason why **ValueOriginRepository**  is not inherited from the **StartableRepository**?
Then Start would become `DoStart `(oder whatever that one is called) and we don't need to deal with `_started `etc. here

rwmcintosh: No, it's not mentioned in the issue or PR where it was introduced. I'll try it.

rwmcintosh: done

Yuri05 (approved):

PR #3604, opened 2026-07-14
Allow re-use of an administration protocol for multiple compounds

### LINKED ISSUE #3603: Allow re-use of administration for more than one compound
It should be possible to use an administration for multiple compounds.

rwmcintosh:
Allows the same administration protocol to be selected for more than one compound in a simulation.

Previously this was blocked in the simulation wizard (`SimulationCompoundProtocolCollectorPresenter`) with the warning *"Each administered compound must have a unique administration protocol."* The restriction existed because each compound's applications are built into an event group named after the protocol — so a shared protocol produced two event groups with the same name (colliding container/model paths).

Original naming remains when not shared
image

Shared
image

Fixes #3603

## Summary by CodeRabbit

- **New Features**
  - Simulations can now use the same administration protocol for multiple compounds.
  - Event groups are automatically named to remain unique and clearly associated with each compound.

- **Bug Fixes**
  - Removed duplicate-protocol validation warnings during compound protocol selection.
  - Corrected event-group naming when protocol and compound combinations could otherwise collide.

- **Tests**
  - Expanded coverage for shared protocols, naming conflicts, and duplicate selections.

### DIFF
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -446,7 +446,6 @@ namespace PKSim.Assets
          public const string AtLeastOneCompoundMustBeSelected = "At least one compound must be selected.";
          public const string AtLeastOneFileRequiredToStartPopulationImport = "At least one file is required to perform the population import.";
          public const string AtLeastOneProtocolRequiredToCreateSimulation = "Select at least one protocol for the administered compound(s).";
-         public const string AProtocolCanOnlyBeUsedOnceInASimulation = "Each administered compound must have a unique administration protocol.";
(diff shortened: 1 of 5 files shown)


Yuri05:
@rwmcintosh In the example above: what If I create another protocol named **Oral300mg-Theophylline** and use it at the same time? 😉

rwmcintosh:
It would have been nicer if I used A/B/C/D but I didn't.

The sequence now is after applying the first pass of composite names, we go look for the simple names that now collide, and we add the compound.

That only moves the problem, to another level, but does not guarantee uniqueness. So after that, we search the names again and apply an index.

So that said, the last event in the list does not actually contain the administered compound name. It's a protocol named `oral 300mg-Theophylline-Theophylline3` that coincides with with the composite name directly above it.



image

Yuri05 (approved):

PR #3616, opened 2026-07-22
Fixes #3614 Offset to the administration for simple protocol events

### LINKED ISSUE #3614: Offset to the administration
>Simple Protocol remains unchanged for basic use cases, with the Events tab available for users who want simple protocols + basic events.
> 
> I think for the simple protocol we should allow **max. 1 event**, and then there is **no need for a separate Events Tab**.
> This will cover the most common use case (_administration before/after/together with meals_).
> The event setup (whether add an event or not , offset to the administration (can be negative here!) can be done directly, smthg like below:
> 
> Image
> 
> 
> -----------
> 
> But then we are not compatible with older simulations. I think we need the Event tabs or how do we deal with other projects?  Or is there confusion with the Event tabs maybe. It's supposed to be in the simulation only 

 _Originally posted by @Yuri05 in [#3424](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3424#issuecomment-4045389457)_

msevestre:
Implements the design from #3614 for the **Simple Protocol** view.

### What changed
- **Removed the Placeholder combo** from the simple protocol view. The event key is no longer user-selected — it defaults to `DEFAULT_EVENT_KEY` under the hood (still resolved to a real event via the simulation's event placeholder mapping). This covers the common *administration before / after / together with a meal* case directly, without a separate step.
- **Relabelled the checkbox** to **"Add an event for every administration"** (new `AddEventForEveryAdministration` asset constant).
- **One event per administration**: `SimpleProtocolToSchemaMapper` now adds an event schema item next to *each* administration item, at `administrationTime + offset`, instead of a single event per schema cycle. This makes the behaviour honest for multi-dose intervals (6-6-12, 6-6-6-6, 8-8-8) as well as single/bi-daily/daily.
- **Negative offsets** flow through unchanged (the `Event offset` parameter already allows negatives: `MinValue = null; MinIsAllowed = true`), so a meal can be placed before each administration.

### Layout
The `Placeholder` row was removed from the table panel; the `Event offset` row moved up accordingly.

### Tests
- Updated the event-offset schema-mapper spec to assert `event start = administration time + offset`.
- Added a spec covering multiple administrations (6-6-12): one event per administration, each offset from its own dose.
- Removed the obsolete `SetSimpleProtocolEventKey` presenter spec.
- All simple-protocol / schema-mapper specs pass (55). The only failing tests in the suite (`application_should_have_at_least_one_event`, `schema_item_should_contain_dose_parameter`, `parameters_should_have_the_accurate_building_block_type`) are pre-existing integration-DB failures on the base branch, unrelated to this change.

Fixes #3614

## Summary by CodeRabbit

* **New Features**
  * Events can now be added for every administration in a simple protocol, creating one event per administration and applying the configured event offset.
* **UI Improvements**
  * Simplified simple-protocol event settings to focus on event offset; event-key selection was removed and the caption updated accordingly.
* **Bug Fixes**
  * Fixed event generation for protocols with multiple administrations.
  * Improved protocol chart display to show events occurring before the first administration.
* **Other**
  * Event start times now support negative values.

### DIFF
--- a/src/PKSim.Core/CoreConstants.cs
+++ b/src/PKSim.Core/CoreConstants.cs
@@ -34,6 +34,8 @@ namespace PKSim.Core
       public const string DEFAULT_TEMPLATE_VERSION = "1.0";
       public const string DEFAULT_FORMULATION_KEY = "Formulation";
       public const string DEFAULT_EVENT_KEY = "EVENT_1";
+      //the simple protocol allows a single event, so it uses an unnumbered placeholder key
+      public const string SIMPLE_PROTOCOL_EVENT_KEY = "EVENT";
       public const string DEFAULT_CALCULATION_METHODS_FILE_NAME_FOR_MOBI = "AllCalculationMethods";
       public const string DEFAULT_EXPRESSION_PROFILE_MOLECULE_NAME = "<MOLECULE>";
       public const TransportType DEFAULT_TRANSPORTER_TYPE = TransportType.Efflux;
--- a/src/PKSim.Core/CoreConstants.cs
+++ b/src/PKSim.Core/CoreConstants.cs
@@ -35,7 +35,7 @@ namespace PKSim.Core
       public const string DEFAULT_FORMULATION_KEY = "Formulation";
       public const string DEFAULT_EVENT_KEY = "EVENT_1";
       //the simple protocol allows a single event, so it uses an unnumbered placeholder key
-      public const string SIMPLE_PROTOCOL_EVENT_KEY = "EVENT";
+      public const string SIMPLE_PROTOCOL_EVENT_KEY = "Event";
       public const string DEFAULT_CALCULATION_METHODS_FILE_NAME_FOR_MOBI = "AllCalculationMethods";
       public const string DEFAULT_EXPRESSION_PROFILE_MOLECULE_NAME = "<MOLECULE>";
       public const TransportType DEFAULT_TRANSPORTER_TYPE = TransportType.Efflux;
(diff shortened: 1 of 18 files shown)


src/PKSim.Core/CoreConstants.cs
```diff
@@ -34,6 +34,8 @@ public static class CoreConstants
       public const string DEFAULT_TEMPLATE_VERSION = "1.0";
       public const string DEFAULT_FORMULATION_KEY = "Formulation";
       public const string DEFAULT_EVENT_KEY = "EVENT_1";
+      //the simple protocol allows a single event, so it uses an unnumbered placeholder key
+      public const string SIMPLE_PROTOCOL_EVENT_KEY = "EVENT";
```
msevestre: @Yuri05 I am not sure about this. This is the name of the event for only ONE item. (e.g. Simple Protocol). Instead of EVENT_1, we see EVENT

Image

msevestre:
image

Ensure that the start time can also be negative in simulation

msevestre:
image

@yuri05 no idea if this works but the time starts in the negaative lol

msevestre:
image
with a 0 offset instead of -1. def does sthg

msevestre:
what's the purpose again of negative start time for event ? @yuri05 ?

src/PKSim.Core/CoreConstants.cs
Yuri05: ~As the admin protocols in the most (all?) cases will be combined with **meal** events: should we rename the visible key to smthg like "**Event (typically meal)**" or "**(Meal) event**" or smthg similar?~

Yuri05:
> image with a 0 offset instead of -1. def does sthg

which is kinda surprising, because it's still using the old SimModel (just created a PR in Core to update simmodel), so that events with negative start times are filtered out, hmm

Yuri05:
> what's the purpose again of negative start time for event ? @Yuri05 ?

To realize the use case "Administrations **after** meals" without shifting the start time of administration protocols if a meal event was added (so that the first administration in a simple protocol is always at t=0)

src/PKSim.Core/CoreConstants.cs
Yuri05: After discussion: should be named "Event" (and the dialog label should be extended by _"(e.g. meal)"_

PR #3623, opened 2026-07-22
Fixes #3620 Refresh the protocol chart when the protocol mode changes

### LINKED ISSUE #3620: Administration Protocol graph does not update when navigating from advanced to simple

msevestre:
Fixes #3620

## Problem
Switching the protocol mode from **Advanced** to **Simple** left the previous (advanced) graph on screen.

## Cause
`CreateProtocolPresenter.ProtocolMode` creates a brand new protocol and swaps it in via `updateView()`, but never re-plots the chart. Simple → Advanced *appeared* to work only by accident: binding the advanced schema grid raises `StatusChanged`, which triggers `refreshPlot()`. The simple protocol view raises nothing, so Advanced → Simple kept showing the stale chart.

## Fix
`updateView()` now calls `refreshPlot()` after the protocol has been swapped, so both directions plot the newly created protocol.

## Tests
Two new specs in `CreateProtocolPresenterSpecs` cover both switch directions; both fail without the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the protocol chart immediately when switching between protocol modes.
  * Ensured the chart displays the newly selected Simple or Advanced protocol without requiring another refresh.

### DIFF
--- a/src/PKSim.Presentation/Presenters/Protocols/CreateProtocolPresenter.cs
+++ b/src/PKSim.Presentation/Presenters/Protocols/CreateProtocolPresenter.cs
@@ -117,6 +117,8 @@ namespace PKSim.Presentation.Presenters.Protocols
          activeProtocolPresenter.EditProtocol(Protocol);
          _view.UpdateEditControl(activeProtocolView);
          updateControls();
+         //the protocol was swapped: the chart still shows the previous protocol and has to be plotted again
+         refreshPlot();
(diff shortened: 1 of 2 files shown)


msevestre:
The bug we saw this morning with the refresh

Yuri05 (approved):

rwmcintosh (approved):

Issue #3459, opened 2026-03-25
Event entry type in protocol domain model


msevestre:
## Parent PRD

#3424

## What to build

Extend the protocol domain model to support event entries alongside administration entries. This is the foundational model layer that both Advanced Protocol and Simple Protocol will build on.

Specifically:
- Extend `SchemaItem` (or create a parallel concept) to support an "Event" entry type
- Event schema entries have: **Start Time** and **Event Placeholder** (string identifier, e.g., `EVENT_1`)
- Event entries do **NOT** have: Dose, Target Organ, Target Compartment
- Events inherit Schema-level repetition settings (Number of Repetitions, Time Between Repetitions)
- The same event placeholder can be used multiple times in a schema (e.g., `EVENT_1` at 0h and `EVENT_1` at 8h)

See PRD sections: "Protocol Model Changes" and "Type Column" for full specification.

## Acceptance criteria

- [ ] `SchemaItem` supports an Event entry type distinct from administration types
- [ ] Event entries store a Start Time and Event Placeholder string
- [ ] Event entries do not carry Dose, Target Organ, or Target Compartment
- [ ] Schema expansion correctly generates event occurrences respecting repetition settings
- [ ] The same event placeholder can appear multiple times in a schema
- [ ] Serialization round-trip works for protocols containing event entries
- [ ] Unit tests cover model creation, validation, schema expansion with events, and serialization

## Blocked by

None — can start immediately.

## User stories addressed

- User story 1: Add event entry to advanced protocol schema
- User story 4: Drug administration at fixed offset from event within same repetition
- User story 6: Event placeholder inherits schema repetition settings
- User story 10: Same event placeholder used multiple times in a schema

msevestre:
## Agent Brief

**Category:** Task (enhancement)
**Summary:** Extend the protocol domain model so `SchemaItem` can represent an "Event" entry type alongside existing administration types.

**Current behavior:**
[...]

Issue #3424, opened 2026-03-04
Event Handling Redesign: Integrate Events with Advanced Protocols


msevestre:
# Event Handling Redesign PRD

## Problem Statement

Currently in PK-Sim, administration protocols and events are separate building blocks with different mechanisms:
- **Protocols** define drug administrations with flexible timing (repetitions, intervals)
- **Events** (meals, exercise, etc.) must be defined individually in the simulation's Events tab, each with its own start time

This separation creates pain points:
1. Users must manually define the same event multiple times for repeated scenarios (e.g., meals at breakfast, lunch, dinner for 7 days = 21 manual entries)
2. Events cannot leverage the powerful repetition system that protocols have
3. Real-world scenarios like "take medication 30 minutes after each meal" require manual calculation of all timing
4. No way to create reusable event schedules that can be swapped across simulations

## Solution

Integrate events into the Advanced Protocol system using a placeholder mechanism (similar to existing formulation placeholders). This allows:
- Events defined as entries within protocol schemas, inheriting repetition behavior
- Event placeholders mapped to actual event building blocks at simulation creation
- Unified timeline visualization of both administrations and events
- Reusable protocol templates with swappable event instances

**Simple Protocol** is extended to support a single optional event with an offset relative to the administration. The Events tab remains available for all simulation types.

## User Stories

1. As a modeler, I want to add an event entry to my advanced protocol schema, so that events follow the same repetition pattern as my administrations.

2. As a modeler, I want to define an event placeholder (e.g., EVENT_1) in my protocol, so that I can map it to different actual events in different simulations.

3. As a modeler, I want to see events and administrations on the same timeline visualization, so that I can verify the timing relationships are correct.

4. As a modeler, I want to define a drug administration at a fixed offset from an event (e.g., drug at 2h, event at 0h within same repetition), so that I can model real-world scenarios like "medication after meals."

5. As a modeler, I want to reuse the same protocol with different event mappings, so that I can compare "heavy breakfast" vs "light breakfast" scenarios without recreating the protocol.

6. As a modeler, I want my event placeholder to inherit the schema's repetition settings, so that defining "3 meals per day for 7 days" only requires one schema entry with appropriate repetitions.

7. As a modeler, I want to add a single event with an offset to my simple protocol, so that I can model common scenarios like "medication after a meal" without switching to advanced mode.

[...]

PR #3624, opened 2026-07-23
Always create a formulation node in the protocol hierarchy (Fixes #3462)

### LINKED ISSUE #3462: Simulation build with event group nodes
## Parent PRD

#3424

## What to build

During simulation construction, create event group nodes from mapped event placeholders. This is the core engine change that turns protocol event entries + placeholder mappings into actual simulation structure.

Specifically:
- Event group nodes created **once** per unique mapped event in the simulation structure
- The event group node contains all event parameters (from the mapped PKSimEvent building block)
- Child event group nodes contain only start times (computed from schema expansion + repetitions)
- Multiple occurrences of the same event placeholder reference the single event group node
- Event parameters are editable in one place only (no duplication)

Additionally, standardize formulation nodes:
- Always include a formulation node in the protocol hierarchy, even for administration types that don't require one (e.g., IV Bolus)
- For IV Bolus and similar, the formulation node is named **"No formulation"**
- This ensures consistent data structure depth

Includes serialization/snapshot support for the new simulation structure.

See PRD sections: "Simulation Structure (Node Hierarchy)" and "Formulation Node Standardization" for full specification.

## Acceptance criteria

- [ ] Event group nodes are created from mapped event placeholders during simulation build
- [ ] Event parameters come from the mapped PKSimEvent building block (centralized)
- [ ] Child event group nodes contain only start times
- [ ] Single event group per unique mapped event (no duplication across occurrences)
- [ ] IV Bolus and similar administration types get a "No formulation" formulation node
- [ ] Serialization/snapshot round-trip works for simulations with event group nodes
- [ ] MoBi users can see event group nodes in simulation structure with parameters in a single location
- [ ] Unit tests cover event node creation, deduplication, "No formulation" node, and serialization

## Blocked by

- Blocked by #3461 (Event placeholder mapping in simulation configuration)

## User stories addressed

- User story 11: Event group nodes in simulation structure with parameters defined once per unique event
- User story 15: Event-specific parameters come from the mapped PKSimEvent building block

msevestre:
## Summary
- Applications that do not require a formulation (IV Bolus, IV Infusion) were added directly under the protocol event group, while applications with a formulation were nested one level deeper
- Every application now sits under a formulation-level container: non-formulation applications go under a single `No formulation` container, created once per protocol event group and shared by all of them
- Formulation container lookup/creation is factored into one helper used by both branches

Application containers for non-formulation routes therefore move from `<Protocol>/Application_N` to `<Protocol>/No formulation/Application_N`, giving the protocol hierarchy a consistent depth (as specified in the PRD #3424 and requested for MoBi structure visibility).

## Test plan
- [x] New specs: IV-only protocol (single shared `No formulation` container, no application directly under the protocol event group, empty formulation type and no formulation parameters passed to the application factory)
- [x] New specs: mixed IV + oral protocol (one `No formulation` container plus one container per mapped formulation, each holding the right applications)
- [x] All 31 `EventBuildingBlockCreator` specs pass
- [ ] Full suite: 5536 passed / 5 failed / 6 skipped — the 5 failures are being investigated and will be addressed in this PR

## Remaining on #3462
- MoBi structure visibility verification (manual)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of protocol applications with and without formulations.
  * Applications without a formulation are now grouped under a dedicated “No formulation” container.
  * Applications requiring formulations are grouped under the appropriate formulation container.
  * Event items no longer incorrectly require dose parameters.

* **Tests**
  * Added coverage for mixed formulation scenarios.
  * Updated integration tests to distinguish administration and event applications.


msevestre:
## Triage of the 5 failing tests

**1 — real work still owed on #3462:**
`When_loading_a_project_saved_before_applications_were_nested_under_a_formulation_container` (spec added on this branch in 8c5d7206) fails with `Expected: "No formulation" But was: "IV"`. A loaded project keeps its persisted model structure — the simulation tree is deserialized, not rebuilt — so applications in previously saved projects still sit directly under the protocol event group. A project conversion is needed to insert the `No formulation` container on load.

**4 — inherited from #3459, unrelated to this change:**
- `schema_item_should_contain_dose_parameter`
- `parameters_should_have_the_accurate_building_block_type` (NullReferenceException on the missing input dose parameter)
- `application_should_have_at_least_one_event`
- `every_application_except_oral_and_bolus_should_have_application_rate_parameter`

These iterate over *every* schema item / application in the PK-Sim database and assume each one is a drug administration. That stopped being true when the **Event** application type was added to the database in 442591cf (Fixes #3459). An Event entry has no dose, no input dose and no application rate. They never build a simulation, so container nesting cannot affect them — they need to exclude the Event application type, which belongs to the parent feature work rather than to this PR.

msevestre:
## Project conversion added — full suite is green

`Converter12To13` now converts saved projects: applications of a protocol without formulation are moved under a `No formulation` container, and every object path going through them is updated (formulas, right hand side formulas and event assignments), since those applications end up one level deeper in the model.

Two findings worth recording:

**The conversion spec was pointed at a project that is already at v13.** `Mucosa_MultipleTransportDirections.pksim5` is saved at version 81 (= `ProjectVersions.V13`, the current version), so `Converter12To13` never ran on it and the spec could not pass regardless of the converter. It now loads `SimpleSimulationIV_710.pksim5` (version 71, protocol `IV`), which goes through the whole converter chain. Consequence to be aware of: projects saved by 13.x dev builds keep the old structure and are never converted.

**Moving the container breaks absolute formula paths.** Before the path update was added, loading failed with:

```
Unable to resolve path 'Human|Applications|IV|Application_1|ProtocolSchemaItem|DrugMass'
for entity 'Total drug mass' with path 'Human|Aciclovir|Total drug mass'
```

The converter therefore walks the model as `IVisitor<IUsingFormula>` / `IVisitor<IParameter>` / `IVisitor<EventAssignment>` — the same pattern as `RenameAbsolutePathVisitor` — and inserts `No formulation` into every path that goes through a moved application. The insertion is idempotent. Note also that in older projects the protocol lives under an `Applications` root container rather than under `Events`, so the converter looks for application containers structurally instead of by a fixed path.

The remaining 4 database-spec failures reported earlier were fixed in the previous commit.

**Full suite: 5543 passed, 0 failed, 6 skipped.**

src/PKSim.Core/Services/EventBuildingBlockCreator.cs
```diff
@@ -291,19 +291,15 @@ private void addApplication(EventGroupBuilder protocolGroupBuilder, ISchemaItem
...
+            applicationParentContainer = formulationContainerFor(protocolGroupBuilder, formulation.Name, () => createFormulationAsEventGroupBuilderFrom(formulation));
             formulationType = formulation.FormulationType;
             formulationParameters = applicationParentContainer.GetChildren<IParameter>();
          }
          else
          {
-            applicationParentContainer = protocolGroupBuilder;
+            //applications that do not require a formulation are nested under a dedicated formulation container so that the
+            //protocol hierarchy always has the same depth
+            applicationParentContainer = formulationContainerFor(protocolGroupBuilder, CoreConstants.ContainerName.NoFormulation, createNoFormulationEventGroupBuilder);
```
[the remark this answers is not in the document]
msevestre: this is done

msevestre:
some fialing tests that were not cought thus far plus adding the no formulation

@Yuri05 @rwmcintosh I am converting simulation and add the missing node here. Does it make sense? or should we keep the formulation inconsistant?

tests/PKSim.Tests/IntegrationTests/RenameBuildingBlockTaskSpecs.cs
```diff
@@ -41,12 +38,11 @@ public class When_renaming_the_application_in_the_project_rename_application_v8
 
       public override void GlobalContext()
       {
-         var converter = new Converter12To13(new CoreConverter121To130(A.Fake<OSPSuite.Core.Domain.ISolverSettingsFactory>()));
```
msevestre, line 44: oh my. fake in integrations test. WHY WHY WHY

src/PKSim.Infrastructure/ProjectConverter/v13/Converter12To13.cs
```diff
@@ -1,31 +1,188 @@
...
+      ///    An IV protocol is converted (the oral protocol below is already nested and stays as it is):
+      ///    <code>
+      ///    Sim|Applications|IV                          Sim|Applications|IV
+      ///                     └─ Application_1     =>                      └─ No formulation
+      ///                                                                     └─ Application_1
+      ///
+      ///    Sim|Applications|Oral                        Sim|Applications|Oral
+      ///                     └─ Tablet            =>                      └─ Tablet
+      ///                        └─ Application_1                             └─ Application_1
+      ///    </code>
```
Yuri05: nice comment :)

Yuri05 (approved):

Yuri05:
> I am converting simulation and add the missing node here. Does it make sense? or should we keep the formulation inconsistant?

Probably OK, but might require snapshot conversion?
[V12_ModifiedInfusionTime.zip](https://github.com/user-attachments/files/30300450/V12_ModifiedInfusionTime.zip)
E.g. can you load from the attached snapshot and run the PI without any errors/warnings?

msevestre:
Snapshot conversion. DAMN. 
Based on your email, should we leave the simulation intact "as is". I am tempting to say yes

Yuri05:
> should we leave the simulation intact "as is". I am tempting to say yes

no strong preference from my side, so yeah. let's leave AS IS

msevestre:
all right removed

PR #3399, opened 2026-02-12
PBBM Workflow: transfer of the status quo

Yuri05:
Fixes #3353

# Description

This is the main part of https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2448 
(the implementation of https://doi.org/10.1016/j.ejps.2025.107164 resp. https://doi.org/10.1016/j.ejps.2025.107189)

Once this PR is integrated, we can rerun all the qualification plans (including the [new one for PBBM](https://github.com/Open-Systems-Pharmacology/Qualification-Oral-PBBM)).
_Und dann beginnt die fröhliche Stunde_ (for those who know 😉 )

Remaining TODOs (not for this PR - separate issues created):
* Project conversion (https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3402)
* Adding some new parameters to the list of boolean parameters (for displaying YES/NO instead of 0/1) (https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3403)
  grafik
* Changing the grouping of some compound parameters, e.g. moving this one to the Lipophilicity parameter alternative (https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3404)
  grafik
* Maybe changing the grouping/visibility of some new parameters (s. screenshots below) (https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3405)
* Maybe also more tests. (https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3406)
* Add/Update value origins (refer to https://doi.org/10.1016/j.ejps.2025.107164 or https://doi.org/10.1016/j.ejps.2025.107189) (https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3407)

## Type of change

Please mark relevant options with an `x` in the brackets.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires documentation changes (link at least one [user](https://github.com/Open-Systems-Pharmacology/docs) or [developer](https://github.com/Open-Systems-Pharmacology/developer-docs) documentation issue):
- [X] Algorithm update - updates algorithm documentation/questions/answers etc.
- [ ] Other (please describe):

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [X] Integration tests
- [ ] Unit tests
[...]

PR #3626, opened 2026-07-23
Fixes #3619 Restore project conversion coverage in integration tests

### LINKED ISSUE #3619: ProcessToProcessBuilderMapperSpecs should fail but they don't
due to conversion issue. 

we are expecting this 
Image

msevestre:
Closes #3619

## Problem

`ProcessToProcessBuilderMapperSpecs` passed while PK-Sim itself failed to load
the very same projects (see the screenshot in #3619).

Root cause is the test data, not the converter code. Both projects used by
those specs were re-saved as **V13** files in 44038308 (#3399, commit message:
*"Fix failing tests"*):

| File | before #3399 | after |
|---|---|---|
| `Mucosa_MultipleTransportDirections.pksim5` | version **78** (V10) | version **81** (V13) |
| `MucosaInflux_Hill.pksim5` | version **78** (V10) | version **81** (V13) |

Every `CONTENTS` blob carries `xmlVersion="81"`, and
`XmlSerializationManager.convertXml` only runs converters while
`version != ProjectVersions.Current` — so **zero converters ran** and the specs
were asserting against an already-migrated project.

## Changes

- Restore both projects to their pre-#3399 V10 versions so the conversion chain
  is actually exercised.
- Add `ProjectConverter/v13/Converter12To13Specs.cs` — a minimal statement of the
  contract: load `SimpleProject_730` (v7.3, untouched by #3399), take its
  converted `Ind` / `Caffeine` / `IV` building blocks, and create a simulation.
  No mucosa/transporter specifics in the way.

## Expected result: these tests FAIL

```
Failed:  3, Passed: 25, Total: 28
```

All three fail with the same error as the app:

```
CannotCreateSimulationException : Parameter 'Fraction of geometric volume filled with liquid'
  with path 'simulation|Organism|Lumen|Stomach|Fraction of geometric volume filled with liquid'
  references an entity with path '..|Volume' that cannot be found
```

thrown at `SimulationModelCreator.cs:62`. The converted individual's
`Lumen|<segment>` containers are missing the new PBBM parameters (`Volume`,
`Bile Salt concentration`, `pH`, `Fraction of geometric volume filled with
liquid`, `Absorption/Secretion of liquid`). `Converter12To13` only delegates to
the Core converter and adds nothing on the PK-Sim side.

**The conversion itself is a deliberate follow-up** — this PR only restores the
coverage that makes the gap visible.

## Summary by CodeRabbit

* **Tests**
  * Added coverage confirming that simulations can be created successfully from converted project building blocks.
  * Verified that the resulting simulation includes a valid model.


msevestre:
@Yuri05 Restored the old project and also added a simple file. We should have 3 tests failing now

Yuri05 (approved):

PR #3646, opened 2026-07-29
Fixes #3397 Open exported files again

### LINKED ISSUE #3397: V13: Exported files do not open automatically
E.g. in the 2 cases below, exported files were always opened in Excel (resp. Text editor).
In V13, the files are exported but not automatically opened.

(Maybe we have more cases, didn't test others)
Image

rwmcintosh:
Fixes #3397

`OSPSuite.Core` family 13.0.140 -> 13.0.141 and `OSPSuite.Utility` 5.0.0.10 -> 5.0.0.11. No PK-Sim code change is needed.

Root cause was in `FileHelper.TryOpenFile` (Open-Systems-Pharmacology/OSPSuite.Utility#58): since .NET Core, `Process.Start(path)` no longer shell-executes, so it threw `Win32Exception` and the surrounding `catch` swallowed it. Both cases from the report go through that one method — *Export Simulation Structure to File* via `SimulationExportTask.CreateSimulationReport`, and *Export Results to Excel* via `ExportToExcelTask` — as do journal export, chart and observed data export to Excel, and the help links in `BaseShell`. Core 13.0.141 additionally fixes the message box hyperlinks (Open-Systems-Pharmacology/OSPSuite.Core#2915).

`OSPSuite.Utility` has to move too, otherwise PK-Sim's direct reference would downgrade what Core 13.0.141 pulls in.

MoBi had the same regression, fixed in Open-Systems-Pharmacology/MoBi#2462.

### DIFF
--- a/src/PKSim.Assets/PKSim.Assets.csproj
+++ b/src/PKSim.Assets/PKSim.Assets.csproj
@@ -22,9 +22,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.140" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.140" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.140" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.141" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.141" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.141" />
   </ItemGroup>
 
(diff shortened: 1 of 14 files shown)


Yuri05:
@rwmcintosh 
Here we don't need to change this somewhere like in MoBi?
image

rwmcintosh:
> @rwmcintosh Here we don't need to change this somewhere like in MoBi? image

No because there isn't a separate `Process.Start` that I can see. PKSim uses the FileHelper for excel and xml and doesn't have a URL open mechanism like we used for module merge behavior

Yuri05 (approved):

PR #3650, opened 2026-08-07
Fixes #3404 Add an Advanced solubility node for the bile salt micelle parameters

### LINKED ISSUE #3404: Change the grouping for some new compound parameters
Changing the grouping of some compound parameters, e.g. moving this one to the Lipophilicity parameter alternative

Image

msevestre:
Closes #3404

## What changed

The five bile salt micelle parameters no longer sit in the particle dissolution group. They render under a new **Advanced solubility** node in the compound's ADME tree, as a sibling of Specific intestinal permeability, split across two parameter groups:

- **`COMPOUND_ADVANCED_SOLUBILITY`** — critical micellar concentration, the two partition coefficient constants and the ionized coefficient, as ordinary editable compound parameters in a plain table.
- **`COMPOUND_BILE_SALT_PARTITION_COEFFICIENT`** — the neutral coefficient alone, registered as a group with a *calculated* default alternative. Its "Calculated" alternative keeps the database formula and expands to show the value `K_n` takes for every Lipophilicity alternative; the user can add, rename, default and delete constant alternatives exactly as for Permeability.

The split is required for correctness, not tidiness: a group with a calculated default blanks **every** parameter of a newly added alternative to `NaN`, and parameters of an alternative group are sourced only from the selected alternative when the molecule is built. Sharing one group would push `NaN` for the CMC and the ionized coefficient into the simulation. `K_i` stays out of the group as well, so its formula reads `K_n` from the built molecule after the selected alternative has been applied — it follows automatically with no user action.

Also renames the `Crititical Micellar concentration` typo.

## Database

Applied as SQL against `PKSimDB.sqlite` with foreign keys on, then the dump and diff were regenerated with the rake tasks. The rename cascaded to `tab_container_parameters`, transitively to `tab_container_parameter_rates`, and directly to `tab_rate_generic_parameters`; `PRAGMA foreign_key_check` is clean and no `Crititical` spelling remains anywhere. The dump diff is 10 lines and `PKSimDB_diff.txt` shows exactly the described changes and nothing else.

## Two converter changes that were not in the issue's scope

The issue put converter work out of scope on the grounds that these parameters have never shipped. That reasoning holds for the parameter *rename*, but not for the new alternative-group *container*: registering a sixth alternative group means no previously saved project has one. The existing suite caught it — **43 tests failed**, every one on project load.

- `Converter721To730` looked the new template group up on the loaded compound and dereferenced `null`. Now skips groups the project does not have, mirroring what `Converter12To13` already did.
- `Converter12To13` null-guarded but then left the compound without the group, so the new node would have thrown for any pre-v13 project. Now clones a wholly missing group from the template, bringing its "Calculated" default with it.

No parameter migration and no new converter class. Happy to strip these and file a follow-up if you would rather keep the PR strictly to scope.

## Reuse note

`PermeabilityGroupPresenterBase`, `PermeabilityAlternativeDTO` and `PermeabilityGroupView` are reused as-is rather than renamed to something parameter-neutral. The only generalisation is that the grid's value-column caption and the calculated-value description now come from the presenter instead of being hardcoded to Permeability. Renaming the whole family would have been a much larger diff for no behavioural gain; say the word if you want it.

## Tests

Full suite green: **5620 passed, 0 failed, 6 skipped**.

| Layer | Covers |
|---|---|
| Domain / integration (`BileSaltPartitionCoefficientSpecs`) | calculated default alternative; the four constants outside any alternative; user alternative holding only `K_n` as an editable constant; molecule built with either alternative, with `K_i` deriving from the user's `K_n` and nothing `NaN`; snapshot round-trip of alternatives, default flag and edited constants |
| Serialization (`CompoundXmlSerializerSpecs`) | XML round-trip of several alternatives including which is default |
| Snapshot mapper (`CompoundMapperSpecs`) | new `BileSaltPartitionCoefficient` property mapped |
| Core service (`CompoundAlternativeTaskSpecs`) | one read-only parameter per Lipophilicity alternative, named after it, evaluating `c1·LogP + c2` |
| Presentation | grid binding, calculated-value hook, column caption, composite forwarding, node→presenter resolution, advanced/simple classification in a simulation |

UI views are exempt per the repo's testing guidance.


src/Db/Dump/PKSimDB_dump.txt
```diff
@@ -12063,14 +12063,14 @@ INSERT INTO tab_container_parameters VALUES(146,'ORGANISM','Organism','Ranz-Mars
 INSERT INTO tab_container_parameters VALUES(146,'ORGANISM','Organism','Ranz-Marshall correlation constant 2',1,1,0,0,'CONSTANT',1,'LOCAL','SIMULATION',0,0);
 INSERT INTO tab_container_parameters VALUES(146,'ORGANISM','Organism','Ranz-Marshall correlation constant 3',1,1,0,0,'CONSTANT',1,'LOCAL','SIMULATION',0,0);
 INSERT INTO tab_container_parameters VALUES(146,'ORGANISM','Organism','Ranz-Marshall correlation constant 4',1,1,0,0,'CONSTANT',1,'LOCAL','SIMULATION',0,0);
-INSERT INTO tab_container_parameters VALUES(1678,'COMPOUND','DRUG','Crititical Micellar concentration',1,0,1,0,'COMPOUND_PARTICLE_DISSOLUTION',17,'GLOBAL','COMPOUND',0,0);
+INSERT INTO tab_container_parameters VALUES(1678,'COMPOUND','DRUG','Critical Micellar concentration',1,0,1,0,'COMPOUND_ADVANCED_SOLUBILITY',17,'GLOBAL','COMPOUND',0,0);
```
msevestre, line 12066: @Yuri05 Assumng this is a typo?

src/Db/Dump/PKSimDB_dump.txt
```diff
@@ -12063,14 +12063,14 @@ INSERT INTO tab_container_parameters VALUES(146,'ORGANISM','Organism','Ranz-Mars
...
-INSERT INTO tab_container_parameters VALUES(1678,'COMPOUND','DRUG','Partition coefficient (bile salt micelle/water) ionized',1,0,1,0,'COMPOUND_PARTICLE_DISSOLUTION',14,'GLOBAL','COMPOUND',0,0);
-INSERT INTO tab_container_parameters VALUES(1678,'COMPOUND','DRUG','Partition coefficient (bile salt micelle/water) neutral',1,0,1,0,'COMPOUND_PARTICLE_DISSOLUTION',11,'GLOBAL','COMPOUND',0,0);
+INSERT INTO tab_container_parameters VALUES(1678,'COMPOUND','DRUG','Partition coefficient (bile salt micelle/water) constant 1',1,0,1,0,'COMPOUND_ADVANCED_SOLUBILITY',12,'GLOBAL','COMPOUND',0,0);
+INSERT INTO tab_container_parameters VALUES(1678,'COMPOUND','DRUG','Partition coefficient (bile salt micelle/water) constant 2',1,0,1,0,'COMPOUND_ADVANCED_SOLUBILITY',13,'GLOBAL','COMPOUND',0,0);
+INSERT INTO tab_container_parameters VALUES(1678,'COMPOUND','DRUG','Partition coefficient (bile salt micelle/water) ionized',1,0,1,0,'COMPOUND_ADVANCED_SOLUBILITY',14,'GLOBAL','COMPOUND',0,0);
+INSERT INTO tab_container_parameters VALUES(1678,'COMPOUND','DRUG','Partition coefficient (bile salt micelle/water) neutral',1,0,1,0,'COMPOUND_BILE_SALT_PARTITION_COEFFICIENT',11,'GLOBAL','COMPOUND',0,0);
```
msevestre, line 12073: @Yuri05 I had to create two groups otherwise I could not separte the parameters

src/PKSim.Assets/PKSimConstants.cs
```diff
@@ -1892,6 +1894,7 @@ public static class UI
          public static readonly string SaveProjectAsDescription = "Save the project to a new file...";
          public static readonly string CompoundTwoPoreParametersNote = "These parameters are only used in a simulation where the model for proteins and large molecules was selected. Their values are irrelevant otherwise and can be left as is.";
          public static readonly string CompoundParticleDissolutionParametersNote = "These parameters are only used in a simulation where the particle dissolution function was selected. Their values are irrelevant otherwise and can be left as is.";
+         public static readonly string CompoundAdvancedSolubilityParametersNote = "These parameters describe how bile salt micelles increase the solubility of the compound in the fluid of the intestinal lumen.";
```
msevestre, line 1897: @Yuri05 no idea?

msevestre (commented):
@Yuri05 I need to review in details one last time but this is looking good. Don't look at it now unless you really have nothign to do

src/Db/Dump/PKSimDB_dump.txt
Yuri05: Lol, it is

src/PKSim.Assets/PKSimConstants.cs
Yuri05: Probably OK for now (at least not wrong :)

Yuri05 (approved):

Yuri05:
> @Yuri05 I need to review in details one last time but this is looking good. Don't look at it now unless you really have nothign to do

Apparently it was the case, so I reviewed 😆 LGFM so far.

Yuri05:
@msevestre Can you please rename 
1 to _Advanced **Intestinal** Solubility_ and
2 to **Bile Salt Micelle Partitioning**

After that, lgfm and can be merged :)

image

PR #3672, opened 2026-08-14
Fixes #3671 OriginData summary report renders gestational age with the wrong dimension

### LINKED ISSUE #3671: OriginData summary report renders gestational age with the wrong dimension
`OriginDataReportBuilder` formats every `OriginDataParameter` through a dimension looked up
from the *unit name alone*. `week(s)` is defined in three dimensions, so gestational age is
resolved against **Age in years** and printed as `value x 52.1786`.

This is the same defect that #3549 fixed in `IndividualToIndividualBuildingBlockMapper`;
that fix did not cover this second formatter.

## Symptom

A correctly-built 30-week preterm individual is reported as `1565.36 week(s)`
(`30 x 52.1786`). The value is only wrong in the *report* — the model itself is fine.
Observed through the R package, where the summary is embedded in
`CannotCreateIndividualWithConstraintsException`:

```
Individual Parameters
Age: 0.10 year(s)
Gestational age: 1565.36 week(s)     <- individual actually holds 30 week(s)
Weight: 1.50 kg
```

The individual PK-Sim builds for the same input holds the right value:

```
Organism|Gestational age   value=30   unit=week(s)
```

The unit string is only a label, never used for the arithmetic — passing `day(s)` yields
`1565.36 day(s)`, the same factor.

## Root cause

`src/PKSim.Infrastructure/Reporting/Summary/OriginDataReportBuilder.cs:91`

```csharp
var dimension = _dimensionRepository.DimensionForUnit(originDataParameter.Unit);
...
return _formatter.Format(dimension.BaseUnitValueToUnitValue(displayUnit, originDataParameter.Value));
```

`week(s)` belongs to three dimensions:

| dimension | base unit |
| --- | --- |
| Age in weeks | `week(s)` |
| Age in years | `year(s)` |
| Time | `min` |

`DimensionForUnit` returns *Age in years*, hence the x52.1786. This is precisely the hazard
called out in the comment added by #3549:

> `DimensionForUnit` can return an incorrect dimension when the unit name is defined in multiple dimensions.

## Suggested fix

Mirror #3549: pass the statically-known dimension per field — `AgeInWeeks` for gestational
age, `AgeInYears` for age, `Length` for height, `BMI`, `Mass` for weight — and keep the
`DimensionForUnit` lookup only as the fallback for disease-state parameters, whose dimension
is not known statically.

`src/PKSim.Presentation/DTO/Mappers/OriginDataParameterToParameterDTOMapper.cs:43` makes the
same call and is worth checking for the same symptom in the desktop UI.

## Scope

Display only. The value stored in `OriginData` and used by the individual-creation algorithm
is correct and is not affected.

Reported from Open-Systems-Pharmacology/OSPSuite-R#2011, whose repro attributes the wrong
number to a dropped `gestationalAgeUnit`; the unit is in fact honoured and only this report
is wrong. Behaviour verified identical on PKSim.R 13.0.98 and 13.0.119.

Related: #3549

rwmcintosh:
Fixes #3671 

`OriginDataReportBuilder` resolved the dimension for every `OriginDataParameter` from the
unit name alone. `week(s)` is defined in three dimensions — `Age in weeks` (base `week(s)`),
`Age in years` (base `year(s)`) and `Time` (base `min`) — so `DimensionForUnit` returned
*Age in years* and gestational age was reported as `value * 52.1786`.

A correctly-built 30-week preterm individual was reported as `1565.36 week(s)`, most visibly
inside `CannotCreateIndividualWithConstraintsException`:

Reported from Open-Systems-Pharmacology/OSPSuite-R#2011.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected formatting for age, gestational age, weight, height, and BMI values by applying the appropriate measurement dimensions.
  * Gestational age is now consistently displayed using the age-in-weeks unit system, including for preterm individuals.
  * Other origin data values continue to use their specific dimensions and formatted values.

### DIFF
--- a/src/PKSim.Infrastructure/Reporting/Summary/OriginDataReportBuilder.cs
+++ b/src/PKSim.Infrastructure/Reporting/Summary/OriginDataReportBuilder.cs
@@ -87,13 +87,13 @@ namespace PKSim.Infrastructure.Reporting.Summary
          reportPart.AddPart(_reportGenerator.ReportFor(originData.AllCalculationMethods()));
       }
 
-      //Fallback for disease state parameters where the dimension is not statically known.
-      //DimensionForUnit can return an incorrect dimension when the unit name is defined in multiple dimensions.
-      private string displayValueFor(OriginDataParameter originDataParameter) =>
-         displayValueFor(originDataParameter, _dimensionRepository.DimensionForUnit(originDataParameter.Unit));
-
-      private string displayValueFor(OriginDataParameter originDataParameter, IDimension dimension)
+      //When the dimension is not known statically (disease state parameters), it is resolved from the unit
+      //name. DimensionForUnit can return an incorrect dimension when the unit name is defined in multiple
+      //dimensions, so callers pass the dimension explicitly whenever they know it.
+      private string displayValueFor(OriginDataParameter originDataParameter, IDimension dimension = null)
       {
+         dimension ??= _dimensionRepository.DimensionForUnit(originDataParameter.Unit);
+
          if (dimension != null)
          {
             var displayUnit = dimension.UnitOrDefault(originDataParameter.Unit);
--- a/src/PKSim.Infrastructure/Reporting/Summary/OriginDataReportBuilder.cs
+++ b/src/PKSim.Infrastructure/Reporting/Summary/OriginDataReportBuilder.cs
@@ -1,4 +1,5 @@
 using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.UnitSystem;
 using OSPSuite.Utility.Extensions;
 using OSPSuite.Utility.Format;
 using PKSim.Assets;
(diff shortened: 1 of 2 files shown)


rwmcintosh:
@coderabbitai review please

src/PKSim.Infrastructure/Reporting/Summary/OriginDataReportBuilder.cs
```diff
@@ -86,9 +87,13 @@ protected override void FillUpReport(OriginData originData, ReportPart reportPar
          reportPart.AddPart(_reportGenerator.ReportFor(originData.AllCalculationMethods()));
       }
 
-      private string displayValueFor(OriginDataParameter originDataParameter)
+      //Fallback for disease state parameters where the dimension is not statically known.
+      //DimensionForUnit can return an incorrect dimension when the unit name is defined in multiple dimensions.
+      private string displayValueFor(OriginDataParameter originDataParameter) =>
```
msevestre: @rwmcintosh Isn't it typpicaly an optional parameter instead of an overlaod?

msevestre (approved):
Works but I think the design is old school.

src/PKSim.Infrastructure/Reporting/Summary/OriginDataReportBuilder.cs
rwmcintosh: yes, done.

Issue #3684, opened 2026-08-18
Commit of a compound from a simulation reports it as in sync although the overwrite parameter sets differ


rwmcintosh:
Follow-up from #3679, and depends on it (it introduces the comparison of overwrite parameter sets and the `ReplaceOverwriteParameterSetsInCompoundCommand` referenced below).

Committing simulation parameters to an overwrite parameter set writes the set to the project compound only, so the compound in the simulation is correctly shown as changed. If the user then runs **Commit** on that compound (simulation -> project), `UpdateUsedBuildingBlockInfoCommand` syncs the version and clears the altered flag - asserting that the two compounds are identical - but the set is still missing from the simulation's copy. The compound turns green while the comparison still reports the missing set.

**Steps to reproduce**

1. Change a compound-dependent parameter in a simulation and commit it to a new overwrite parameter set.
2. Right-click the compound in the simulation -> Commit.
3. Compare the compound in the simulation with the compound building block: the set is reported as missing, although the compound is shown as unchanged.

**Suggested fix**

Bring the project compound's sets into the simulation's copy as part of the commit, the way `BuildingBlockParametersToSimulationUpdater` already does, placed just before `UpdateUsedBuildingBlockInfoCommand` in `SimulationParametersToBuildingBlockUpdater`:

```csharp
new ReplaceOverwriteParameterSetsInCompoundCommand(simulationCompound, templateCompound.OverwriteParameterSets).Run(_executionContext)
```

Removing the set from the project compound instead would be wrong: the sets of a compound in a simulation cannot be edited, so a missing set always means that the copy is behind and never that the user removed it, and the removal would also bypass the guard in `OverwriteParameterSetTask.RemoveSet`.

PR #3685, opened 2026-08-18
Fixes #3679 New OPS not shown in the simulation <=> compound comparison

### LINKED ISSUE #3679: New OPS not shown in the simulation <=> compound comparison
[OPS_C1.zip](https://github.com/user-attachments/files/31173277/OPS_C1.zip)

I modified some simulation compound-dependent parameters and committed them to a new OPS.
The compound is now shown as changed in the simulation (correct).

But when displaying the differences between the compound in the simulation and compound-BB: "No diffs found" is shown.

Would expect something like **"OPS xyz: Compound C1 - present, Simulation S1 - absent"** in the diff view.

rwmcintosh:
Fixes #3679

`CompoundDiffBuilder` did not compare `OverwriteParameterSets`, so a set defined in one compound and not in the other was not reported. It is now compared by name, and a new `OverwriteParameterSetDiffBuilder` compares the content of sets defined in both.

Committing simulation parameters to a set now writes it to the project compound only. The compound in the simulation is therefore correctly shown as changed, and the comparison names the missing set. The sets are brought into the compound of the simulation when the building block is updated in the simulation, by a new reversible `ReplaceOverwriteParameterSetsInCompoundCommand` that reconciles them by name in place so that a simulation having selected a set keeps its selection.

A second commit fixes the commit dialog, which compared the selected set with the sets of the template compound by reference. A selection made in the simulation configuration holds the set of the compound in the simulation, so the comparison never matched and the dialog always defaulted to creating a new set instead of offering to update the set in use.

Follow-up filed as #3684.

`dotnet test tests/PKSim.Tests/PKSim.Tests.csproj` - 5658 passed, 0 failed, 6 skipped.

## Summary by CodeRabbit

* **New Features**
  * Added support for replacing overwrite parameter sets in compounds, including undo.
  * Compound comparisons now identify missing and changed overwrite parameter sets.
  * Template overwrite parameter sets now synchronize with simulation compounds.
* **Bug Fixes**
  * Parameter commits now update the template compound consistently.
  * Improved mapping when selections originate from simulations or restored snapshots.
* **Tests**
  * Added coverage for synchronization, comparison, commits, and parameter-set mapping.

### DIFF
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -279,8 +279,8 @@ namespace PKSim.Assets
          public static string UpdateOverwriteParameterSetInCompound(string overwriteParameterSetName, string compoundName) =>
             $"Update {ObjectTypes.OverwriteParameterSet.ToLower()} '{overwriteParameterSetName}' in {ObjectTypes.Compound.ToLower()} '{compoundName}'";
 
-         public static string UpdateOverwriteParameterSetsInCompound(string compoundName) =>
-            $"Update {ObjectTypes.OverwriteParameterSet.ToLower()}s in {ObjectTypes.Compound.ToLower()} '{compoundName}'";
+         public static string ReplaceOverwriteParameterSetsInCompound(string compoundName) =>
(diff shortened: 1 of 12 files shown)


Yuri05:
@coderabbitai full review

Yuri05 (approved):

PR #3686, opened 2026-08-18
Fixes #3684 Commit of a compound from a simulation reports it as in sync although the overwrite parameter sets differ

### LINKED ISSUE #3684: Commit of a compound from a simulation reports it as in sync although the overwrite parameter sets differ

rwmcintosh:
Fixes #3684
Follow-up from #3679, which introduced the comparison of overwrite parameter sets and the `ReplaceOverwriteParameterSetsInCompoundCommand` used here.

Committing simulation parameters writes the overwrite parameter set to the project compound only. Committing that compound then cleared the altered flag on the used building block - asserting that the two compounds are identical - while the set was still missing from the simulation's copy, so the compound turned green and the comparison kept reporting the set as missing.

`SimulationParametersToBuildingBlockUpdater` now brings the project compound's sets into the simulation's copy, the way `BuildingBlockParametersToSimulationUpdater` already does on the update path, just before the used building block info is synchronized.

## Summary by CodeRabbit

* **Bug Fixes**
  * Overwrite parameter sets defined in template compounds are now synchronized with their corresponding simulation compounds.
  * Updating building blocks now correctly preserves template-defined parameter overrides.

### DIFF
--- a/src/PKSim.Core/Services/SimulationParametersToBuildingBlockUpdater.cs
+++ b/src/PKSim.Core/Services/SimulationParametersToBuildingBlockUpdater.cs
@@ -72,11 +72,23 @@ namespace PKSim.Core.Services
          var synchronizeCommand = synchronizeBuildingBlocks(templateBuildingBlock, updateTemplateParametersCommand, simulation);
          updateCommands.Add(synchronizeCommand);
 
+         //then bring over the overwrite parameter sets defined in the template so that both compounds are in sync
+         updateCommands.Add(updateOverwriteParameterSets(templateBuildingBlock, usedBuildingBlock));
+
          //now make sure that the used building block is updated with the template building block info
          updateCommands.Add(new UpdateUsedBuildingBlockInfoCommand(simulation, usedBuildingBlock, templateBuildingBlock, _executionContext).Run(_executionContext));
 
          _executionContext.PublishEvent(new BuildingBlockUpdatedEvent(templateBuildingBlock));
          return updateCommands;
+      }
+
+      private ICommand updateOverwriteParameterSets(IPKSimBuildingBlock templateBuildingBlock, UsedBuildingBlock usedBuildingBlock)
+      {
+         if (templateBuildingBlock is not Compound templateCompound)
+            return new PKSimEmptyCommand();
+
+         var simulationCompound = usedBuildingBlock.BuildingBlock.DowncastTo<Compound>();
+         return new ReplaceOverwriteParameterSetsInCompoundCommand(simulationCompound, templateCompound.OverwriteParameterSets).Run(_executionContext);
       }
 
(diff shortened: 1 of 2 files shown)


rwmcintosh:
@coderabbitai review please

src/PKSim.Core/Services/SimulationParametersToBuildingBlockUpdater.cs
```diff
@@ -72,13 +72,25 @@ public ICommand UpdateParametersFromSimulationInBuildingBlock(Simulation simulat
          var synchronizeCommand = synchronizeBuildingBlocks(templateBuildingBlock, updateTemplateParametersCommand, simulation);
          updateCommands.Add(synchronizeCommand);
 
+         //then bring over the overwrite parameter sets defined in the template so that both compounds are in sync
+         updateCommands.Add(updateOverwriteParameterSets(templateBuildingBlock, usedBuildingBlock));
```
rwmcintosh, line 76: This is saying that you are bringing the OPS to the simulation building block on commit. Would it be more valid to delete it from the project building block?

Yuri05: > Would it be more valid to delete it from the project building block?

Would not do this - it would become totally confusing for the user

Yuri05 (approved):

Issue #3680, opened 2026-08-18
OPS: Unit and value origin of committed parameters are not shown in compound


Yuri05:
Simulation: 

<img width="900" src="https://github.com/user-attachments/assets/2582334e-94a8-4e07-92a1-e051d69ee2ed" />

After committing of these parameters to a new OPS and opening this OPS in the compound: both unit and value origin of modified parameters are not shown.

further issues:
* the value of the OPS parameters are shown in compound -BB in the **internal** unit and not in the **display** unit of a parameter
* seems that both unit and value origin are **locked** in the OPS in compound - should be editable

<img width="700" src="https://github.com/user-attachments/assets/f5ccb1c4-7d0f-49be-8681-ac256ec1f76c" />

PR #3687, opened 2026-08-18
Fixes #3681 No value bounds check for OPS parameter values in a compound

### LINKED ISSUE #3681: No value bounds check for OPS parameter values in a compound
E.g. in the simulation:

---

And in the OPS in compound:

---

Later, when trying to use this OPS with invalid values in a simulation, I get the info message 

After that, the invalid parameter values are still taken over into the simulation though (because of the message I was assuming the invalid values would be rejected):

rwmcintosh:
Fixes #3681. An OPS entry in a compound stored only path and value, so the grid had no bounds check and showed values in base units with no unit.

- Commit now carries the parameter's dimension, display unit and allowed range (min/max) into the OPS entry.
- The OPS grid shows the value in its display unit and validates edits against the bounds (same messages as the simulation grid).
- The snapshot `ParameterValue` gains dimension/unit/bounds so the range survives round-trip; legacy snapshots load as before.

Out-of-range legacy values are still applied and flagged rather than blocked. Editable value origin in the OPS is tracked separately in #3680.

The misleading "parameters will not be used" reconfigure dialog seen on this issue is a separate problem (the parameter's building-block type changes from Simulation to Compound when an OPS is applied) — split out into #3695.

### DIFF
--- a/src/PKSim.Core/Services/SimulationParametersUpdater.cs
+++ b/src/PKSim.Core/Services/SimulationParametersUpdater.cs
@@ -1,4 +1,5 @@
-﻿using PKSim.Core.Model;
+﻿using System.Linq;
+using PKSim.Core.Model;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Services;
 
@@ -9,6 +10,8 @@ namespace PKSim.Core.Services
       /// <summary>
       ///    Updates all simulation parameter values from the source simulation in the target simulation. Returns a <see cref="ValidationResult"/> containing
       /// one message for each parameter that was changed in the <paramref name="sourceSimulation "/> and that does not exist in the <paramref name="targetSimulation"/> anymore.
+      /// Parameters whose value is defined by an <see cref="OverwriteParameterSet"/> selected in the <paramref name="targetSimulation"/> are not reported: their
+      /// value is not lost but taken from the overwrite parameter set.
       /// </summary>
       /// <param name="sourceSimulation">Simulation with the original values</param>
       /// <param name="targetSimulation">Simulation that will be updated</param>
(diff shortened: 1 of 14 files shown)


src/PKSim.Core/Services/SimulationParametersUpdater.cs
```diff
@@ -32,17 +35,20 @@ public ValidationResult ReconciliateSimulationParametersBetween(Simulation sourc
...
+      private ValidationResult validateParameterUpdates(PathCache<IParameter> sourceParameters, PathCache<IParameter> targetParameters, Simulation targetSimulation)
       {
          var validationResult = new ValidationResult();
          foreach (var sourceParameterKeyValue in sourceParameters.KeyValues)
          {
             if (targetParameters.Contains(sourceParameterKeyValue.Key))
                continue;
 
+            if (isDefinedByOverwriteParameterSet(targetSimulation, sourceParameterKeyValue.Key))
+               continue;
```
rwmcintosh: This guards against this message

Image

But the message is telling you that you are about to replace your hand edit with something from the building block. 

I think it should stay @Yuri05 ?

rwmcintosh:
@coderabbitai review please

src/PKSim.Core/Services/SimulationParametersUpdater.cs
Yuri05: Never ever seen this message before. Was it introduced in v13?

rwmcintosh: ok, it's my mistake trying to understand how this dialog actually comes about. What it actually means:
- a simulation has a `SIMULATION` parameter with a hand edit
- the reconfigured model does not have a `SIMULATION` parameter of this path
- conclusion: the edit is lost

In our case, the model applies an OPS to the parameter and makes it `COMPOUND`. I think I'll fire up another issue for this to separate, and also undo this guard in the PR here.

Yuri05: sounds good for me

rwmcintosh: removed

src/PKSim.Presentation/DTO/Compounds/OverwriteParameterValueDTO.cs
```diff
@@ -1,18 +1,78 @@
...
+               return Validation.ValueStrictBiggerThanMin(parameterName, dto.displayValueFor(info.MinValue), dto.Unit);
+         }
+
+         if (info.MaxValue.HasValue)
+         {
+            if (value > info.MaxValue.Value)
+               return Validation.ValueSmallerThanMax(parameterName, dto.displayValueFor(info.MaxValue), dto.Unit);
+
+            if (value == info.MaxValue.Value && !info.MaxIsAllowed)
+               return Validation.ValueStrictSmallerThanMax(parameterName, dto.displayValueFor(info.MaxValue), dto.Unit);
```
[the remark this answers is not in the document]
rwmcintosh: that goes against the convention of the code base as it is. It might be a legitimate concern, but beyond the scope of this PR. I'll create an issue for consideration

Yuri05 (approved):

PR #3293, opened 2025-10-10
3272-unnecessary-project-has-changed-save-changes-prompt

benjaperez1983:
Fixes #3272

# Description

No changes should be set to the HasChanged variable when either opening a project or a simulation from the project.

## Type of change

Please mark relevant options with an `x` in the brackets.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires documentation changes (link at least one [user](https://github.com/Open-Systems-Pharmacology/docs) or [developer](https://github.com/Open-Systems-Pharmacology/developer-docs) documentation issue):
- [ ] Algorithm update - updates algorithm documentation/questions/answers etc.
- [ ] Other (please describe):

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

Mark everything that needs to be checked before merging the PR.

- [ ] Check if the code is well documented
- [ ] Check if the behavior is what is expected
- [ ] Check if the code is well tested
- [ ] Check if the code is readable and well formatted
- [ ] Additional checks (document below if any)
- [ ] Check if documentation update issue(s) are created if the option `This change requires a documentation update` above is selected

# Screenshots (if appropriate):

# Questions (if appropriate):

Issue #3272, opened 2025-09-22
Unnecessary `Project has changed. Save changes?` prompt


Yuri05:
* Open the project attached (initially created with 12.1) [P1.zip](https://github.com/user-attachments/files/22472186/P1.zip)

* Open the only simulation
* Close the project

Security prompt `Project has changed. Save changes?` is shown - but nothing was changed in the project.

Yuri05:
might be related to https://github.com/Open-Systems-Pharmacology/MoBi/issues/2135

rwmcintosh:
It's related to changes in chart notifications, probably from batch mode. This will be a core change that affects both PK-Sim and MoBi

msevestre:
interesting

PR #3689, opened 2026-08-19
Fixes #3677 Building blocks converted again and again on every project open

### LINKED ISSUE #3677: V13 project conversion: building blocks converted again and again
[V12_BBs.zip](https://github.com/user-attachments/files/31171956/V12_BBs.zip)
The project was created with 12.3

* Open the project in V13 and show history: all building blocks were converted from v12 to v13 (correct).

* Save the project and close it.
* Reopen the project and show history: all building blocks were converted from v12 to v13 again?

This goes on with every new save/open of the project. Seems that the project version is not updated in converted building blocks?

(**Simulations** are converted properly: after the first conversion, no new history conversion entries appear after project save/reopen)

msevestre:
Fixes #3677

## Problem
Opening a pre-v13 project converts all building blocks correctly, but the conversion is never persisted: after save and reopen, the same building blocks are converted again, accumulating history entries on every open/save cycle.

## Cause
`ProjectTask.openProjectFromFile` unconditionally sets `Project.HasChanged = false` after the project is opened (introduced with #3293 to suppress a spurious save prompt). The `PKSimProject.HasChanged` setter cascades `false` into every building block, wiping the `HasChanged` flags that the converters set during deserialization. On save, a building block's content is only re-serialized when `IsLoaded && HasChanged`, so the stored old-version XML was never rewritten while the project row was stamped with the current version. Simulations were unaffected because they are lazy-loaded after the reset.

## Fix
Before resetting the project changed state, capture all loaded building blocks still flagged as changed (i.e. converted during the load), restore their flags after the reset, and keep the project flagged as changed when any conversion happened so the user is prompted to save. Opening a project already at the current version still resets the flag, preserving the fix for #3272.

### DIFF
--- a/src/PKSim.Presentation/Services/ProjectTask.cs
+++ b/src/PKSim.Presentation/Services/ProjectTask.cs
@@ -346,7 +346,7 @@ namespace PKSim.Presentation.Services
             // which is not loaded but AFTER the deserialization.
             // Ideally this should go down in the callstack, but we are getting a circular dependency issue.
             _workspace.Project.All<Individual>().Each(_lazyLoadTask.Load);
-            _workspace.Project.HasChanged = false;
+            resetProjectChangedState();
          }
 
          try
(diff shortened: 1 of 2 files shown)


src/PKSim.Presentation/Services/ProjectTask.cs
```diff
@@ -378,6 +378,16 @@ void openProject()
          }
       }
 
+      private void resetProjectChangedState()
+      {
+         var project = _workspace.Project;
+         //loaded building blocks still flagged as changed were converted during the load and must remain flagged so that the conversion is saved
+         var convertedBuildingBlocks = project.All<IPKSimBuildingBlock>().Where(x => x.IsLoaded && x.HasChanged).ToList();
```
msevestre, line 385: we need to make sure we saved which one were changed so that we can restore the flag after the HasChanged which globally resets everything

msevestre (commented):
Tested and this fixes the issue

Yuri05 (approved):

PR #3644, opened 2026-07-29
Fixes #3622 Add the model calculation methods a saved simulation needs to reconfigure

msevestre:
## What

Follow-up to #3630 (v13 project conversion). Fixes a gap found by reconfiguring/recreating a converted simulation: the model rebuild failed because parameters the new oral-absorption model introduces (for instance the pure water concentration on the organism) were never created, so the formulas referencing them could not resolve.

## Why

Those parameters belong to **model** calculation methods (e.g. the fundamental physical constants). A simulation saved before v13 keeps the calculation methods it was saved with, and the conversion was only adding the one calculation method new to v13 (`LumenSegmentVolume`) — the other model calculation methods the new model relies on were missing, so their parameters were never generated.

Running the *stored* model hid this: the parameters were baked into the saved model. It only surfaces when the model is rebuilt from the building blocks — i.e. **Reconfigure** or **create a new simulation** — which is exactly the workflow Yuri raised on #3630.

## How

`IndividualCalculationMethodsUpdater.AddMissingModelCalculationMethodsTo(simulation)` aligns a simulation's model calculation methods with the current default for its model (`ModelPropertiesTask.DefaultFor`), adding **only** the categories the simulation does not already fill, so a user-selected option (e.g. the Du Bois body-surface-area method) is never replaced. `Converter12To13` calls it after the individual has been converted, so the individual's own calculation methods have already flowed into the model properties.

## Tests

Extends the v13 conversion tests against the provided `V12_TestProject`:

- **reconfigure** every simulation — the model is rebuilt from the building blocks, which is where an incomplete conversion surfaces (verified: the test fails fast without the fix). Not run, since a broken conversion already throws at construction and running all twelve would be prohibitively slow;
- **run** one population simulation, trimmed to two individuals, to prove a rebuilt model still solves end to end;
- **create a new** oral particle-dissolution simulation from the converted standalone building blocks.

The earlier smoke test that ran all twelve stored models is removed — it ran the pre-built models (so it never exercised the rebuild) and running them all was slow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3690, opened 2026-08-19
Fixes #3640 Reconfigure and run every converted simulation of the V12 test project

### LINKED ISSUE #3640: Project conversion: cannot reconfigure old simulation
In the integration test project https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/develop/tests/PKSim.Tests/Data/V12_TestProject.pksim5
Configure the simulation **01_Ind_01_Human_Default_Healthy** and click OK (no changes in the Configure dialog).
The error below comes.

Image

When trying to re-configure the simulation **16_Pop_2pores_IV_11_Human_Modified_Healthy**, similar error message comes:

Image

From the error messages: it seems that the building blocks **in a simulation** are not (fully) converted.

Would extend the existing V13 project conversion integration tests with 
•	“For each simulation in the test project V12_TestProject: Reconfigure and Run”
•	“For each simulation in the test project V12_TestProject: Create a new simulation with the same configuration and Run”

msevestre:
## What

Extends the v13 conversion integration tests to the full matrix requested in #3640:

- **Reconfigure and run** every simulation of `V12_TestProject.pksim5` — the model is rebuilt from the simulation's converted building blocks (where an incomplete conversion surfaces as an unresolved reference) and then solved, so a defect that only shows at solve time is caught as well.
- **Create and run a new simulation** from the converted standalone building blocks for the configuration of every simulation of the project: same subject, compounds, protocols, formulation and events, on the same model with fresh default model properties — as a user would when creating a new simulation and picking the same building blocks.

These two classes replace the previous reconfigure-only test, the single population run and the single new-simulation creation, which they subsume.

## Why

The reconfigure errors reported in #3640 were fixed by #3644 (missing model calculation methods), but the test coverage the issue asked for was only partially in place: nothing ran the reconfigured models, and only one of the twelve configurations was recreated from the standalone building blocks. See the [triage notes](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3640#issuecomment-5336552289) for the full picture.

## How

- The run/trim helpers moved to the shared test-project context: individual simulations run through `ISimModelManager` so the returned `SimulationRunResults` can be inspected (the higher-level engine swallows solver errors), population runs surface failures as exceptions and are trimmed to two individuals to keep CI time reasonable.
- The new-simulation test wires events exactly as `EventBuildingBlockCreator` resolves them: the template event as used building block keyed by its template id, plus an `EventMapping` carrying the stored simulation's start time.

## Tests

The test project builds cleanly; the two new classes are running locally as this PR is opened (all twelve simulations are rebuilt and solved twice, which takes a while) — the local result will be reported in a comment, and CI verifies independently.

Fixes #3640

### DIFF
--- a/tests/PKSim.Tests/ProjectConverter/v13/Converter12To13Specs.cs
+++ b/tests/PKSim.Tests/ProjectConverter/v13/Converter12To13Specs.cs
@@ -35,10 +35,7 @@ namespace PKSim.ProjectConverter.v13
       protected void LoadAll<TBuildingBlock>() where TBuildingBlock : class, IPKSimBuildingBlock =>
          All<TBuildingBlock>().Each(Load);
 
-      /// <summary>
-      ///    Every parameter the new absorption model added to the lumen of an individual. They are looked up by name
-      ///    rather than by path so the assertion does not depend on which segments a given species defines.
-      /// </summary>
+      //Looked up by name so the assertion does not depend on which segments a species defines
       protected void ShouldHaveTheNewLumenParameters(Individual individual)
       {
          var newLumenParameterNames = new[]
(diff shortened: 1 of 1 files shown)


msevestre:
> *This was generated by AI during triage.*

Local run completed after the PR was opened — **both new test classes pass** against the converted `V12_TestProject`:

- `should_rebuild_the_model_of_every_converted_simulation_and_run_it` — passed (4 m 26 s)
- `should_create_and_run_a_new_simulation_for_the_configuration_of_every_simulation_of_the_project` — passed (4 m 29 s)

Total suite time for the two classes: ~9.3 minutes (populations trimmed to two individuals).

PR #3696, opened 2026-08-19
Fixes #3680 OPS: Unit and value origin of committed parameters are not shown in compound

### LINKED ISSUE #3680: OPS: Unit and value origin of committed parameters are not shown in compound

rwmcintosh:
Fixes #3680 

Committing a parameter to an `OverwriteParameterSet` dropped its value origin, and unit and value origin were read only in the compound.

- Commit carries the simulation parameter's `ValueOrigin` into the OPS entry, and the snapshot `ParameterValue` keeps it over a round trip. `UpdateAllFrom` is used so the id survives and a database value origin is still omitted from the snapshot.
- Unit and value origin are editable in the compound, each through its own undoable command.
- The unit follows the parameter paradigm: shown and edited inside the value cell (`UxComboBoxUnit`), and changing it keeps the displayed number while recalculating the value in base unit, as `SetParameterUnitCommand` does. The separate Unit column is gone.
- The commit dialog shows the value in its display unit plus the value origin, both read only.

Dimension, display unit and the display-unit conversion were already handled by #3687.

Read Only during commit dialog, but displayed for info
image

standard editors in the compound editor
image

The display of these two new fields was already covered in another PR, so this just makes them editable in the compound and visible in the commit dialog

## Summary by CodeRabbit

- **New Features**
  - Added editing of display units for overwrite parameter values while preserving displayed numbers.
  - Added editing and display of value-origin metadata.
  - Added unit-aware formatting and value-origin columns to parameter views.
  - Preserved value-origin information when committing parameters and saving or loading snapshots.

- **Bug Fixes**
  - Improved validation and display of parameter values using their selected units.
  - Added undo support for unit and value-origin changes.

### DIFF
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -291,6 +291,12 @@ namespace PKSim.Assets
          public static string UpdateParameterValueInOverwriteParameterSet(string parameterPath, string overwriteParameterSetName, string compoundName) =>
             $"Update value of '{parameterPath}' in {ObjectTypes.OverwriteParameterSet.ToLower()} '{overwriteParameterSetName}' of {ObjectTypes.Compound.ToLower()} '{compoundName}'";
 
+         public static string UpdateParameterValueUnitInOverwriteParameterSet(string parameterPath, string displayValue, string oldUnit, string newUnit, string overwriteParameterSetName, string compoundName) =>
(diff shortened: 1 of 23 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

PR #3697, opened 2026-08-19
Fixes #3695 Misleading "parameters will not be used" dialog when reconfiguring a simulation that uses an Overwrite Parameter Set

### LINKED ISSUE #3695: Misleading "parameters will not be used" dialog when reconfiguring a simulation that uses an Overwrite Parameter Set
When reconfiguring a simulation that uses an Overwrite Parameter Set (OPS), this info dialog can appear even when nothing is actually lost:

**Cause — the parameter building-block types don't match.** On reconfigure, `SimulationParametersUpdater.ReconciliateSimulationParametersBetween` carries over hand edits by comparing only parameters of `BuildingBlockType.Simulation` between the original and the rebuilt simulation, and warns about any that had a non-default value in the original but are missing from the rebuilt one.

Applying an OPS during model construction (`OverwriteParameterSetApplicationTask.applyValue`) flips each affected parameter from `Simulation` to `Compound`, and this runs *before* the reconciliation. So the reconciler sees the parameter as present-and-changed in the original (`Simulation`) but absent from the rebuilt one's `Simulation` set (it is now `Compound`), and reports it as a lost hand edit — even though its value is supplied by the OPS.

The message is a legitimate warning that a simulation-level hand edit is about to be replaced, but because the types no longer match after OPS application it fires for every OPS-covered parameter regardless of whether anything is genuinely lost (e.g. it fires even when the OPS value equals the hand edit).

Context: discussed in #3681 / #3687, where a guard suppressing the warning for OPS-covered paths was proposed and then reverted so this can be addressed properly here.

rwmcintosh:
Fixes #3695 

## Problem

When reconfiguring a simulation that uses an Overwrite Parameter Set (OPS), the "parameters will not be used" info dialog appeared even when nothing was actually lost.

## Root cause

On reconfigure, `SimulationParametersUpdater.ReconciliateSimulationParametersBetween` carries over hand edits by comparing only parameters of `BuildingBlockType.Simulation` between the original and the rebuilt simulation, and warns about any that had a non-default value in the original but are missing from the rebuilt one.

Applying an OPS during model construction (`OverwriteParameterSetApplicationTask.applyValue`) flips each affected parameter from `Simulation` to `Compound`, and this runs *before* reconciliation. So the reconciler sees the parameter as present-and-changed in the original (`Simulation`) but absent from the rebuilt one's `Simulation` set (it is now `Compound`), and reports it as a lost hand edit — even though its value is supplied by the OPS. This fired for every OPS-covered parameter, even when the OPS value equalled the hand edit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parameter validation when overwrite selections convert parameters into compound parameters.
  * Prevented incorrect validation warnings for parameters supplied through overwrite selections.
  * Continued reporting warnings for genuinely missing parameters.

* **Tests**
  * Added coverage to verify correct handling of converted and genuinely lost parameters.

### DIFF
--- a/src/PKSim.Core/Services/SimulationParametersUpdater.cs
+++ b/src/PKSim.Core/Services/SimulationParametersUpdater.cs
@@ -1,4 +1,6 @@
-﻿using PKSim.Core.Model;
+﻿using System.Collections.Generic;
+using System.Linq;
+using PKSim.Core.Model;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Services;
 
@@ -32,15 +34,18 @@ namespace PKSim.Core.Services
          var sourceParameters = simulationParametersCacheFor(sourceSimulation, buildingBlockType);
          var targetParameters = simulationParametersCacheFor(targetSimulation, buildingBlockType);
          _parameterSetUpdater.UpdateValues(sourceParameters, targetParameters);
-         return validateParameterUpdates(sourceParameters, targetParameters);
+         return validateParameterUpdates(sourceParameters, targetParameters, overwrittenCompoundParameterPathsFor(targetSimulation));
       }
 
-      private ValidationResult validateParameterUpdates(PathCache<IParameter> sourceParameters, PathCache<IParameter> targetParameters)
+      private ValidationResult validateParameterUpdates(PathCache<IParameter> sourceParameters, PathCache<IParameter> targetParameters, ISet<string> overwrittenCompoundParameterPaths)
       {
          var validationResult = new ValidationResult();
          foreach (var sourceParameterKeyValue in sourceParameters.KeyValues)
          {
             if (targetParameters.Contains(sourceParameterKeyValue.Key))
+               continue;
+
+            if (overwrittenCompoundParameterPaths.Contains(sourceParameterKeyValue.Key))
                continue;
 
             var sourceParameter = sourceParameterKeyValue.Value;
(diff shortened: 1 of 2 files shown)


src/PKSim.Core/Services/SimulationParametersUpdater.cs
```diff
@@ -32,17 +34,20 @@ public ValidationResult ReconciliateSimulationParametersBetween(Simulation sourc
...
-      private ValidationResult validateParameterUpdates(PathCache<IParameter> sourceParameters, PathCache<IParameter> targetParameters)
+      private ValidationResult validateParameterUpdates(PathCache<IParameter> sourceParameters, PathCache<IParameter> targetParameters, ISet<string> overwrittenCompoundParameterPaths)
       {
          var validationResult = new ValidationResult();
          foreach (var sourceParameterKeyValue in sourceParameters.KeyValues)
          {
             if (targetParameters.Contains(sourceParameterKeyValue.Key))
                continue;
 
+            if (overwrittenCompoundParameterPaths.Contains(sourceParameterKeyValue.Key))
```
rwmcintosh, line 48: If the parameter was already overwritten by OPS then it would normally be shown in this dialog because it's now a `COMPOUND` type. But for those parameters, if they are present in an OPS, then we are going to bypass the dialog warning.

rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

PR #3691, opened 2026-08-19
Fixes #3489 Add the event offset parameter to simple protocols saved before v13

msevestre:
Fixes #3489

## Problem
`Converter12To13` never visits protocols, so a `SimpleProtocol` from a pre-v13 project permanently lacks the `EVENT_OFFSET` parameter that `ProtocolFactory` only adds to newly created protocols. Enabling the new event feature on such a converted protocol showed an inert "Event offset" field (bound to a `NullParameterDTO`) and simulations silently used an offset of 0. Contrary to the original report, no null crash exists — all consumers are null-tolerant — but the feature was half-broken on converted protocols.

## Fix
- `ProtocolFactory.AddEventOffsetParameterTo(SimpleProtocol)` is now part of `IProtocolFactory`, so the factory and the converter share one definition of the parameter (value 0, time dimension, no minimum, hidden).
- `Converter12To13` visits `SimpleProtocol` building blocks and the copies each simulation carries, adding the parameter when missing (idempotent, flags the conversion).

## Tests
- Unit: converting a degraded pre-v13 protocol restores the parameter with the factory definition; converting a protocol that already has it changes nothing.
- Integration: every simple protocol of the V12 test project — standalone and simulation-owned — carries the parameter after conversion.

Targeted specs are green; the full v13 converter suite (including the reconfigure-and-run specs) is running and results will be posted here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3617, opened 2026-07-22
Feature/event protocol

msevestre:
Fixes #

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

Please mark relevant options with an `x` in the brackets.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires documentation changes (link at least one [user](https://github.com/Open-Systems-Pharmacology/docs) or [developer](https://github.com/Open-Systems-Pharmacology/developer-docs) documentation issue):
- [ ] Algorithm update - updates algorithm documentation/questions/answers etc.
- [ ] Other (please describe):

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Integration tests
- [ ] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

Mark everything that needs to be checked before merging the PR.

- [ ] Check if the code is well documented
- [ ] Check if the behavior is what is expected
- [ ] Check if the code is well tested
- [ ] Check if the code is readable and well formatted
- [ ] Additional checks (document below if any)
- [ ] Check if documentation update issue(s) are created if the option `This change requires a documentation update` above is selected

# Screenshots (if appropriate):

# Questions (if appropriate):

PR #3698, opened 2026-08-19
Fixes #3648 Event protocol review follow-ups

### LINKED ISSUE #3648: PR Review Event protocol
## Silent auto-selection of default event

In `SimulationCompoundProtocolEventPresenter.createEventMapping()`, when a used event key has no existing mapping, it falls back to `defaultEvent()`, which returns `_eventTask.All().FirstOrDefault()` — an arbitrary event in the project — instead of leaving the mapping unset. This silently pre-selects an unrelated event building block for the user, which can produce wrong simulations if the user doesn't notice the pre-filled ("guessed") value; it also weakens the "all placeholders must be mapped" validation this PR describes, since a value is always present even if never intentionally chosen. [[github](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3617)](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3617)

## Potential null-reference in `SaveConfiguration`

If a project has no `PKSimEvent` building blocks at all, `defaultEvent()` returns `null`, so `dto.Event` stays `null`. `SaveConfiguration()` then calls `_eventTask.Load(dto.Event)` and reads `dto.Event.Id` unconditionally, which would throw a `NullReferenceException` rather than surfacing a friendly validation message. [[github](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3617)](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3617)

## Exception-based error handling in `EventBuildingBlockCreator.resolveEvent`

`resolveEvent` throws a generic `PKSimException` when a template event id can't be resolved, rather than being guarded earlier by the "all placeholders mapped" validation rule; if that validation is ever bypassed (e.g. programmatic simulation creation, snapshot import with stale template ids), the exception message doesn't clearly point to which protocol/schema item is at fault. [[github](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3617)](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3617)

## Business rule only partially wired for `SchemaItem` vs `SimpleProtocol`

`SchemaItemRules.eventKeyValid` and `SimpleProtocol.eventPlaceholderValid` implement essentially the same "event key required when IsEvent/HasEvent" rule independently in two places (`SchemaItem.cs` and `SimpleProtocol.cs`). This duplication is a maintenance risk — a future change to the validation message or condition in one place could easily be missed in the other since they aren't sharing a common rule definition despite very similar semantics. [[github](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3617)](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3617)

## Negative start-time bound relaxation applied ad hoc

Both `SchemaItem.allowNegativeStartTimeForEvents()` and `EventBuildingBlockCreator.allowNegativeStartTime()` independently null out `MinValue`/set `MinIsAllowed` for event start times, rather than centralizing this in one helper. Any future third code path that creates/clones event start-time parameters (e.g. via a different mapper or snapshot import) could forget this step and reject legitimate negative offsets. [[github](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3617)](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3617)

## No handling for zero available events in event mapping UI

`AllEventsFor()` builds selections from `_eventTask.All()`; if a project has zero `PKSimEvent` building blocks, the dropdown will be empty, and combined with the `defaultEvent()` null fallback above, users could hit the mapping step with no valid event to select and no clear guidance beyond "create new" / "load from template" buttons — this UX edge case doesn't appear to be tested in the diff. [[github](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3617)](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3617)

_Originally posted by @Yuri05 in https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3617#issuecomment-5195599768_

msevestre:
Fixes #3648

# Description

Code-quality follow-ups from the review of the event-protocol feature (PR #3617), as specified in the agent brief on #3648. Rebased onto `develop` after #3691 (event offset parameter) was merged.

- **Clear unresolved-event error** – `EventBuildingBlockCreator` now throws `NoEventFoundException` (mirrors `NoFormulationFoundForRouteException`) naming the protocol and placeholder key, or the template id of an unresolvable standalone event mapping, instead of `PKSimException("Could not resolve event building block for template id ''.")`.
- **Dead rule removed** – `SimpleProtocol.eventPlaceholderValid` was a tautology (`HasEvent` ⇔ `EventKey` non-empty); removed together with the unused `PKSimConstants.Error.EventPlaceholderRequired`. `SchemaItemRules.eventKeyValid` remains the single rule.
- **One helper for "allow negative time"** – new `IParameter.AllowNegativeValues()` replaces the three copies of `Info.MinValue = null; Info.MinIsAllowed = true;` in `SchemaItem`, `EventBuildingBlockCreator` and `ProtocolFactory.AddEventOffsetParameterTo`.
- **Presenter hardening** – `SimulationCompoundProtocolEventPresenter.SaveConfiguration()` skips unmapped placeholders instead of dereferencing a null event, and no longer creates a blank `EventSelectionDTO` when the project has no event (so `AllEventsFor` is empty instead of offering a blank entry).
- Default event pre-selection (finding 1 of the issue) is intentionally unchanged, per triage decision.

## Type of change

- [ ] This change requires documentation changes
- [ ] Algorithm update
- [x] Other (please describe): code-quality / hardening follow-ups, dev-only

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [ ] Manual tests
- [ ] No tests required

New / updated specs:
- `EventBuildingBlockCreatorSpecs`: unmapped placeholder, placeholder mapped to an event not used by the simulation, unresolvable standalone mapping – each asserts `NoEventFoundException` and the exact message.
- `ParameterExtensionsSpecs`: `AllowNegativeValues()` removes the bound, tolerates null `Info` / null parameter.
- `SimpleProtocolEventSpecs`: protocol with / without event and after clearing the event is valid.
- `SimulationCompoundProtocolEventPresenterSpecs`: zero-events project (null selection, invalid DTO, empty `AllEventsFor`, mapping visible) and save with an unmapped placeholder (no throw, no mapping, no `Load`).

Full `PKSim.Tests` suite run locally: 5677 passed, 0 failed, 6 skipped.

# Reviewer checklist

### DIFF
--- a/src/PKSim.Assets/PKSimConstants.cs
+++ b/src/PKSim.Assets/PKSimConstants.cs
@@ -573,7 +573,6 @@ namespace PKSim.Assets
          public static string FormulationIsRequiredForType(string applicationType) => $"Formulation is required for type '{applicationType}'.";
 
          public static readonly string EventKeyRequired = "Event key is required for event entries.";
-         public static readonly string EventPlaceholderRequired = "A placeholder is required when 'Administer with event' is selected.";
 
          public static string BuildingBlockNotDefined(string buildingBlock) => $"No {buildingBlock} defined. Please use create.";
 
@@ -657,6 +656,12 @@ namespace PKSim.Assets
 
          public static string NoFormulationFoundForRoute(string protocolName, string applicationRoute) =>
             $"No formulation found for route '{applicationRoute}' in administration protocol '{protocolName}'.";
+
+         public static string NoEventFoundForPlaceholder(string protocolName, string eventKey) =>
+            $"No event found for placeholder '{eventKey}' in administration protocol '{protocolName}'.";
+
+         public static string NoEventFoundForSimulationEvent(string templateEventId) =>
+            $"No event found for simulation event with template id '{templateEventId}'.";
 
          public static string UnableToCreateSimulationWithMoleculesHavingSameName(string duplicateName)
          {
--- a/src/PKSim.Core/Model/NoEventFoundException.cs
+++ b/src/PKSim.Core/Model/NoEventFoundException.cs
@@ -1,0 +1,19 @@
+using PKSim.Assets;
+
+namespace PKSim.Core.Model
+{
+   public class NoEventFoundException : PKSimException
+   {
+      //event placeholder of a protocol that could not be resolved to an event building block
+      public NoEventFoundException(Protocol protocol, string eventKey)
+         : base(PKSimConstants.Error.NoEventFoundForPlaceholder(protocol.Name, eventKey))
+      {
+      }
+
+      //standalone simulation event mapping that could not be resolved to an event building block
+      public NoEventFoundException(string templateEventId)
+         : base(PKSimConstants.Error.NoEventFoundForSimulationEvent(templateEventId))
+      {
+      }
+   }
+}
--- a/src/PKSim.Core/Model/ParameterExtensions.cs
+++ b/src/PKSim.Core/Model/ParameterExtensions.cs
@@ -73,6 +73,15 @@ namespace PKSim.Core.Model
          return parameter;
       }
 
+      //removes the lower bound so that a time parameter can be negative (e.g. an event scheduled before an administration)
+      public static void AllowNegativeValues(this IParameter parameter)
+      {
+         if (parameter?.Info == null) return;
+
+         parameter.Info.MinValue = null;
+         parameter.Info.MinIsAllowed = true;
+      }
+
       public static bool NeedsDefault(this IParameter parameter)
       {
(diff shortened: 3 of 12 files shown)


msevestre:
@rwmcintosh Really some tidying up of code and overall cleanup

Yuri05 (approved):

PR #3703, opened 2026-08-20
Fixes #3682 No differences compound <=> simulation are shown when OPS parameters are modified in the simulation

### LINKED ISSUE #3682: OPS: No differences compound <=> simulation are shown when OPS parameters are modified in the simulation
[OPS_C1.zip](https://github.com/user-attachments/files/31175000/OPS_C1.zip)

E.g. in the project attached the highlighted parameters are part of an OPS which used in the simulation:

---
After changing the parameters in the simulation: no differences between the compound in simulation ind compound BB are shown

rwmcintosh:
Fixes #3682
The parameters an overwrite parameter set overwrites are compound-dependent *simulation* parameters with no counterpart in the compound building block. Applying a set nevertheless set `Origin.BuilingBlockId` and `BuildingBlockType = Compound`, which broke change reporting in both directions:

- the origin flipped the altered flag of the compound in the simulation, so it turned red and offered *Show Differences* — but with no origin parameter to write to, the two compounds stayed identical and the comparison reported "No difference found";
- the building block type hid the edit from the change tracker, so it was never offered in *Commit Simulation Parameters to Compound* and never reached the project compound for the comparison to report.

The origin is no longer set, and the tracker now also picks up parameters overwritten by the set applied to the simulation. `BuildingBlockType` stays `Compound` so these parameters keep losing the reset behaviour of a simulation parameter.

**Second commit** — applying a set also stamped a synthetic `Overwrite Parameter Set '<name>'` value origin onto the parameter and ignored the one stored in the entry, so a value origin recorded in the compound survived a single commit and was then overwritten by the label on the next one. The label predates entries having a value origin of their own; it is dropped, and the value origin is now applied together with the value. Committing this here because the first commit is what makes the overwriting reachable — previously these parameters could not be committed at all.

> [!IMPORTANT]
> Projects saved before this change persist the bad origin and `altered="1"`, so reopening one still shows the old behaviour — the fix applies when the model is rebuilt. Verify with a simulation configured after this change.

### DIFF
--- a/src/PKSim.Core/Services/OverwriteParameterSetApplicationTask.cs
+++ b/src/PKSim.Core/Services/OverwriteParameterSetApplicationTask.cs
@@ -4,7 +4,6 @@ using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Domain.Services;
 using OSPSuite.Utility.Extensions;
-using PKSim.Assets;
 using PKSim.Core.Model;
 
 namespace PKSim.Core.Services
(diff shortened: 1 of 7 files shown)


rwmcintosh:
@coderabbitai review please

src/PKSim.Core/Services/OverwriteParameterSetApplicationTask.cs
```diff
@@ -99,29 +98,28 @@ private void removeAdvancedParametersFor(Simulation simulation,
...
       }
 
-      private static void applyValue(IParameter parameter, ParameterValue parameterValue, OverwriteParameterSet overwriteParameterSet)
+      private static void applyValue(IParameter parameter, ParameterValue parameterValue)
       {
          parameter.Value = parameterValue.Value.Value;
          parameter.BuildingBlockType = PKSimBuildingBlockType.Compound;
-         parameter.ValueOrigin.Source = ValueOriginSources.Other;
-         parameter.ValueOrigin.Description = $"{PKSimConstants.ObjectTypes.OverwriteParameterSet} '{overwriteParameterSet.Name}'";
+         parameter.ValueOrigin.UpdateAllFrom(parameterValue.ValueOrigin);
```
rwmcintosh, line 122: No synthetic ValueOrigin now. We always just use the one from the OPS. It's editable and it's also sourced from the original simulation commit to building blocks.

Yuri05 (approved):

PR #3707, opened 2026-08-24
Show the orange uncommitted-changes indicator on the simulation icon

### LINKED ISSUE #3660: Show the orange "!" in the simulation icon when a simulation parameter is changed
Right now, the sign only appears in the compounds icon, but the simulation does not indicate any change. I think we should also add the "!" to the simulation icon once a simulation parameter is changed for any compound. Otherwise, the users will not see this change.

rwmcintosh:
Overlays the orange "!" badge on the simulation icon when it has uncommitted compound-parameter changes, so the change is visible at the simulation level (previously only shown on the compound icon).

Bumps the OSPSuite packages to 13.0.159 for the new overlay icons (OSPSuite.Core#2955).

Also bumps `LAYOUT_VERSION` so persisted ribbon/dock layouts reset — the new icons shift the ordinal image indices, which would otherwise show stale icons for existing dev/beta profiles.

Fixes #3660

## Summary by CodeRabbit

- **New Features**
  - Simulations now indicate when they contain uncommitted parameter changes.
  - Simulation icons display an orange overlay when changes are pending, while preserving the simulation’s status color.

- **Improvements**
  - Updated the application’s underlying platform components for improved compatibility and consistency.
  - Updated the layout version to support the latest simulation data format.

### DIFF
--- a/src/PKSim.Assets/PKSim.Assets.csproj
+++ b/src/PKSim.Assets/PKSim.Assets.csproj
@@ -22,9 +22,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.153" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.153" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.153" />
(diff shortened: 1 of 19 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

Issue #3711, opened 2026-08-25
Add a --cores option to PKSim.CLI to control the degree of parallelism


rwmcintosh:
**TL;DR** — PKSim.CLI always uses `ProcessorCount − 1` cores for its parallel work (running simulations during snapshot loads, and model construction once #3710 lands) with no way to override it. Add a `--cores <n>` option so callers can control it — in particular the QualificationRunner, which spawns one PKSim.CLI process per project and will pass `--cores 1` so that parallelism is managed at the process level.

### Summary

The CLI host hardcodes its degree of parallelism at bootstrap (`CLIUserSettings.cs:31`):

```csharp
MaximumNumberOfCoresToUse = Math.Max(Environment.ProcessorCount - 1, 1);
```

No verb exposes an option to change it. This one value drives:

- the parallel simulation runs during snapshot/project loads (`ProjectMapper.runParallelSimulations`),
- parallel model construction during snapshot loads (#3710),
- parameter identification / sensitivity analysis engines.

When the QualificationRunner processes a qualification plan it spawns one PKSim.CLI process per project, throttled at `Environment.ProcessorCount` concurrent processes. Each child then runs its own internal work with `ProcessorCount − 1` workers — on a 20-core machine that is up to 20 × 19 concurrent workers. This oversubscription exists **today** for the simulation-run phase and would extend to model construction with #3710.

### Proposed work

- Add a `--cores <n>` option (n ≥ 1) to the shared CLI command base so every verb (`snap`, `qualification`, `run`, `export`, ...) accepts it.
- When provided, set `ICoreUserSettings.MaximumNumberOfCoresToUse` after container initialization, before the runner starts.
- When absent, keep the current default (`ProcessorCount − 1`) so interactive and single-project CLI use keeps full parallelism.

### Design decision (agreed 2026-08-25)

Parallelism gets exactly one owner per level: the QualificationRunner manages process-level fan-out and passes `--cores 1` to each child; a standalone CLI invocation owns the whole machine by default. See the companion issues below.

### Companion issues

- MoBi side: Open-Systems-Pharmacology/MoBi#2519
- QualificationRunner side: Open-Systems-Pharmacology/QualificationRunner#189
- Motivation/measurements: #3710

Issue #3709, opened 2026-08-25
Building-block lazy-load/registration chain is not thread-safe and is already exercised concurrently


rwmcintosh:
**TL;DR** — Importing simulations from a snapshot already maps them concurrently (`Task.WhenAll` in shipping code), and that mapping lazy-loads template building blocks through a chain that deserializes content into shared instances and bulk-mutates a global, unsynchronized id registry. Whether a multi-simulation import into a `.pksim5`-loaded project corrupts state is a timing lottery today. This needs fixing on its own merits — and it is a prerequisite for parallelizing snapshot project load (#3710).

### The concurrent driver that exists today

`LoadSimulationFromSnapshotPresenter.cs:57-60` ("Import simulation from snapshot"):

```csharp
var tasks = snapshots.Select(x => _simulationMapper.MapToModel(x, simulationContext));
return await Task.WhenAll(tasks);
```

All simulations in the file are mapped concurrently, resolving template building blocks against the same live project (`_projectRetriever.Current`).

### The chain

1. `SimulationMapper.findBuildingBlock` (`SimulationMapper.cs:513-521`) calls `_executionContext.Load(buildingBlock)` for every template building block a simulation references (individual, compound, protocol, ...).
2. `LazyLoadTask.Load` (`LazyLoadTask.cs:41-61`) is a check-then-act: `IsLoaded` is checked at `:43` and set at `:60`; in between sits the entire content load (database read + XML deserialization). Two threads loading the same building block both pass the check.
3. `loadObjectBase` (`:85-97`) then runs `_registrationTask.Unregister(...)` → `_contentLoader.LoadContentFor(...)` → `_registrationTask.Register(...)`. `LoadContentFor` deserializes the building block's content **into the shared instance** — two threads doing this to the same `Individual` is concurrent structural mutation of its parameter hierarchy.
4. `RegisterObjectVisitor.cs:38-44` walks the **whole object graph**, doing per-node `Unregister` + `Register` into the global Singleton `IWithIdRepository` (`OSPSuite.Core CoreRegister.cs:90`), whose implementation is a non-thread-safe check-then-act over a plain `Cache` (`WithIdRepository.cs:18-37`). So even loads of *different* building blocks race on the registry. Core-side hardening is proposed in Open-Systems-Pharmacology/OSPSuite.Core#2958.

[...]

PR #3712, opened 2026-08-25
Fixes #3710 Parallelize simulation model construction during snapshot load

### LINKED ISSUE #3710: Parallelize simulation model construction during snapshot/project load
**TL;DR** — Loading a snapshot builds every simulation's model one at a time, even though the model constructor is now thread-safe (Open-Systems-Pharmacology/OSPSuite.Core#2886) and this same loader already runs the finished simulations in parallel. Parallelizing the construction loop speeds up project load in the UI, `loadSimulationsFromSnapshot()` / `runSimulationsFromSnapshot()` in R, and CLI/qualification runs — model construction is ~35% of CPU in a qualification run (Open-Systems-Pharmacology/OSPSuite.Core#2793). Proposing for V13.

### Summary

When a project is loaded from a snapshot, each simulation's model is rebuilt sequentially in `ProjectMapper.allSimulationsFrom` (`ProjectMapper.cs:338-352`):

```csharp
//do not run tasks in parallel as the same mapper instance may be used concurrently to load two different snapshots
foreach (var snapshot in snapshots)
{
   try
   {
      var simulation = await _simulationMapper.MapToModel(snapshot, simulationContext);
      simulations.Add(new(simulation, snapshot));
      simulationContext.NumberOfSimulationsLoaded++;
   }
   catch (Exception e)
   {
      _logger.AddException(e);
      _logger.AddError(PKSimConstants.Error.CannotLoadSimulation(snapshot.Name));
   }
}
```

Each iteration runs `SimulationMapper.MapToModel` → `SimulationModelCreator.CreateModelFor` → `IModelConstructor.CreateModelFrom` (`SimulationModelCreator.cs:59` — the only call site of the model constructor in the repo).

Every snapshot surface funnels through this one loop, so parallelizing it benefits:

- **UI**: Load project from snapshot (`LoadProjectFromSnapshotUICommand` → `ProjectTask.cs:243`)
- **R**: `loadSimulationsFromSnapshot()` (`PKSim.R SnapshotTask.LoadSimulationsFromSnapshot`, `SnapshotTask.cs:63`)
- **R/CLI**: `runSimulationsFromSnapshot()` (`JsonSimulationRunner.cs:140`) and snapshot ↔ project conversion (`SnapshotRunner.cs:92`)
- **CLI**: qualification runs (`QualificationRunner.cs:51-56`)

Note: the snapshot path already parallelizes the simulation **run** (`ProjectMapper.runParallelSimulations`, `:173-193`, `Parallel.ForEachAsync` with `MaximumNumberOfCoresToUse`) — only the model **construction** is serial.

### History — this loop was parallel once

Until May 2018 this was `_simulationMapper.MapToModels(...)` — a `Task.WhenAll` fan-out. #690 (fixing #684/#648: "parameters not found" warnings and failed loads) serialized it, because the mapping chain was not safe to run concurrently.

What changed since:

- `ModelConstructor` and all its sub-tasks are now stateless (Open-Systems-Pharmacology/OSPSuite.Core#2886, shipped in the current 13.0.x packages), proven by an 8-way parallel integration test comparing structural fingerprints against a sequential baseline.
- The remaining shared state in the PK-Sim chain is identified and gets explicit fixes (#3709) instead of a blanket revert.
- MoBi already ships parallel model construction in its UI batch update (Open-Systems-Pharmacology/MoBi#2189) and has the mirror issue for its own snapshot load (Open-Systems-Pharmacology/MoBi#2424).
- Core-side parallel construction precedent has been in production for years: `CategorialParameterIdentificationRunInitializer.cs:41-78` runs full PK-Sim model construction under `Parallel.ForEach`.

### Proposed work

Replace the `foreach` with `Parallel.ForEachAsync`, modeled on the adjacent `runParallelSimulations` (`ProjectMapper.cs:173-193`):

- degree of parallelism from `ICoreUserSettings.MaximumNumberOfCoresToUse`
- per-item try/catch preserving current semantics (one bad simulation logs `CannotLoadSimulation` and does not abort the load)
- `Interlocked` (or per-item contexts) for `SimulationContext.NumberOfSimulationsLoaded`
- collect results in a thread-safe collection; keep project mutation (`addSimulations`, classifications) in a sequential commit phase

Favorable facts: `SimulationMapper`/`ProjectMapper` hold no mutable instance fields (all readonly injected services); the snapshot path already sets `ShowProgress = false` (`SimulationModelCreator.cs:14`), so no Core progress interleaving; and `LoadSimulationFromSnapshotPresenter.cs:57-60` already maps simulations with `Task.WhenAll` today (single-simulation import), so concurrent mapping is not new territory — it just needs to be made safe.

### Prerequisites

- [ ] #3709 — thread-safety of the building-block lazy-load/registration chain (already racy today under the `Task.WhenAll` above)
- [ ] Open-Systems-Pharmacology/OSPSuite.Core#2958 — Core-side invariant guards (`WithIdRepository`, validator lifestyles)
- [ ] Verify every host (UI, CLI, R) awaits `StartableWarmup.AwaitCompletion()` before the parallel region, so the singleton ORM repositories (`StartableRepository.Start()` is documented not thread-safe) are read-only by then

### Follow-up candidates (out of scope here)

- Parallelize the building-block mapping loop (`mapSnapshotToBuildingBlocks`, `ProjectMapper.cs:376-403`) — same shape, likely a bigger win on individual/population-heavy snapshots; must preserve the expression-profiles-first ordering (`:359-361`).
- `LoadSimulationsFromSnapshot(file, names)` builds **every** simulation in the snapshot and filters afterwards (`PKSim.R SnapshotTask.cs:67-71`), and rebuilds each returned simulation's configuration a second time (`:93`) — constructing only the requested simulations is an orthogonal win for R users.

rwmcintosh:
Fixes #3710

> Requires OSPSuite.Core **13.0.165** (thread-safe `ModelConstructor` chain: Open-Systems-Pharmacology/OSPSuite.Core#2959, Open-Systems-Pharmacology/OSPSuite.Core#2965) and the merged #3716 lazy-load thread-safety fix — both in `develop`.
>
> **Merge ordering**: qualification workflows spawn one PKSim.CLI child per project; until #3711 / Open-Systems-Pharmacology/QualificationRunner#189 land (`--cores 1` for children), a qualification run multiplies process count by construction parallelism. This PR should merge together with #3711 or with that caveat accepted.

### What changes

- `ProjectMapper.allSimulationsFrom` maps simulations sequentially until one succeeds (warming remaining lazily initialized services), then maps the rest under `Parallel.ForEachAsync` bounded by `MaximumNumberOfCoresToUse`. Results are collected per index so the project receives simulations in snapshot order, and per-simulation error isolation is preserved. This one loop serves the UI project load, `loadSimulationsFromSnapshot()` in R, and the CLI/qualification runners. The warm-up and fan-out only run when there is something to parallelize: single-simulation snapshots and `MaximumNumberOfCoresToUse = 1` load sequentially with lazy initialization exactly as before.
- `StartableWarmup` is a proper latch: the warm-up task is created under a lock and never replaced, concurrent callers block on the same task, and hosts that never call `Begin` (CLI, R, qualification) warm up deterministically on first use by resolving every `IStartable` themselves. A failure is scoped to the startable that failed: it is logged by name, the remaining startables still start, and the failed one stays cold — its only retry is the pre-existing lazy initialization at first use, since the warm-up must not re-run `Start()` on a repository that may already have mutated part of its state (making that retry safe is OSPSuite.Utility#60; most of the 39 `DoStart` implementations append). `AwaitCompletion` reports whether everything warmed instead of throwing: a load with a cold repository degrades to sequential processing — construction *and* runs — with a warning naming the reason, and the UI starts fully initialized regardless. An out-of-memory failure during warm-up faults it and rethrows on every call. `RemoteTemplateRepository` treats its missing (best-effort provisioned) summary file as an expected state (debug-logged), and its `DoStart` resets before filling so a lazy retry cannot duplicate content.
- Out-of-memory failures never degrade the load silently: a shared `Exception.IsOutOfMemory()` extension (iterative, cycle-detecting via a visited set, unwrapping `AggregateException` and inner-exception chains of any width) backs every broad catch on the load path. During the load phase an OOM fails the load; during the optional run phase it aborts the runs with a named error but keeps the loaded project, and the analyses then map as in a no-run load. Cancelled runs are logged — except the collateral cancellations that follow an OOM abort — and a run phase with cancellations does not report "All Simulations Finished Running."
- Progress is reported as completions ("Loading 60 Simulation(s)…", then "Simulation 'X' loaded (k/60)"); each mapping also logs a debug start line. A failed simulation logs the exception and an error naming it, both under the project category, and a load that lost simulations (including a mapper returning null) ends with an "Only k/N Simulation(s) Loaded" warning. The degree of parallelism is logged at debug level per phase. The now-unused `SimulationContext` counters are retired (Core cleanup: Open-Systems-Pharmacology/OSPSuite.Core#2966).
- Model construction runs with the core progress stream suppressed (`SimulationModelCreator`'s default), so parallel constructions cannot interleave it — pinned by a spec.
- `Cloner` resolves its per-operation clone manager per call instead of mutating the injected singleton's `FormulaCache` (this corrupted 59/60 simulations in the first parallel benchmark run); the load-bearing transient lifestyle is documented.
- The configuration pipeline (`SimulationConfigurationTask` → spatial structure → ORM queries) is not thread-safe and runs serialized on a single lock — the configuration all benchmarks below ran with. Only the thread-safe `IModelConstructor.CreateModelFrom` runs unserialized on parallel workers. Parallelizing the configuration pipeline is the follow-up with the next ~2-3×.
- `ServerGarbageCollection` is enabled for PKSim.exe, PKSim.CLI and PKSim.BatchTool — parallel construction is allocation-heavy and workstation GC forfeits most of the gain (below).

### Measurements

Pure snapshot load (no simulation runs, no project save), 20 logical cores, Rifampicin-Midazolam-DDI from the CYP3A4 DDI qualification plan — 60 simulations, 2 individuals:

| Pure load, 60 sims | Workstation GC | Server GC |
|---|---|---|
| Sequential | 670 s | 399 s |
| **Parallel (this PR)** | ~700 s | **121 s** |

The workstation-GC parallel cell was an approximate, stopped-early harness run under ~90% system memory. In desktop A/B runs on this branch (workstation GC on both sides), parallel outperformed sequential ~1.7× (9 sims: 99 s vs 173 s; 60 sims: ~12 min vs 20.3 min) at a higher memory footprint — server GC is what unlocks the full gain.

Same-process A/B in the desktop app (server GC, 1 vs 19 cores, measured with temporary phase instrumentation): construction 8 m 01 s → 2 m 31 s warm / 1 m 37 s cold (**3.2–5.0×, parallelism alone**); simulation runs 25 m 25 s → 2 m 30 s (~10×, the run fan-out predates this PR).

User-witnessed against the shipped release (13.0.49 portable), "run simulations" unticked:

| GUI load | 9 sims | 60 sims |
|---|---|---|
| Released PK-Sim 13.0.49 (sequential, workstation GC) | 173 s | 20.3 min |
| This PR, workstation GC | 99 s | ~12 min |
| **This PR + server GC** | **41 s** | **132 s** |
| **Speedup vs release** | **4.2×** | **9.2×** |

### Tests

- The ordering spec stubs `MaximumNumberOfCoresToUse = 3` so the parallel path actually runs, with staggered completions verifying simulations are still added in snapshot order.
- A handshake spec pins sequential-until-first-success: three snapshots, the first two fail, and the third mapping must not start while the second is still running (verified by mutation against an immediate fan-out).
- The corrupted-snapshot fixture covers error isolation, the named failure log, and the partial-load warning; a null-mapping fixture pins that a simulation mapped to nothing does not count as loaded; a failing-run fixture pins that the remaining count includes failed runs.
- Out-of-memory fixtures pin that a wrapped `OutOfMemoryException` fails the load without being logged as a per-simulation error; `ExceptionExtensions` specs cover direct, nested, aggregate (including one 64 siblings wide), null and cyclic exception chains. A cancelled-run fixture pins the cancellation log and the suppressed finished message.
- `StartableWarmup` specs cover the headless self-start (exactly once across callers), `Begin`-provided startables, concurrent callers sharing one warm-up, a failing startable (warm-up reports incomplete, still starts the rest, logs the failure by name, never runs `Start` again on the next call), and an out-of-memory warm-up rethrowing on every call. `ProjectMapper` fixtures pin the gate for both loops: a single-simulation load never awaits the warm-up, and an incomplete warm-up constructs and runs everything on a single core with a warning.
- A `Cloner` spec asserts a fresh clone manager per `Clone` call; a `SimulationModelCreator` spec pins that construction runs with core progress suppressed (verified by mutation).
- All ProjectMapper specs and all 9 PKSim.R end-to-end snapshot-loading specs pass.

Related: `--cores` control is #3711 / Open-Systems-Pharmacology/QualificationRunner#189. Sibling MoBi change: Open-Systems-Pharmacology/MoBi#2520 (merged).


rwmcintosh:
@coderabbitai review please

Yuri05:
@rwmcintosh **Review No. 1**

I reviewed the diff plus the surrounding code on the branch. Issues only, grouped by severity.

## Correctness

**The warm-up guarantee is heuristic, not a guarantee**
`allSimulationsFrom` maps simulations sequentially until one succeeds, then fans out, on the premise that the first success has warmed "the lazily initialized repositories." A single successful mapping only touches the repositories that *its own* path needs. A snapshot whose first simulation is an individual simulation does not warm the population/gene-expression/aging paths, so the first population simulation still initializes those repositories concurrently — exactly the race the warm-up is meant to prevent. The deterministic fix is to explicitly `Start()`/resolve the required repositories before the fan-out rather than inferring warm-up from an unrelated success.

**`StartableWarmup.AwaitCompletion()` is one-shot and unsynchronized, and correctness now depends on it**
`_warmupTask` is a non-volatile field that `AwaitCompletion` nulls out before awaiting ([[StartableWarmup.cs](https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/3710-parallel-snapshot-construction/src/PKSim.Infrastructure/Services/StartableWarmup.cs)](https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/3710-parallel-snapshot-construction/src/PKSim.Infrastructure/Services/StartableWarmup.cs)). A second caller — and `PKSimMainViewPresenter` already calls it — races on that field and returns immediately while warm-up is still running. Previously that was harmless (worst case, the UI thread did the loading itself); now `ProjectMapper` relies on it as a precondition for parallel construction. It needs a latch/`Lazy`-style implementation that all callers block on, not a self-clearing field.

**`Begin` is only called from the UI host**
`container.Resolve<IStartableWarmup>().Begin(...)` exists only in `PKSim.UI/BootStrapping/ApplicationStartup.cs`. In PKSim.CLI, the qualification runner and PKSim.R, `AwaitCompletion()` is a documented no-op, so those hosts rely entirely on the heuristic first-simulation warm-up above. The comment "the startable repositories must finish initializing before models are constructed in parallel" is not true in the headless hosts.

## Server GC does not reach two of the affected hosts

Your own measurements show parallel + workstation GC is *no better than sequential* (~700 s vs 670 s) while holding ~8 GB. But `ServerGarbageCollection` was added only to `PKSim.csproj` and `PKSim.CLI.csproj`:

- **PKSim.R is a library** (`net10.0`, no `OutputType`), so its GC configuration comes from the R host process. `loadSimulationsFromSnapshot()` takes the new parallel path with workstation GC and therefore inherits the worst cell of your table — slower than today plus a large memory spike — with no way to opt out.
- **PKSim.BatchTool** is a `WinExe` that loads snapshots via `SnapshotRunPresenter` and did not get the property.

Either add the property to BatchTool and document/mitigate the R case (e.g. gate the fan-out on `GCSettings.IsServerGC`, or expose a switch), or the R users are the ones who regress.

## Resource bounding

**Core count is the wrong bound for an allocation-heavy phase**
`MaxDegreeOfParallelism = Math.Max(1, _userSettings.MaximumNumberOfCoresToUse)` reuses a setting meant for solver threads to bound memory-hungry model construction. On a 32-core/16 GB machine this is an OOM waiting to happen, especially with server GC's per-core heaps. Consider a memory-aware cap in addition to the core cap.

**Qualification runs will oversubscribe until #3711 lands**
`QualificationRunner` spawns child PKSim.CLI processes in parallel; each child now constructs in parallel with all cores *and* server GC. The `--cores 1` control for children is #3711 / QualificationRunner#189, which is not in this PR, so merging this alone multiplies process count by core count. This ordering dependency should be stated as a merge blocker or handled defensively (default children to 1 core).

**OOM now degrades silently**
`mapSimulationAt` catches `Exception`, so an `OutOfMemoryException` during construction becomes a dropped simulation and a log warning rather than a failed load. The user gets a project that silently lost simulations and can save over the original. The probability of that class of failure rises sharply with this change; consider rethrowing non-recoverable exceptions instead of isolating them.

## `SimulationModelCreator` lock

**The critical section is wider than the stated need**
The PR says only the configuration pipeline is thread-unsafe, but `updateSimulationAfterModelCreation` takes the *same* static lock. That method does `CacheAllChildren<IParameter>` over the whole model and walks all molecules — expensive, allocation-heavy, and (from what it touches: `_containerTask`, `_parameterIdUpdater`, `_entityPathResolver`) not obviously in need of mutual exclusion with the config phase. Sharing one lock means a worker in the post-model phase blocks every other worker's configuration, directly cutting the achievable speedup. Use a separate lock if it needs one at all, and justify why it does.

**Process-wide static lock affects interactive paths**
`CreateModelFor` is also called from `CreateSimulationPresenter`, `ConfigureSimulationPresenter` and `SimulationFactory` (DDI-ratio and PK-parameter simulations). A static lock means any UI-initiated model creation now queues behind snapshot-load workers. Worth confirming no path can hold the lock while waiting on the per-object lazy-load gates introduced in #3709.

**Field placement**: `private static readonly object _configurationLock` is declared between the constructor and `CreateModelFor`, unlike every other field in the class.

## Tests

**The ordering spec is vacuous** — this is the most concrete test problem. `When_converting_a_project_snapshot_with_multiple_simulations_to_project` never stubs `_userSettings.MaximumNumberOfCoresToUse`, so the fake returns `0`, `Math.Max(1, 0)` yields `MaxDegreeOfParallelism = 1`, and `Parallel.ForEachAsync` runs strictly sequentially. The `Task.Delay(100)`/`Delay(50)` staggering never produces out-of-order completion, so the comment "the mappings complete out of order to verify that the simulations are nevertheless added in snapshot order" is false and the assertion would still pass against a completion-order implementation. The PR description's claim that `MaximumNumberOfCoresToUse` is "stubbed so the parallel path actually runs" does not match the committed spec — contrast the sibling fixtures at lines 574 and 606, which *do* stub it. Add `.Returns(3)` (or more) here.

**`When_converting_a_project_snapshot_whose_first_simulation_cannot_be_loaded` adds no observation.** It only reorders `_snapshot.Simulations` and inherits the parent's assertions. Because parallelism is again effectively 1 (same unstubbed setting), nothing in it distinguishes "kept mapping sequentially until one succeeded" from "started the fan-out immediately" — the regression it is meant to pin is not observable. It needs an explicit assertion, e.g. that no mapping began before the first success returned.

**`ClonerSpecs` does not pin the fix.** `A.CallTo(() => _container.Resolve<ICloneManagerForBuildingBlock>()).Returns(_cloneManagerForBuildingBlock)` returns the *same* instance on every call, so the spec passes identically against the old field-based implementation. The behavior that matters — a fresh clone manager per `Clone` call, so concurrent calls do not share `FormulaCache` — is untested. Return a new fake per invocation and assert distinctness. (The production fix itself is sound: `CloneManagerForBuildingBlock` is picked up by Core's scanner, whose default is `LifeStyle.Transient`, so per-call resolution really does yield separate instances — but that lifestyle is now a silent load-bearing assumption of `Cloner` and deserves a comment.)

**Reflection-based invocation of `runParallelSimulations`** in two fixtures is brittle against rename/signature changes and gives no compile-time safety. Making the method internal with `InternalsVisibleTo`, or testing through the public entry point, would be more robust.

## Minor

- **Service locator in `Cloner`**: injecting `IContainer` into a domain service to resolve a dependency per call is a pattern the maintainers have generally avoided; a `Func<ICloneManagerForBuildingBlock>` factory would express "new instance per operation" without handing `Cloner` the whole container. It also adds a container resolution per `Clone` call on a hot path.
- **Lost start-of-load feedback**: removing the `Captions.LoadingSimulation` line means a snapshot with one large simulation now logs "Loading 1 Simulation(s)…" and then nothing until it finishes. Previously the user saw which simulation was being worked on.
- **Progress counter never reaches N when anything fails**: `SimulationsLoadedMessage` counts successes only, so a 60-simulation load with one failure ends at "(59/60)". Intentional, given the follow-up warning, but the two messages read inconsistently.
- **Null as the success sentinel**: `mappedSimulations[index] != null` conflates "mapper returned null" with "mapping threw." A separate success flag would be clearer and would stop a null-returning mapper from silently extending the sequential warm-up phase.
- **Stale PR description**: it still describes mapping "the first simulation alone," which the last commit changed to "sequentially until one succeeds."

rwmcintosh:
Correctness

Partially correct — fixed at the root. The startable warm-up is now deterministic in every host: AwaitCompletion self-starts it (resolving every IStartable itself) when Begin was never called. Sequential-until-first-success stays as defense-in-depth for non-startable lazy services, no longer as the load-bearing guarantee.
Correct — fixed. StartableWarmup is a latch: task created under a lock, never self-clears, concurrent callers block on the same task, and a faulted warm-up resets so the next call retries. New specs cover self-start, single execution, and retry.
Correct — fixed by the same redesign; interface docs state the headless behavior.
Server GC — BatchTool: correct, added. On R regressing: the data says otherwise — desktop A/B on this branch with workstation GC on both sides shows parallel outperforming sequential ~1.7× (99 s vs 173 s on 9 sims; ~12 min vs 20.3 min on 60). The "no better than sequential" cell was the approximate harness run stopped under ~90% system memory. R inherits an improvement at a higher memory footprint; server GC for the R host is Open-Systems-Pharmacology/rSharp#238. No IsServerGC gate — it would deny R users a real gain.

Resource bounding — OOM: correct, both loops rethrow it. Qualification ordering: agreed, now a merge note in the description (#3711 / Open-Systems-Pharmacology/QualificationRunner#189). Memory cap: declined for now — ~8 GB measured at DOP 19 over 60 heavy simulations under server GC, the setting is user-bounded and GUI-validated ≤ core count; worth revisiting with an actual OOM profile.

Lock — correct that it was wider than stated: both phases were serialized empirically during the benchmark-driven fixes. Now split, so finalizing no longer blocks configuring; removing the serialization entirely is the filed follow-up. Ordering verified: the configuration phase takes the #3709 gates inside the lock and no path under a gate calls CreateModelFor — strictly one-directional, no inversion. Field placement fixed.

Tests — ordering spec: correct, fixed (Returns(3)), and the description's claim was wrong on this point; corrected. Corrupted-first spec: correct — replaced with a deterministic handshake (first two fail; the third mapping must not start while the second runs), which fails against an immediate fan-out (verified by mutation). ClonerSpecs: correct, fixed — fresh fake per resolve, one resolution asserted per Clone, lifestyle documented. Reflection: declined — no InternalsVisibleTo exists anywhere in PK-Sim; introducing that convention is a maintainer call.

Minor — start feedback restored at Debug level; success flag replaces the null sentinel; description updated. The per-call resolve mirrors the pattern merged in Open-Systems-Pharmacology/MoBi#2520 (SimulationFactory); a Func factory would need a delegate-factory facility the PK-Sim container doesn't use. The 59/60 count is intentional: it counts successes, and the warning states the loss.

Yuri05:
@rwmcintosh 
## Review No. 2

Reviewed `f195e70` + `65028b8` against `develop`, plus the author's responses.

### Blocking

**1. The self-starting warm-up makes every headless snapshot load depend on *all* startables, and one bad startable now aborts the whole load.**
`AwaitCompletion` (`StartableWarmup.cs:39`) resolves `_container.ResolveAll<IStartable>()` and the task body runs `startables.Each(x => x.Start())` with no per-item guard. The call site (`ProjectMapper.cs:338`) is unguarded, so any startable that throws aborts `allSimulationsFrom` → the entire project load fails.

That set is not limited to the repositories model construction needs. It includes `RemoteTemplateRepository`, whose `DoStart()` deserializes `_configuration.RemoteTemplateSummaryPath` (`RemoteTemplateRepository.cs:118`). That file is provisioned best-effort: `PKSimConfiguration.tryCopyRemoteTemplate()` copies it from the all-users install path and swallows `IOException`, and the all-users folder is only created on Windows. In an R/rSharp or container deployment where the file is missing or truncated, `Deserialize` throws (wrapped by `.Result`) and **a snapshot load that previously worked now fails outright** — for a repository that nothing in the model-construction path uses. Same class of exposure for any other startable with an environment-dependent `DoStart()`.

Also note `Each` aborts at the first failure, so the remaining repositories stay cold even in the failure case — the deterministic-warm-up guarantee the PR is built on does not hold when any single startable fails.

Suggestion: warm only the repositories the construction path needs (or catch + log per startable and let the failing one fall back to lazy init), and don't let warm-up failure fail the load.

**2. The retry path can corrupt repository contents and re-run `SettingsLoader`.**
`OSPSuite.Utility.Collections.StartableRepository<T>.Start()` sets `_initialized = true` only *after* `DoStart()` returns (and the base class is explicitly commented `//not thread safe!`). So a startable that fails midway through `DoStart()` stays `_initialized == false` while having already mutated its own state — e.g. `RemoteTemplateRepository._allTemplates.AddRange(...)`. The new "drop the task so the next call retries" logic (`StartableWarmup.cs:47-58`) then calls `DoStart()` again on the same singleton, appending a second time → duplicated repository entries rather than a clean retry.

Worse in the UI host: `ApplicationStartup.startStartableObject` deliberately excludes `SettingsLoader` from `Begin` and starts it synchronously on the UI thread. `PKSimMainViewPresenter.cs:84` already calls `AwaitCompletion()`; if that call fails, the task is dropped, and the next call (the project load) takes the **self-start** branch, which resolves *every* `IStartable` including `SettingsLoader`. `SettingsLoader.Start()` has no idempotency guard, so `_userSettingsPersistor.Load()` + `_applicationSettingsPersistor.Load()` re-run mid-session on a background thread, discarding in-memory user/application settings changed since startup. The retry silently warms a different, larger set than the host asked for.

**3. A `null` mapping still counts as loaded, so the "only k/N loaded" warning is suppressed.**
`mapSimulationAt` increments `numberOfSimulationsLoaded` and logs `SimulationsLoadedMessage` *before* checking the result (`ProjectMapper.cs:352-355`), then returns `mappedSimulations[index] != null`. If the mapper returns `null`, the counter and the log say "loaded", but the simulation is skipped at `ProjectMapper.cs:383-385`, and because `numberOfSimulationsLoaded == snapshots.Length` the warning at `:388` never fires. Simulations disappear from the project with an affirmative log line. The counter should be incremented only for results that are actually added.

**4. The new latch guards the task, not the startables — the underlying warm-up race is untouched.**
`Begin` still starts repository `Start()` on a background thread while the UI thread continues building the main window and resolving presenters that hit the same repositories lazily via `All()` → `Start()`. With `_initialized` non-volatile and unsynchronized, both threads can see `false` and run `DoStart()` concurrently on the same singleton (duplicated/interleaved cache fill). Making `_warmupTask` thread-safe does not address this; the fix belongs in how the repositories are started (e.g. warm them before anything else can touch them, or guard `Start()`).

### Medium

**5. The OOM rethrow does not surface as `OutOfMemoryException` from the parallel loops.**
`Parallel.ForEachAsync` faults its task with an `AggregateException`, so the sequential warm-up loop propagates a bare `OutOfMemoryException` while the fan-out (`ProjectMapper.cs:379`) and `runParallelSimulations` (`:182`) propagate a wrapped one. Any caller that filters on `OutOfMemoryException` sees inconsistent behaviour. For the same reason, the filter `when (ex is not OutOfMemoryException)` misses an OOM that any intermediate layer has already wrapped (e.g. `AggregateException` from a `.Result`, `TargetInvocationException`) — those are still swallowed as ordinary per-simulation failures, which is the silent degradation the change is meant to prevent.

**6. OOM during the run phase produces misleading log noise and discards a fully loaded project.**
When one worker in `runParallelSimulations` rethrows, `Parallel.ForEachAsync` cancels the shared `ct`; every in-flight `RunSimulation` throws `OperationCanceledException`, which *is* caught by the filter, so each logs `AddException` + a decrement — the log shows N spurious per-simulation failures before the real OOM surfaces. `AllSimulationsFinishedMessage` is skipped and the exception propagates out of `MapToModel`, so a project whose simulations all loaded successfully is thrown away because the optional run phase hit OOM. Consider failing the run phase without failing the load.

**7. The lock split is asserted, not justified.**
Both phases reach the same singletons: `updateSimulationAfterModelCreation` uses `_containerTask` and `_entityPathResolver`, and `_simulationConfigurationTask.CreateFor` sits on top of the same ORM/mapper/path-resolution layer. The previous single lock guaranteed the two phases never overlapped; the split now allows it, and the benchmark evidence quoted in the PR was gathered with the combined lock. Conversely, if `updateSimulationAfterModelCreation` only touches simulation-local state (its own `simulation.Individual` clone and `simulation.Model`), then no process-wide lock is needed there at all. Either way the comment should state the invariant being protected. Note `_finalizationLock` is still `static`, so a background snapshot load continues to serialize against interactive simulation creation.

**8. Container resolution now happens on worker threads, contradicting the comment the PR itself added.**
`ApplicationStartup.startStartableObject` documents "Castle Windsor resolution stays single-threaded" as the reason startables are resolved on the UI thread. Yet `Cloner.Clone` resolves `ICloneManagerForBuildingBlock` per call from parallel workers, and the warm-up self-start calls `ResolveAll<IStartable>()` from whichever thread happens to run the project load. Please reconcile: either concurrent resolution is safe (then drop/adjust the ApplicationStartup comment) or it isn't (then the per-clone resolve needs to move off the worker threads).

### Minor

**9.** Hardcoded English log strings at `ProjectMapper.cs:350` (`"Constructing simulation '…'"`), `:378` and `:181` (`"Constructing/Running simulations with up to … core(s)"`), while every neighbouring message goes through `PKSimConstants`. The PR already edits `PKSimConstants.cs`.

**10.** The null sentinel is not actually removed — `return mappedSimulations[index] != null` just re-derives success from the array slot; a `null` return is still indistinguishable from a failure to the caller (and drives issue 3).

**11.** `Begin` now returns silently if a task already exists, so the host-provided list is discarded whenever a self-start happened first; combined with issue 2 this means the effective warm-up set depends on call order and prior failures. Worth at least documenting on `Begin`.

**12.** `StartableWarmupSpecs` does not cover the property the latch exists for: concurrent `AwaitCompletion` callers sharing one task (all three specs call sequentially). `When_the_warmup_fails` also packs two state-changing actions plus the assertion into one `[Observation]`, so the retry action runs inside the observation rather than in `Because`.

**13.** `When_converting_a_project_snapshot_whose_leading_simulations_cannot_be_loaded` blocks the test thread on a fixed `Wait(500 ms)` negative assertion. It can only ever false-pass (a slow CI agent hides a fan-out regression) and adds a guaranteed half-second to every run.

**14.** Still open from Review No. 1: `PKSim.R` remains a library with no server-GC opt-in, so the R host — the most common headless consumer of this path — keeps workstation GC while the property was added to `PKSim`, `PKSim.CLI` and now `PKSim.BatchTool`. Tracking this in rSharp#238 is reasonable, but the PR then ships a parallel allocation-heavy path whose GC configuration is met in three hosts out of four; that asymmetry deserves a note in the code or the issue link next to the property, not only in the PR discussion.

rwmcintosh:
Blocking

1. Partially taken — the warm-up stays strict by design. A startable that cannot start is a broken installation, and the UI already treats it that way today (AwaitCompletion rethrows before the first project load). What changed: the failure now names the startable, and your concrete example is fixed at its owner — RemoteTemplateRepository's summary file is provisioned best-effort by its own design (tryCopyRemoteTemplate deliberately swallows IOException), so a missing file is an expected state and its DoStart now returns empty instead of throwing. Startables with genuinely mandatory resources keep failing loudly.

2. Correct — the retry is deleted. A failed warm-up keeps its faulted task: every call rethrows the same exception without running Start() again on a repository that already mutated its state. That also closes the SettingsLoader path — a completed-or-faulted Begin task means the self-start branch is unreachable in the UI. The partial-DoStart-then-lazy-retry corruption that remains is pre-existing OSPSuite.Utility behavior — filed as Open-Systems-Pharmacology/OSPSuite.Utility#60 together with point 4.

3. Correct — fixed. A null mapping no longer counts as loaded; the warning fires. Covered by a new fixture. (Also resolves 10.)

4. Correct in substance, but pre-existing: Begin-on-background and the presenter's AwaitCompletion are on develop; this PR only changed StartableWarmup's internals. The fix belongs in StartableRepository.Start() itself — Open-Systems-Pharmacology/OSPSuite.Utility#60.

Medium

5. Correct — fixed. A recursive isOutOfMemory check (unwrapping AggregateException and inner exceptions) replaces the naive filter in both loops, pinned by a fixture that fails the load on a wrapped OOM.

6. Correct — fixed as suggested. Run-phase OOM aborts the run phase with a named error and keeps the loaded project; in-flight cancellations are no longer logged as failures.

7. Correct — reverted to the single lock, which is the configuration the benchmarks ran with. The comment now states the invariant: both phases sit on the same ORM/mapper/path-resolution layer whose thread-safety is unverified; narrowing it is the config-pipeline follow-up with its own verification.

8. Reconciled: Windsor's concurrent Resolve is thread-safe, and the merged MoBi sibling already resolves per call from workers — the ApplicationStartup parenthetical was wrong and is corrected.

Minor

9. Correct — fixed (all three via PKSimConstants).

10. Resolved via 3.

11. Documented on Begin.

12. Correct — fixed: a concurrent-callers fixture (both block on the same warm-up, Start exactly once) and a sticky-failure fixture with the actions in Because replace the retry fixture.

13. Partially taken. The observation is now a flag recorded inside the third mapping ("did it start before the second completed"), so a detection is exact rather than inferred from a timeout — but the bounded window stays, because a negative concurrency assertion needs one: without it, releasing the second mapping races the (hypothetically broken) fan-out's scheduling, which we confirmed by mutation. Correct code passes with zero flake risk; the window only bounds mutation-detection sensitivity, and the comment says so.

14. Two updates. rSharp main already runs server GC — rSharp#238 is closed and RSharp.runtimeconfig.json carries "System.GC.Server": true — so all four hosts of this path are covered. And even an R installation on an older rSharp still gains from this PR: in the measured A/B with workstation GC on both sides, parallel construction outperformed sequential ~1.7×.

Yuri05:
@rwmcintosh 
## Review No. 3

Reviewed `0ff2ed1` against `65028b8`, and re-checked the base: `StartableWarmup`, the background `Begin` and `PKSimMainViewPresenter.Run`'s `AwaitCompletion` are indeed already on `develop`, so points 4 (and the pre-existing part of 2) from Review No. 2 are correctly classified as inherited. rSharp#238 is closed via rSharp#239 (26 Aug 2026), so the R server-GC gap is genuinely resolved upstream.

### Blocking

**1. A single failing startable now permanently disables snapshot loading for the whole process.**
This is a change against `develop`, not the status quo. On `develop`, `AwaitCompletion` is one-shot — it nulls `_warmupTask` before awaiting — so a startable failure surfaces once in `PKSimMainViewPresenter.Run()` and every later load simply proceeds with lazy initialization. With `0ff2ed1` the faulted task is kept forever, so every subsequent `ProjectMapper.allSimulationsFrom` rethrows the same exception for the life of the process.

The premise that this only affects broken installations does not hold. The code base itself documents transient I/O on exactly these files — `PKSimConfiguration.tryCopyRemoteTemplate` swallows `IOException` with the comment "the file is probably used by the process, happens in parallel unit testing" — and a DB or dimension file briefly locked by a concurrent PK-Sim instance or a qualification child process produces the same fatal, sticky state. In a long-lived R session there is then no recovery short of restarting R, for an operation that would have succeeded lazily on `develop`.

Please scope the failure to the startable that failed (name it, log it, let it be lazily retried) or expose a reset path. Two secondary points in the same area:
- `startables.Each(start)` still aborts at the first failure, so every startable after it stays cold — the deterministic-warm-up guarantee the PR rests on silently does not hold in the failure case.
- `IStartableWarmup.AwaitCompletion`'s doc still says "rethrowing any exception it raised", but the implementation now rethrows a `PKSimException` wrapper; and in the UI the throw happens inside `Run()` *before* `_projectTask.Run(StartOptions)` and `_postLaunchChecker`, so the app comes up half-initialized rather than reporting an unusable installation.

**2. The eager warm-up is now applied to headless loads that cannot benefit from it.**
`AwaitCompletion()` is called unconditionally in `allSimulationsFrom`, before the simulation count or the degree of parallelism is known, and in CLI/R it takes the self-start branch that resolves and starts *every* `IStartable`. On `develop` those hosts warmed nothing. So `loadSimulationsFromSnapshot()` in R with a single simulation — and any load with `MaximumNumberOfCoresToUse == 1` — now pays the full repository initialization up front, where `develop` initialized lazily only what the mapping actually touched. The PR states "single-simulation snapshots skip the fan-out", but they do not skip the warm-up, and no measurement is given for that case. Since the warm-up exists solely to make concurrent first use safe, gate it on `snapshots.Length > 1` and a degree of parallelism above 1.

### Medium

**3. The OOM containment stops one step too early.**
After the new catch keeps the project, execution continues into `addAnalysesToSimulations(..., runSimulations: projectContext.RunSimulations)`, which is unguarded and maps analyses for simulations whose runs were cancelled or never executed, with a `SimulationContext` that still says results are present. Any exception there — including a second OOM, since the process is still memory-starved and this step allocates charts for every simulation — propagates out of `MapToModel` and discards exactly the project the catch was added to save. Either skip/flag the analyses step when the run phase aborted, or extend the guard over it.

**4. User-initiated cancellation became silent.**
The new `catch (OperationCanceledException)` only decrements the counter, with no log entry, and `runParallelSimulations` still ends with `AllSimulationsFinishedMessage()`. A cancelled run phase therefore reports "All Simulations Finished Running." with nothing in the log showing that anything was cancelled. Only the OOM-induced internal cancellation should be silent; distinguish it (e.g. by an internal abort flag set before rethrow) and keep logging an externally cancelled run.

**5. `isOutOfMemory` recursion is unbounded.**
`default: return isOutOfMemory(e.InnerException)` recurses without a depth limit, and the `AggregateException` branch fans out per inner exception. Exception chains coming from interop, serialization round-trips or third-party wrappers can be self-referencing or cyclic, in which case this stack-overflows inside an exception filter — i.e. in the one place where failure is unrecoverable. An iterative walk with a depth cap avoids the risk entirely.

### Minor

**6. The PR description is stale against `0ff2ed1` again.** It still states "A failed warm-up is retried on the next call", "StartableWarmup specs cover … retry after a failed warm-up" (the retry is deleted in this very commit), and "the post-creation update is serialized on a separate lock so a worker finalizing does not block a worker configuring" (reverted to one `_configurationLock`). The run-phase OOM containment — a user-visible behaviour change — is not described at all.

**7. `When_converting_a_project_snapshot_where_a_simulation_runs_out_of_memory` under-asserts.** It only checks `ShouldThrowAn<AggregateException>()`, which does not pin that the OOM is in the chain, nor that the per-simulation `CannotLoadSimulation` error was *not* logged — that omission is the actual "no silent degradation" property. `_logger` is already verified in the neighbouring fixture, so both assertions are cheap.

**8. `When_the_warmup_fails` depends on exception instance identity.** `ReferenceEquals(_firstFailure, _secondFailure)` holds only because `TaskAwaiter.GetResult` rethrows via `ExceptionDispatchInfo` — which also means that single instance accumulates a rethrow stack-trace entry on every `AwaitCompletion` call, and with the sticky semantics of item 1 that is for the life of the process. Asserting type and message pins the intended behaviour without relying on identity.

**9. `isOutOfMemory` is a private static on `ProjectMapper`** while the rule it encodes ("never swallow an OOM") applies to every broad `catch (Exception)` on the load path (`SimulationMapper`, the building-block mappers). Moving it somewhere shared would let the rule actually be applied consistently in the follow-ups.

**10. `RemoteTemplateRepository.DoStart`'s early return leaves `Version` null and `_allTemplates` empty with no trace.** Benign today — `PostLaunchChecker.checkForNewTemplateVersionAsync` swallows everything and the null most likely triggers a summary re-download that self-heals — but "no remote templates configured" and "summary file missing" are now indistinguishable in the log. A debug/warning line makes the state diagnosable.

**11. Server GC for `PKSim.R` is now a release-coordination item, not a resolved one.** rSharp#239 lands the setting upstream; PKSim.R only inherits it once the shipped rSharp build containing it is the one the PK-Sim R package uses. Worth stating that dependency next to the `ServerGarbageCollection` properties so it is not lost.

rwmcintosh:
1. Taken, with the design adjusted rather than reverted. A failure is now scoped to the startable that failed: it is logged by name, the remaining startables still start (the Each(start) abort is gone), and only the failed ones are retried on the next AwaitCompletion, so a transient lock self-heals and nothing is ever disabled process-wide. AwaitCompletion returns whether everything warmed instead of throwing, so the UI starts fully initialized and the docs match the behavior. The guarantee is kept where it matters: a load that finds a cold repository degrades to sequential construction — lazy initialization is safe single-threaded — instead of failing or fanning out unsafely.

2. Correct — fixed. The warm-up is gated on snapshots.Length > 1 && MaxDegreeOfParallelism > 1, short-circuited before AwaitCompletion, so single-simulation and single-core loads initialize lazily exactly as on develop. Pinned by fixtures: a single-simulation load never awaits the warm-up, and an incomplete warm-up loads everything on one core.

3. Correct — fixed. An aborted run phase flags the load and the analyses then map as in a no-run load.

4. Correct — fixed as suggested: an internal abort flag is set before the OOM rethrows; only those collateral cancellations stay quiet, any other cancelled run is logged.

5. Correct — fixed: iterative walk with an explicit stack and a visit cap, pinned by a cyclic-chain spec.

6. Correct — the description is rewritten against the new commit.

7. Correct — fixed: the fixture now asserts the OOM is in the chain and that CannotLoadSimulation was not logged.

8. Resolved by the redesign in 1: the sticky-failure fixture is replaced by a fail-once fixture — incomplete result reported, remaining startables still started, failure logged by name, and only the failed startable retried, with the healthy one started exactly once.

9. Correct — moved to ExceptionExtensions.IsOutOfMemory() in PKSim.Core with its own specs, so the follow-ups can apply it to the other broad catches on the load path.

10. Correct — fixed: DoStart logs a debug line naming the missing summary path.

11. There is nothing in this repo to attach that note to: ServerGarbageCollection only affects executables, and PKSim.R is a library — the R host's GC mode comes from rSharp's own runtimeconfig, already on rSharp main via rSharp#239. Which rSharp build the R package bundles is a release concern of that package, not of this PR.

Yuri05:
@rwmcintosh 
## Review No. 4

Reviewed `fe7ccee` against `0ff2ed1`. The description now matches the code, and the previous blocking items (sticky poisoning, unconditional headless warm-up, OOM containment scope, cancellation logging, shared `IsOutOfMemory`) are addressed. New findings below.

### Blocking

**1. `StartableWarmup.start` now swallows `OutOfMemoryException`, and the fallback silently corrupts the repository that failed.**
`catch (Exception e)` catches everything, including OOM: the failure is logged as `"'X' failed to start and will be retried on its next use"`, `AwaitCompletion` returns `false`, and the load continues sequentially in a memory-exhausted process. That is precisely the silent degradation the new `ExceptionExtensions.IsOutOfMemory` exists to prevent — the same commit that codifies "an out-of-memory failure must fail the operation rather than degrade it silently" introduces a broad catch that violates it.

The second-order effect is worse than a slow load. A repository whose `DoStart()` died halfway has already mutated its own state while `StartableRepository._initialized` is still `false` (the base class sets it only after `DoStart()` returns). The degraded sequential load then touches that repository, lazy `Start()` runs `DoStart()` a second time, and appending repositories (`RemoteTemplateRepository._allTemplates.AddRange`, `Version`) end up with duplicated content — inside the same load, with nothing but a Debug/Error line to indicate it. Please add `when (!e.IsOutOfMemory())` so an OOM faults the warm-up task, and see item 6 for how the faulted task must then be handled.

**2. The reinstated retry re-opens the partial-`DoStart` corruption that `0ff2ed1` removed it to avoid.**
This is the same mechanism as above, now on the intended path: `AwaitCompletion` re-runs `Start()` on exactly the startables that failed, and for any repository that appends to its own fields before completing, the retry duplicates what the failed attempt already wrote. In `0ff2ed1` this was the stated reason for deleting the retry, and it was filed upstream as OSPSuite.Utility#60 — but the retry is now automatic on every subsequent multi-simulation load, so the corruption path is reachable in normal operation rather than once.

The transient-lock argument for retrying is sound, but retry is only correct if `DoStart()` is idempotent. Either make the retried implementations reset first (for `RemoteTemplateRepository` that is `_allTemplates.Clear(); Version = null;` at the top of `DoStart` — a local one-liner), or gate the retry on the upstream fix. Retrying a non-atomic `DoStart` while citing OSPSuite.Utility#60 as the reason not to is inconsistent.

**3. `IsOutOfMemory`'s cap bounds visited nodes, not depth, so a wide `AggregateException` can hide the OOM.**
`MAXIMUM_NUMBER_OF_EXCEPTIONS_TO_INSPECT = 20` counts every node popped, and both `Parallel.ForEachAsync` loops run at `MaximumNumberOfCoresToUse`, which defaults to `Environment.ProcessorCount - 1` in `UserSettings` and `CLIUserSettings` — 19 on the benchmark machine, 31/63 on larger ones. A fan-out where many workers fail (the realistic shape of an OOM: one worker exhausts memory, siblings fail too) produces an `AggregateException` with more inner exceptions than the cap, and because the walk is depth-first via `Stack`, a single deep early branch can consume the budget on its own. `IsOutOfMemory` then returns `false` and the OOM is logged as an ordinary per-simulation failure — the silent degradation the helper exists to prevent, in exactly the high-core scenario the PR targets. Bound the traversal *depth* (or track visited instances) instead of the node count, or at minimum set the cap well above the maximum possible degree of parallelism.

### Medium

**4. The parallel run phase is not gated on the warm-up result, only construction is.**
`allSimulationsFrom` degrades to `MaxDegreeOfParallelism = 1` when `AwaitCompletion()` returns `false`, with the stated rationale that "a cold repository is not safe to initialize from parallel workers". `runParallelSimulations` then calls `parallelOptions()` fresh and fans out at full width regardless, and simulation runs also touch lazily initialized services. Either the rationale applies to both loops or it applies to neither; as written, the mitigation covers the loop this PR added and skips the pre-existing one it is now knowingly running alongside a known-cold repository.

**5. A genuinely broken installation no longer produces a diagnosable failure.**
`AwaitCompletion` cannot throw anymore, and the only consequence of a failed mandatory repository is DOP 1. If the PK-Sim DB or the dimensions file is unreadable, the load proceeds sequentially and emits one `CannotLoadSimulation` error per simulation — 60 for the benchmark project — while the actual cause is a single earlier error line from the warm-up. The fallback itself is visible only at Debug (`ConstructingSimulationsWithCoresMessage(1)`), so a user reporting "loading is slow again" leaves no usable signal above Debug either. Log a warning at the load site naming the reason for the sequential fallback, and consider distinguishing mandatory from optional startables so the mandatory case can still fail loudly.

**6. `_warmupTask.IsCompleted` should be `IsCompletedSuccessfully`.**
`IsCompleted` is also `true` for a faulted or canceled task, so `_warmupTask.Result.Any()` in the retry condition throws `AggregateException` — while holding `_lock`, and out of a method whose documented contract is now "returns `true` when every startable started". This is latent today only because `start` swallows everything; applying the OOM filter from item 1 makes it live, and then the faulted task is never replaced, so every later call rethrows the same fault with no retry — the poisoning behaviour this commit set out to remove, reintroduced through a different door. Decide explicitly what a faulted warm-up task means and handle it (drop it and let the next call re-run, or propagate once and document it).

### Minor

**7.** `Interlocked.CompareExchange(ref abortedByOutOfMemory, 0, 0) == 0` is a read written as a read-modify-write. `Volatile.Read(ref abortedByOutOfMemory) == 0` says the same thing without implying a mutation.

**8.** A user-cancelled run phase now logs each cancellation and then still ends with `AllSimulationsFinishedMessage()` — "All Simulations Finished Running." after N cancellations. Skip the final message when anything was cancelled.

**9.** `should_terminate_on_a_cyclic_inner_exception_chain` reflects on the private `_innerException` field of `System.Exception`. That name is a runtime implementation detail, not contract: on a runtime that renames it, `GetField` returns `null` and the spec fails with a `NullReferenceException` rather than a readable message. Guard the lookup (and note the test currently passes because of the visit cap, not because cycles are detected — if the cap is widened per item 3, the cycle needs its own protection).

**10.** `parallelOptions()` is now created before the sequential phase and then mutated (`options.MaxDegreeOfParallelism = 1`). Computing the degree of parallelism first and constructing the options once reads better and avoids a mutable options object shared across two phases.

rwmcintosh:
1. Agreed and fixed.

2. Agreed — the retry is removed rather than made safe. We audited the reset-first option: there are 39 DoStart implementations, appending is the norm (SpeciesRepository, OntogenyRepository, CalculationMethodCategoryRepository, …), and DimensionRepository even mutates the shared dimension factory, which cannot be reset locally. So the warm-up never re-runs Start(): a failed startable is logged by name and stays cold, and its only retry is the pre-existing lazy initialization at first use — develop's semantics, with making that retry safe tracked in OSPSuite.Utility#60. Loads stay sequential after a real failure, which point 5's warning makes visible. We did take the RemoteTemplateRepository one-liner: DoStart resets before filling.

3. Agreed and fixed (visited set instead of the cap).

4. Agreed and fixed.

5. Partially taken: the fallback now logs a warning pointing at the earlier named error. Distinguishing mandatory from optional startables we'd leave to a follow-up — the failure is already an error with the full exception, and classifying 39 startables belongs with the OSPSuite.Utility#60 work.

6. Resolved structurally: with the retry gone, the .Result-under-lock branch no longer exists — the task is created once and never replaced. A faulted task (only an OOM can fault it now) is kept and rethrows on every call, documented on the interface: the "propagate and document it" option, applied only to the failure mode this PR says must never degrade silently.

7. Agreed and fixed.

8. Agreed and fixed.

9. Agreed and fixed.

10. Agreed and fixed.

Yuri05 (approved):

PR #690, opened 2018-05-07
Fixes #684 Issue #648 still will not load

msevestre:

PR #3716, opened 2026-08-26
Fixes #3709 Make the building-block lazy-load chain thread-safe

### LINKED ISSUE #3709: Building-block lazy-load/registration chain is not thread-safe and is already exercised concurrently

rwmcintosh:
Fixes #3709

> Requires OSPSuite.Core **13.0.165** (contains Open-Systems-Pharmacology/OSPSuite.Core#2959 — thread-safe `WithIdRepository` — and Open-Systems-Pharmacology/OSPSuite.Core#2965 — volatile `IsLoaded` publication) — already referenced by this branch.

### What changes

- `LazyLoadTask` loads each object under its own gate: a private lock object per loaded instance, held in a `ConditionalWeakTable` so a gate never outlives the object it guards. The already-loaded fast path stays lock-free; the `IsLoaded` check is re-tested inside the gate, so one thread loads the content while others wait — and loads of unrelated objects do not block each other.
- `IsLoaded` is published through a volatile field (`PKSimBuildingBlock`, the comparisons, and Core's `ParameterIdentification`/`SensitivityAnalysis` via Core 13.0.165), so a thread that takes the fast path also sees the writes that loaded the object. `Simulation` and `Population` flag their owned children before the volatile parent write, so a reader that sees the parent loaded also sees its children loaded.
- `LoadResults` takes the same single gate for the load and the results, so the two steps are atomic per simulation.
- `DefaultIndividualRetriever` caches per population through `Lazy<Individual>` in a `ConcurrentDictionary`, creating the default individual once; a creation that throws is evicted so the next call retries instead of rethrowing a cached exception.

### Why

`LoadSimulationFromSnapshotPresenter` already maps snapshot simulations concurrently (`Task.WhenAll`), and that mapping resolves template building blocks through `IExecutionContext.Load` → `LazyLoadTask`. The load is a check-then-act around a database read plus XML deserialization: two threads can both pass the `IsLoaded` check, deserialize into the same shared building block, and bulk-register its object graph into the global `IWithIdRepository` at the same time. That is the most likely mechanism behind #648/#684, which led #690 to serialize the snapshot mapping loop in 2018.

This is also a prerequisite for #3710 (parallel model construction during snapshot load).

### Tests

The concurrency specs are handshake-based (events, no sleeps for correctness) and each guard was verified by mutation — reverting it makes the spec fail:

- loading the same object from two threads blocks the second thread and loads and registers the content exactly once
- loading two unrelated objects does not block across gates
- retrieving the default individual for the same population from two threads creates it once and hands both callers the same instance
- retrieving for a different population is not blocked by a creation in progress
- a failed creation is not cached: the next call retries
- an integration spec pins the retriever to a single container registration, which the per-instance cache relies on

### Follow-ups filed

- #3718 — `DefaultIndividualRetriever` hands out a shared mutable `Individual`; ownership to be decided separately
- #3720 — defer nested lazy loads so a thread holds at most one gate, removing the acyclic-references invariant this PR documents

## Summary by CodeRabbit

- **Improvements**
  - Improved reliability when creating and retrieving default individuals concurrently.
  - Improved consistency when multiple operations load the same item simultaneously.

- **Maintenance**
  - Updated supporting platform components to version 13.0.164 across the application and test suites.

- **Tests**
  - Added coverage for concurrent default-individual retrieval and simultaneous item loading to help prevent duplicate processing and inconsistent results.

### DIFF
--- a/src/PKSim.Assets/PKSim.Assets.csproj
+++ b/src/PKSim.Assets/PKSim.Assets.csproj
@@ -22,9 +22,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.164" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.165" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.165" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.165" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PKSim.Assets/PKSim.Assets.csproj
+++ b/src/PKSim.Assets/PKSim.Assets.csproj
@@ -22,9 +22,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.161" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.161" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.161" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PKSim.BatchTool/PKSim.BatchTool.csproj
+++ b/src/PKSim.BatchTool/PKSim.BatchTool.csproj
@@ -40,8 +40,8 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="DevExpress.Win.Design" Version="25.2.7" Condition="'$(ExcludeDesigner)' != 'true'" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.161" />
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.161" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" />
     <PackageReference Include="OSPSuite.FuncParser" Version="5.0.0.3" />
     <PackageReference Include="OSPSuite.SimModel" Version="5.0.0.80" />
     <PackageReference Include="OSPSuite.SimModelSolver_CVODES" Version="5.0.0.53" />
--- a/src/PKSim.BatchTool/PKSim.BatchTool.csproj
+++ b/src/PKSim.BatchTool/PKSim.BatchTool.csproj
@@ -40,8 +40,8 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="DevExpress.Win.Design" Version="25.2.7" Condition="'$(ExcludeDesigner)' != 'true'" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.165" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.165" />
     <PackageReference Include="OSPSuite.FuncParser" Version="5.0.0.3" />
     <PackageReference Include="OSPSuite.SimModel" Version="5.0.0.80" />
     <PackageReference Include="OSPSuite.SimModelSolver_CVODES" Version="5.0.0.53" />
--- a/src/PKSim.CLI.Core/PKSim.CLI.Core.csproj
+++ b/src/PKSim.CLI.Core/PKSim.CLI.Core.csproj
@@ -23,10 +23,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.161" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" />
     <PackageReference Include="OSPSuite.Utility" Version="5.0.0.11" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.161" />
-    <PackageReference Include="OSPSuite.CLI.Core" Version="13.0.161" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.CLI.Core" Version="13.0.164" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PKSim.CLI.Core/PKSim.CLI.Core.csproj
+++ b/src/PKSim.CLI.Core/PKSim.CLI.Core.csproj
@@ -23,10 +23,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.165" />
     <PackageReference Include="OSPSuite.Utility" Version="5.0.0.11" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
-    <PackageReference Include="OSPSuite.CLI.Core" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.165" />
+    <PackageReference Include="OSPSuite.CLI.Core" Version="13.0.165" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/PKSim.CLI/PKSim.CLI.csproj
+++ b/src/PKSim.CLI/PKSim.CLI.csproj
@@ -47,9 +47,9 @@
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.9.1" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="9.0.2" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
-    <PackageReference Include="OSPSuite.Presentation" Version="13.0.164" />
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.165" />
+    <PackageReference Include="OSPSuite.Presentation" Version="13.0.165" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.165" />
     <PackageReference Include="OSPSuite.Utility" Version="5.0.0.11" />
     <PackageReference Include="OSPSuite.FuncParser" Version="5.0.0.3" />
     <PackageReference Include="OSPSuite.SimModel" Version="5.0.0.80" />
--- a/src/PKSim.CLI/PKSim.CLI.csproj
+++ b/src/PKSim.CLI/PKSim.CLI.csproj
@@ -47,9 +47,9 @@
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.9.1" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="9.0.2" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.161" />
-    <PackageReference Include="OSPSuite.Presentation" Version="13.0.161" />
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.161" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Presentation" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" />
     <PackageReference Include="OSPSuite.Utility" Version="5.0.0.11" />
     <PackageReference Include="OSPSuite.FuncParser" Version="5.0.0.3" />
     <PackageReference Include="OSPSuite.SimModel" Version="5.0.0.80" />
--- a/src/PKSim.Core/Chart/IndividualSimulationComparison.cs
+++ b/src/PKSim.Core/Chart/IndividualSimulationComparison.cs
@@ -9,7 +9,15 @@ namespace PKSim.Core.Chart
    {
       private readonly ICache<string, IndividualSimulation> _allSimulations;
 
-      public bool IsLoaded { get; set; }
+      //see ILazyLoadable: lazy loading reads this flag without holding the lock that guards the load, so it is
+      //published with a volatile field to make the writes that loaded the object visible with it
+      private volatile bool _isLoaded;
+
+      public bool IsLoaded
+      {
+         get => _isLoaded;
+         set => _isLoaded = value;
+      }
 
       public IndividualSimulationComparison()
       {
--- a/src/PKSim.Core/Model/PKSimBuildingBlock.cs
+++ b/src/PKSim.Core/Model/PKSimBuildingBlock.cs
@@ -31,7 +31,15 @@ namespace PKSim.Core.Model
 
    public abstract class PKSimBuildingBlock : Entity, IPKSimBuildingBlock
    {
-      public virtual bool IsLoaded { get; set; }
+      //see ILazyLoadable: lazy loading reads this flag without holding the lock that guards the load, so it is
+      //published with a volatile field to make the writes that loaded the object visible with it
+      private volatile bool _isLoaded;
+
+      public virtual bool IsLoaded
+      {
+         get => _isLoaded;
+         set => _isLoaded = value;
+      }
       public virtual bool HasChanged { get; set; }
       public virtual string ExtendedDescription { get; set; }
       public virtual CreationMetaData Creation { get; set; }
--- a/src/PKSim.Core/Model/Population.cs
+++ b/src/PKSim.Core/Model/Population.cs
@@ -140,9 +140,11 @@ namespace PKSim.Core.Model
       {
          set
          {
-            base.IsLoaded = value;
+            //the individual is flagged first so that a reader seeing the population loaded also sees its
+            //individual loaded (the base setter is the volatile publication)
             if (FirstIndividual != null)
                FirstIndividual.IsLoaded = value;
+            base.IsLoaded = value;
          }
       }
 
--- a/src/PKSim.Core/Model/Population.cs
+++ b/src/PKSim.Core/Model/Population.cs
@@ -140,8 +140,8 @@ namespace PKSim.Core.Model
       {
          set
          {
-            //the individual is flagged first so that a reader seeing the population loaded also sees its
-            //individual loaded (the base setter is the volatile publication)
+            //the individual is flagged first so that a load never publishes the population as loaded
+            //before its individual (the base setter is the volatile publication)
             if (FirstIndividual != null)
                FirstIndividual.IsLoaded = value;
             base.IsLoaded = value;
--- a/src/PKSim.Core/Model/PopulationSimulationComparison.cs
+++ b/src/PKSim.Core/Model/PopulationSimulationComparison.cs
@@ -15,7 +15,15 @@ namespace PKSim.Core.Model
    {
       private readonly ICache<string, PopulationSimulation> _allSimulations = new Cache<string, PopulationSimulation>(x => x.Id);
       private readonly IList<ISimulationAnalysis> _allSimulationAnalyses = new List<ISimulationAnalysis>();
-      public bool IsLoaded { get; set; }
+      //see ILazyLoadable: lazy loading reads this flag without holding the lock that guards the load, so it is
+      //published with a volatile field to make the writes that loaded the object visible with it
+      private volatile bool _isLoaded;
+
+      public bool IsLoaded
+      {
+         get => _isLoaded;
+         set => _isLoaded = value;
+      }
       public virtual PopulationSimulation ReferenceSimulation { get; set; }
       public virtual GroupingItem ReferenceGroupingItem { get; set; }
       public virtual ParameterDistributionSettingsCache SelectedDistributions { get; }
--- a/src/PKSim.Core/Model/Simulation.cs
+++ b/src/PKSim.Core/Model/Simulation.cs
@@ -618,8 +618,8 @@ namespace PKSim.Core.Model
       {
          set
          {
-            //the used building blocks are flagged first so that a reader seeing the simulation loaded also
-            //sees its building blocks loaded (the base setter is the volatile publication)
+            //the used building blocks are flagged first so that a load never publishes the simulation as
+            //loaded before its building blocks (the base setter is the volatile publication)
             UsedBuildingBlocks.Each(bb => bb.IsLoaded = value);
             base.IsLoaded = value;
          }
--- a/src/PKSim.Core/Model/Simulation.cs
+++ b/src/PKSim.Core/Model/Simulation.cs
@@ -618,8 +618,10 @@ namespace PKSim.Core.Model
       {
          set
          {
-            base.IsLoaded = value;
+            //the used building blocks are flagged first so that a reader seeing the simulation loaded also
+            //sees its building blocks loaded (the base setter is the volatile publication)
             UsedBuildingBlocks.Each(bb => bb.IsLoaded = value);
+            base.IsLoaded = value;
          }
       }
 
--- a/src/PKSim.Core/PKSim.Core.csproj
+++ b/src/PKSim.Core/PKSim.Core.csproj
@@ -26,10 +26,10 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageReference Include="OSPSuite.Utility" Version="5.0.0.11" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.164" />
-    <PackageReference Include="OSPSuite.Infrastructure.Import" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.165" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.165" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.165" />
+    <PackageReference Include="OSPSuite.Infrastructure.Import" Version="13.0.165" />
     <PackageReference Include="System.ComponentModel.Annotations" Version="5.0.0" />
   </ItemGroup>
 
--- a/src/PKSim.Core/PKSim.Core.csproj
+++ b/src/PKSim.Core/PKSim.Core.csproj
@@ -26,10 +26,10 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageReference Include="OSPSuite.Utility" Version="5.0.0.11" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.161" />
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.161" />
-    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.161" />
-    <PackageReference Include="OSPSuite.Infrastructure.Import" Version="13.0.161" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Infrastructure.Import" Version="13.0.164" />
     <PackageReference Include="System.ComponentModel.Annotations" Version="5.0.0" />
   </ItemGroup>
 
--- a/src/PKSim.Core/Services/DefaultIndividualRetriever.cs
+++ b/src/PKSim.Core/Services/DefaultIndividualRetriever.cs
@@ -14,6 +14,7 @@ namespace PKSim.Core.Services
       private readonly IIndividualFactory _individualFactory;
       private readonly OriginDataMapper _originDataMapper;
       private readonly ICache<SpeciesPopulation, Individual> _individualCacheProSpecies = new Cache<SpeciesPopulation, Individual>();
+      private readonly object _locker = new object();
 
       public DefaultIndividualRetriever(
          ISpeciesRepository speciesRepository,
@@ -49,21 +50,24 @@ namespace PKSim.Core.Services
 
       public Individual DefaultIndividualFor(SpeciesPopulation speciesPopulation)
       {
-         if (!_individualCacheProSpecies.Contains(speciesPopulation))
+         lock (_locker)
          {
-            var originDataSnapshot = new OriginData
+            if (!_individualCacheProSpecies.Contains(speciesPopulation))
             {
-               Species = speciesPopulation.Species,
-               Population = speciesPopulation.Name,
-               Gender = DefaultGenderFor(speciesPopulation).Name
-            };
+               var originDataSnapshot = new OriginData
+               {
+                  Species = speciesPopulation.Species,
+                  Population = speciesPopulation.Name,
+                  Gender = DefaultGenderFor(speciesPopulation).Name
+               };
 
-            //We do not need to pass any valid snapshot context in this case.
-            var originData = _originDataMapper.MapToModel(originDataSnapshot, new SnapshotContext(new PKSimProject(), SnapshotVersions.Current)).Result;
-            _individualCacheProSpecies[speciesPopulation] = _individualFactory.CreateStandardFor(originData);
+               //We do not need to pass any valid snapshot context in this case.
+               var originData = _originDataMapper.MapToModel(originDataSnapshot, new SnapshotContext(new PKSimProject(), SnapshotVersions.Current)).Result;
+               _individualCacheProSpecies[speciesPopulation] = _individualFactory.CreateStandardFor(originData);
+            }
+
+            return _individualCacheProSpecies[speciesPopulation];
          }
-
-         return _individualCacheProSpecies[speciesPopulation];
       }
    }
 }
--- a/src/PKSim.Core/Services/DefaultIndividualRetriever.cs
+++ b/src/PKSim.Core/Services/DefaultIndividualRetriever.cs
@@ -1,6 +1,8 @@
-﻿using OSPSuite.Core.Snapshots;
+﻿using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using OSPSuite.Core.Snapshots;
 using OSPSuite.Core.Snapshots.Mappers;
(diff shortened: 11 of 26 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05:
@rwmcintosh Please check

<details><summary><b>Review by Perplexity</b></summary>

## Correctness / design issues

**1. The gate is a single global static lock, not the per-object gate the issue asked for.** `LazyLoadTask` uses `private static readonly object _loadLock` and holds it for the entire load body, so *every* lazy load in the process is serialized — including loads of unrelated building blocks in unrelated projects, and across all `LazyLoadTask` instances (CLI, BatchTool, PKSim.R, tests running in parallel). Issue #3709's own proposed work says "double-checked `IsLoaded` under a **per-object** gate so one thread loads while others wait", and the PR body says "shared gate" without acknowledging the downgrade. Since the guarded region is a DB read plus XML deserialization, and this PR is explicitly a prerequisite for parallel snapshot/model construction (#3710), the fix converts the intended parallel hot path into a globally serial one. A `ConcurrentDictionary<object, object>`/`ConditionalWeakTable` keyed on the target (or a lock on the target object itself) gives the same safety without the global choke point. [[github](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3709)](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3709)

**2. Long-running work under a process-wide lock can hang or deadlock the UI.** The lock is held across `_contentLoader.LoadContentFor`, `loadSimulations` (results, analyses, charts loading) and registration. Two consequences: (a) any UI-thread `Load` (routine navigation/expansion in PK-Sim) blocks for the full duration of a background import's deserialization — an apparent freeze; (b) if anything inside the guarded chain needs the UI thread synchronously — project converters raising warnings/dialogs via `IDialogCreator`, or progress marshalling — a background holder plus a UI thread blocked on the same static lock is a hard deadlock. Worth verifying that no content-load path marshals to the UI thread before merging.

**3. Broken double-checked locking: the fast-path `IsLoaded` read is unsynchronized.** `if (objectToLoad == null || objectToLoad.IsLoaded) return;` reads a plain, non-volatile `bool` outside any lock, while the writer sets it inside the lock. A reader has no acquire barrier, so it may observe `IsLoaded == true` before the deserialized object-graph writes are visible and then hand out a partially initialized building block. This is the classic non-volatile DCLP hole; benign in practice on x64 with the CLR's stronger model, but not guaranteed by the ECMA memory model (relevant if ARM64 is ever targeted). Either make `IsLoaded` volatile in Core, use `Volatile.Read`, or drop the unlocked fast path.

**4. Lock-ordering hazard between the two new locks.** Project converters run during content loading and are listed in #3709 as callers of `DefaultIndividualRetriever`, giving the order `_loadLock → _locker`. If any path inside `DefaultIndividualFor` (`OriginDataMapper.MapToModel`, `IndividualFactory.CreateStandardFor`) reaches `ILazyLoadTask.Load`, the reverse order exists and the two locks can deadlock. Nothing in the PR documents or enforces an ordering.

**5. Sync-over-async now happens while holding a lock.** `_originDataMapper.MapToModel(...).Result` is blocking inside `lock (_locker)`. Under the concurrent `Task.WhenAll` mapping this PR is meant to enable, pool threads queue on the lock while the holder blocks on a continuation that itself needs a pool thread — thread-pool starvation, with deadlock in the worst case. Pre-existing `.Result`, but the lock materially raises the risk.

**6. `DefaultIndividualRetriever` is only half-fixed, yet the PR claims "Fixes #3709".** The issue states the retriever "returns the **same mutable `Individual`** instance to every caller — any downstream mutation is cross-caller corruption no locking can prevent," and the proposed work is to make the cache thread-safe **and** stop handing out the shared instance (clone, or establish immutability). Only the locking half is implemented. Worse, the new spec `should_return_the_same_cached_instance_to_all_callers` locks in the behavior the issue wants removed. Also untouched: the duplicate Singleton registration (`CoreRegister` + `InfrastructureRegister`) called out in the same issue, which means the instance-level lock only guarantees single creation *per instance*. [[github](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3709)](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3709)

**7. Sibling check-then-act paths left unguarded.** `LoadResults<TSimulation>` still does `if (simulation.HasResults) return; _simulationResultsLoader.LoadResultsFor(simulation);` outside any lock, and results/charts loading is only incidentally protected when reached through `Load`. Two threads calling `LoadResults` on the same simulation still race. Similarly, the branch that falls into `else return;` never sets `IsLoaded`, so unsupported types re-enter the lock on every call.

## Test issues

**8. The tests are sleep-based, not deterministic.** The PR states the specs "reproduce the races deterministically," but both rely on `Thread.Sleep(300)` inside a stub to widen the window — that is probabilistic, and it adds ~0.6 s of wall time to the suite. A `ManualResetEventSlim`/`Barrier` handshake would actually be deterministic.

**9. The concurrency specs can't detect the memory-visibility problem in #3.** `_objectToLoad` is `A.Fake<IPKSimBuildingBlock>()`, so `IsLoaded` is a FakeItEasy-managed property whose access is internally serialized — the test never exercises the real non-volatile field it is meant to protect. FakeItEasy fakes are also not designed for concurrent use, so the spec's stability depends on undocumented behaviour.

**10. No test pins the property the design actually needs.** Nothing asserts that loads of *different* objects still proceed concurrently — the exact regression introduced by the static lock in #1 — and nothing asserts that the second (waiting) caller observes fully loaded content rather than just "loaded once."

**11. Weak assertion in the retriever spec.** `_firstResult.ShouldBeEqualTo(_secondResult)` asserts equality where reference identity is the intent; with fakes it happens to coincide, but the assertion doesn't state the invariant.

## Housekeeping

**12.** `private static readonly object _loadLock` is declared between the constructor and `Load`, not with the other fields at the top of the class, and uses instance-field naming for a static — inconsistent with the surrounding file and repo style.

**13.** `DefaultIndividualRetriever.cs` still ends without a trailing newline (pre-existing, but the file is being touched here anyway).

**14.** The 13.0.161 → 13.0.164 bump is broad but the PR body claims Core 13.0.164 is *required* for correctness (it carries the thread-safe `IWithIdRepository`). That makes the safety of this PR dependent on OSPSuite.Core #2959 being merged and released — worth stating as a merge-order gate, since without it the registry races described in #3709 step 4 remain open for every writer that doesn't go through `LazyLoadTask`. [[github](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3709)](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3709)

</details>

rwmcintosh:
1. Global static lock instead of a per-object gate — correct, and the downgrade was mine. Loading now takes a gate keyed on the object being loaded (ConditionalWeakTable), so unrelated loads no longer serialize.

2. UI freeze / deadlock — the freeze half is resolved by 1. The deadlock half does not apply: there is no IDialogCreator, Invoke or BeginInvoke anywhere in the content loader, serialization or project converter paths, so nothing inside the guarded region marshals to the UI thread.

3. Non-volatile fast-path read — correct. The unlocked fast path is gone; IsLoaded is only checked inside the gate now.

4. Lock ordering between the two locks — does not apply: IndividualFactory, OriginDataMapper and DefaultIndividualRetriever contain no reference to ILazyLoadTask, so the reverse order never forms. Both locks are per-key now in any case.

5. .Result while holding the lock — correct. The cache is now ConcurrentDictionary<SpeciesPopulation, Lazy<Individual>> with ExecutionAndPublication, so a creation is gated per population instead of under a shared lock.

6. DefaultIndividualRetriever only half-fixed — correct. The thread-safety half stays here; handing out the shared mutable instance is an ownership decision with a real cost, so it is split into #3718 and the #3709 checklist is narrowed accordingly. The spec now asserts single creation rather than enshrining the sharing. On the duplicate registration: both registers do declare it, but the container resolves the first registration, so only one instance is ever effective.

7. LoadResults unguarded — correct, it now takes the same gate. The else return branch not setting IsLoaded only causes re-entry rather than incorrect state, so I left it.

8. Sleep-based tests — correct, and the PR text overclaimed. Both specs now use ManualResetEventSlim handshakes with no sleeps.

9. Fakes don't exercise the real field — fair; the concurrency specs use real Individual objects now.

10. Nothing pins that different objects load concurrently — added When_loading_different_objects_concurrently_from_multiple_threads, which fails under the previous global lock.

11 / 12 / 13. Reference-identity assertion, gate declared with the other fields, trailing newline — all done.

14. Merge-order gate on Core 13.0.164 — already stated in the PR description; OSPSuite.Core#2959 is merged and released as 13.0.164, which this branch references.

Re-measured after the rework, 60-simulation snapshot in the desktop app with server GC and "run simulations" off: 139 s vs 132 s before — unchanged within noise, since building blocks created during a snapshot load are already loaded and barely touch the gate.

Yuri05:
@rwmcintosh New Review :)

<details><summary><b>Review Nr. 2 by Perplexity</b></summary>

Re-reviewed at commits `4a58473` + `239d187`. The previous findings are addressed; the following are new or still open. [[github](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3716)](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3716)

## New issues introduced by the rework

**1. `Lazy<T>` with `ExecutionAndPublication` caches failures permanently.** In `DefaultIndividualRetriever`, the faulted `Lazy<Individual>` stays in the `ConcurrentDictionary` forever, so a single transient failure inside `createDefaultIndividualFor` — a `.Result` fault from `OriginDataMapper.MapToModel`, a DB hiccup in `CreateStandardFor` — permanently poisons that population for the rest of the process: every later `DefaultIndividualFor` re-throws the same cached exception. The old code retried on the next call. This is a behaviour regression on a service reachable from PK analyses (`VSSCalculator`) and the project converters, i.e. one failure could make "default human" unavailable for the whole session. Fix by removing the entry on failure (catch → `TryRemove(key, out _)` → rethrow) or by not caching faulted entries.

**2. Multiple gates can now be held at once, so lock-order inversion is structurally possible.** The old single static lock was serial but deadlock-free by construction. Now `Load` holds `gateFor(A)` across the whole content-load chain, and that chain performs nested loads (`ISimulationComparison` content loading pulls in its simulations, `loadSimulations` → charts/analyses loaders resolve referenced building blocks, `SimulationMapper.findBuildingBlock` → `IExecutionContext.Load`). A thread can therefore hold gate(A) while requesting gate(B). Correctness now depends on the nested-load graph being acyclic — an invariant the PR neither states nor tests. A cycle deadlocks hard (Monitor reentrancy only saves the same-thread case). Please either document the acyclicity argument next to `_loadGates`, or adopt the same lock-free strategy used for the retriever (a per-object `Lazy`-style gate whose value is published without holding a monitor across dependent loads).

**3. Removing the unlocked fast path puts a `ConditionalWeakTable` lookup, a gate allocation and a Monitor acquire on the already-loaded path.** `Load` is called constantly on already-loaded objects (UI navigation, `AllBuildingBlocks<T>().Each(Load)` loops, per-parameter walks). Every distinct object ever passed to `Load` now permanently gets a gate object plus a CWT dependent handle for the object's lifetime — on a large project that is thousands of dependent handles, which is GC-visible work, and none of it is needed for objects that are already loaded. The author's own measurement (139 s vs 132 s on the 60-simulation snapshot) is ~5% *slower* in the only direction that matters here and is a single sample described as "within noise" without a noise estimate; that number does not support the claim. Cheaper option: keep the fast path but close the visibility hole properly — `Volatile.Read` on `IsLoaded`, or make `IsLoaded` volatile in Core — so the gate is only touched when a load actually happens.

**4. `LoadResults` releases and re-takes the gate.** `Load(simulation)` acquires and releases `gateFor(simulation)`, then `LoadResults` acquires it again for the `HasResults` check. The two checks are no longer one atomic operation: another thread can slot in between, and for an `IndividualSimulation` the results were already loaded inside `Load` via `loadSimulations`, so the second region can still issue a redundant `LoadResultsFor`. Either hold one gate across both steps or drop the second region and rely on `Load`.

**5. Failed loads now fail N times instead of once.** With waiters queueing on the gate, a load that throws leaves `IsLoaded == false` and the object unregistered (`loadObjectBase` unregisters *before* loading), and each queued waiter then retries the same failing DB read serially against a partially mutated object graph. Pre-existing failure semantics, but the gate turns one failure into a fan-out of failures on a half-torn object. Worth a decision on whether a failed load should be marked as such.

## Still open

**6. Assumption about the duplicate Singleton registration is unverified.** The reply states "the container resolves the first registration, so only one instance is ever effective." That is plausible for Windsor but it is exactly the kind of thing that silently changes; because the cache is now per instance, two effective instances would mean two different "default individuals" handed out. A one-line spec resolving `IDefaultIndividualRetriever` twice and asserting reference equality would retire this assumption permanently.

**7. Deadlock/UI dismissal rests on a grep, not on an invariant.** "No `IDialogCreator`, `Invoke`, or `BeginInvoke` anywhere in the content loader, serialization, or project converter paths" is true today and untested tomorrow. Given the gate is held across arbitrary deserialization, consider stating that constraint as a comment on `Load` so the next person adding a progress or warning hook inside a content loader sees it.

## Test issues

**8. Both concurrency handshakes are still racy in the pass direction.** In `When_loading_the_same_object_concurrently_from_multiple_threads`, `_releaseFirstLoad.Set()` is called immediately after `Task.Run` starts the second load — nothing guarantees the second thread has reached `Load` (let alone blocked on the gate) before the first load completes and sets `IsLoaded`. If it arrives late it returns via the in-gate `IsLoaded` check and every assertion still passes — including against the unfixed code, so the spec can quietly stop being a regression test. Same structure in `When_retrieving_the_default_individual_for_the_same_population_from_multiple_threads`. Make the second caller's blocking observable before releasing, e.g. assert `secondLoad.Wait(shortTimeout).ShouldBeFalse()` before `Set()`, or have the stub itself wait for a "second caller arrived" signal.

**9. Assertions run inside stubs on worker threads.** `_releaseFirstLoad.Wait(...).ShouldBeTrue()` and `_creationStarted.Wait(...).ShouldBeTrue()` execute inside FakeItEasy callbacks, so a timeout surfaces as an `AggregateException` out of `Task.WaitAll` rather than as a readable spec failure. For the retriever this is worse: the throw happens inside the `Lazy` factory, so it is cached (issue 1) and the second caller fails with a confusing rethrow of the first thread's timeout.

**10. The negative case costs 5 s of wall clock.** `When_loading_different_objects_concurrently_from_multiple_threads` detects the regression only by a 5-second `Wait` timeout, so a real regression makes CI 5 s slower per run and the assertion is wall-clock dependent (a starved CI agent that needs pool-thread injection is the false-failure mode). A 1-2 s budget is enough to distinguish "proceeded" from "blocked behind a global lock."

**11. Asymmetric coverage for the retriever.** `LazyLoadTask` got the "unrelated objects don't serialize" spec; `DefaultIndividualRetriever` has no equivalent for two different populations creating concurrently, and nothing covers the exception-caching behaviour in issue 1 (a spec for it would currently fail).

**12. `ManualResetEventSlim` instances are never disposed** in either spec.

## Nits

**13.** Brace style in the new specs is inconsistent with the repo and with the same files — `new[] { firstLoad, secondLoad }` and `new Individual { Id = "IND" }` use inner spaces while neighbouring code uses `new[] {_usedBB1, _usedBB2}` and `new SpeciesPopulation {Name = "POP", Species = "HUMAN"}`.

**14.** `_speciesRepository` is faked but unused in `concern_for_DefaultIndividualRetriever`.

**15.** The comment on `_loadGates` reads "gated per object rather than on the object itself" — the intent (a private gate instead of locking the target's own monitor) is right, but the phrasing inverts it; "per object, on a private gate rather than on the object's own monitor" would be clearer.
</details>

rwmcintosh:
1. Lazy<T> caches failures permanently — correct, and the most serious of these: one transient fault would have poisoned that population for the rest of the session. A faulted entry is now removed (the KeyValuePair overload of TryRemove, so only that entry goes and never a newer good one) and the exception rethrown. New spec covers it.

2. Multiple gates held at once — correct, and worth stating plainly: I traded a design that was deadlock-free by construction for one that depends on the nested-load graph being acyclic. The nesting is real — XElementExtensions.AddReferencedObject calls lazyLoadTask?.Load(...) during deserialization, inside the gate. Today those references only run from a container to a simulation (comparison, parameter identification, chart → simulation) and never back, so gates are only ever taken in that one direction, and that argument is now documented next to _loadGates. Rather than leave it as an invariant nothing enforces, #3720 tracks removing it: the nested load is fire-and-add — AddReferencedObject never reads the loaded content, it only attaches the reference — so nested loads can be deferred until the outer gate is released, leaving a thread holding at most one gate and making deadlock impossible whatever shape the graph takes. Kept out of this PR because it deserves its own spec, one that builds a deliberately cyclic graph and proves it completes.

3. Removing the fast path — correct on both counts, including that my 139 s vs 132 s single sample did not support it. The fast path is back, so already-loaded objects cost nothing and get neither a gate nor a dependent handle. Your earlier point about the unsynchronised IsLoaded read is closed rather than traded away: Open-Systems-Pharmacology/OSPSuite.Core#2965 gave the Core implementations a volatile backing field and stated the requirement on ILazyLoadable, and this branch does the same for PKSimBuildingBlock and the two simulation comparisons. Both apps are on the 13.0.165 release that carries it.

4. LoadResults releases and re-takes the gate — correct. A single gate now spans loading the simulation and loading its results.

5. Failed loads fail N times — this is the pre-existing contract rather than a regression: before the gate, concurrent callers each failed too, and retry-on-next-call is what the old code did. No change.

6. Duplicate registration unverified — fair. Added an integration spec resolving IDefaultIndividualRetriever twice and asserting reference equality.

7. Deadlock dismissal rests on a grep — agreed; the constraint is now a comment on Load.

8. Handshakes racy in the pass direction — correct, and the one that mattered most. The second caller now signals that it has started and must fail to complete within a short window before the first is released. I checked this by mutation rather than by assertion: removing the gate from Load makes two of the four observations fail, and reverting the failure-eviction makes the new retriever spec fail.

9. Assertions inside stubs — fixed; stubs only wait, outcomes are asserted in observations.

10. Negative case budget — reduced to 2 s. For the record it is only paid on regression; on success the Wait returns as soon as the task completes.

11. Asymmetric coverage — added specs for two different populations creating concurrently, and for a failed creation being retried.

12 / 13 / 15. Events disposed via Cleanup(), brace style matched to the file, gate comment reworded.

14. _speciesRepository faked but unused — it is a constructor dependency of the retriever, so it has to be supplied. No change.

Yuri05:
@rwmcintosh Next :)

<details><summary><b>Review no. 3 by Perplexity</b></summary>

Re-reviewed at `0063d58` + `dd354eb`, and read the author's replies. Points 1, 3, 4, 6-13 and 15 are properly closed; I verified the two claims made without code changes (no `IDialogCreator`/`Invoke`/`MessageBox` anywhere in `ContentLoader`, `XmlSerializationManager`, the converters, `SimulationResultsLoader`, `SimulationChartsLoader` or `ChartTask`; and `AddReferencedObject` is only wired with a `lazyLoadTask` from `AddReferencedSimulations`, while `AddReferencedObservedData` passes `null` — so the container→simulation direction does hold today). New findings below. [[github](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3716#issuecomment-5441062792)](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3716#issuecomment-5441062792)

## The restored fast path is undermined by cascading `IsLoaded` setters

**1. A load writes `IsLoaded` on objects other than the one whose gate it holds.** `Simulation.IsLoaded` is a setter-only override that does `base.IsLoaded = value; UsedBuildingBlocks.Each(bb => bb.IsLoaded = value);`, and `UsedBuildingBlock.IsLoaded` forwards to `BuildingBlock.IsLoaded`. `Population.IsLoaded` does the same to `FirstIndividual`. So the final `objectToLoad.IsLoaded = true` inside `Load` flips the flag on every building block the simulation uses, while holding only `gateFor(simulation)`. Now that the fast path is back, that flag is exactly what other threads trust without taking a gate: a thread inside `Load(individual)` (holding gate(individual), flag still false, content half-deserialized) can have the flag set true underneath it by a simulation load on another thread, and a third thread then skips the gate and uses the half-loaded object. The simulation's used building blocks are its own copies, which limits how often this is reachable, but `ExpressionProfileUpdater.Load(simulationSubject)`, `OverwriteParameterSetTask` (`buildingBlocksUsingCompound.Each(_lazyLoadTask.Load)`) and `PKCalculationOptionsFactory.Load(s)` all call `Load` on objects reachable from simulations. At minimum the gate comment must state that a load mutates the flag of objects it does not hold the gate for; better, do the cascade through `LazyLoadTask` under each target's gate.

**2. The cascade publishes the container as loaded before its children.** In both overrides `base.IsLoaded = value` runs *first*, then the children are marked. With the new volatile publication, a reader can therefore observe `simulation.IsLoaded == true` while `usedBuildingBlock.IsLoaded` still reads false — and that reader will call `Load` on a building block whose content was already deserialized as part of the simulation, re-entering `Unregister` → `LoadContentFor` → `Register` on a live object graph. Pre-existing ordering, but it directly weakens the guarantee `dd354eb` is claiming ("a reader that sees it set also sees the writes that loaded the object"), because for these two types "loaded" is a composite state. Swapping the two statements in each override is a one-line fix.

## Test robustness

**3. Handshake timeouts are now silently ignored, and a timeout hangs the run instead of failing it.** In closing point 9 the assertions moved out of the stubs, but they were dropped from `Because()` as well, where asserting is perfectly legal: `_creationStarted.Wait(TimeSpan.FromSeconds(5));`, `_secondLoadStarted.Wait(...)`, `Task.WaitAll(new[] {firstRetrieval, secondRetrieval}, TimeSpan.FromSeconds(5));` and `blockedLoad.Wait(...)` all discard their result. If a handshake ever times out, `_firstResult = firstRetrieval.Result` then blocks with no timeout and the suite hangs rather than reporting a failure. Re-add `.ShouldBeTrue()` on the waits performed on the test thread in `Because()`.

**4. The new integration spec is malformed.** In `tests/PKSim.Tests/IntegrationTests/DefaultIndividualRetrieverSpecs.cs`, `When_resolving_the_default_individual_retriever_more_than_once` was inserted *inside* `When_retrieving_the_default_individual_for_a_human`, immediately after `should_return_the_expected_results()`, at the wrong indentation:

```
      }
      public class When_resolving_the_default_individual_retriever_more_than_once : ContextForIntegration<IDefaultIndividualRetriever>
   {
```

It compiles as a nested fixture, so whether it is discovered and run at all depends on the runner's nested-class handling — the spec that is supposed to retire the duplicate-registration assumption may never execute. It should be a sibling of the other fixtures (and preferably derive from the `concern_for_DefaultIndividualRetriever` directly above it, which is what sets `sut`). The file also still ends without a trailing newline — the same nit that was fixed for `DefaultIndividualRetriever.cs`.

**5. "Was blocked" is inferred from a signal set before the call.** `_secondLoadStarted.Set()` runs *before* `sut.Load(...)`, so a second thread that has not yet reached the gate within the 500 ms window produces a false pass. The bias is toward passing, not flaking, so it is not urgent — but signalling from a point that is genuinely contending (or checking `secondLoad.Status`) would make the observation mean what its name says. The author's mutation check is the stronger evidence here; consider noting it in the spec comment so it isn't lost.

## Smaller items

**6. `LoadResults` has no fast path.** `Load` now skips the gate for already-loaded objects, but `LoadResults` takes `gateFor(simulation)` unconditionally, including the common case of a simulation that already has results — and it is called per simulation from `TimeProfileChartsDataCreator`, `PKAnalysesTask` and `PKCalculationOptionsFactory`. Adding the symmetric fast path needs care, because `HasResults` has no publication guarantee of the kind `IsLoaded` just received, so either give it the same treatment or state why the gate is cheap enough here.

**7. The PR description is stale.** It still reads "Requires OSPSuite.Core **13.0.164** (contains OSPSuite.Core#2959)" while the branch is on 13.0.165 throughout (47 references, no leftovers) and now also depends on OSPSuite.Core#2965. The "What changes" section still describes the shared-gate design and the original two-spec test list, none of which survived the rework, and the follow-ups (#3718, #3720) are not mentioned. A reviewer reading the description sees the design that was replaced two rounds ago.

**8. The gate comment does not mention the presentation-layer edge.** `loadObjectBase` → `loadSimulations` → `_simulationChartsLoader.LoadChartsFor` reaches `SimulationChartsLoader`, which lives in `PKSim.Presentation` (namespace `PKSim.Presentation.Infrastructure.Services`), and `ChartTask.UpdateObservedDataInChartsFor` from there. It is clean today, but it is the single most likely place for a future UI dependency to appear inside the gate, so naming it next to the "nothing reached from here may wait on the UI thread" comment is worth the one line.

**9. `Cleanup()` overrides do not call `base.Cleanup()`** in any of the three new fixtures.

**10. Minor spec inconsistencies.** `When_the_creation_of_the_default_individual_fails` asserts two behaviours (throws, then retries) in a single observation and uses `ShouldBeEqualTo` where the sibling spec uses `ReferenceEquals` for the same identity intent; and there are now two `concern_for_DefaultIndividualRetriever` classes and two `DefaultIndividualRetrieverSpecs.cs` files in the same test project (different namespaces, so it builds, but failure reports become ambiguous).
</details>

rwmcintosh:
1. The cascade is real, but the race it enables is not reachable — comment added as the "at minimum" ask. The flags the cascade flips belong to owner-private copies, not shared objects: UsedBuildingBlock.BuildingBlock is "Clone of the project building block", assigned only while the simulation's own content deserializes (UsedBuildingBlockXmlSerializer reads it inline from the simulation's XML — it has no independent content row), and Population.FirstIndividual is Settings.BaseIndividual, the population's own clone, null until the population's content loads. So the cascade targets don't exist or aren't reachable until the owner's gated load completes — no thread can be mid-Load(x) under gate(x) for them. The call sites named load different instances: ExpressionProfileUpdater loads project templates, and OverwriteParameterSetTask's buildingBlocksUsingCompound is, despite its name, a list of simulations (SimulationsUsing). The gate comment now documents that a completing load also flags owned children without their gates, and why that is safe.

2. Correct — fixed. Both Simulation and Population now flag their children before the volatile base write, so a reader that fast-paths on the parent's flag also sees the children flagged.

3. Correct — fixed. Every test-thread handshake wait in Because() (both spec files, including the Task.WaitAlls) now asserts .ShouldBeTrue(), so a timeout fails the run instead of hanging it.

4. Correct — fixed. The spec is a proper sibling deriving from concern_for_DefaultIndividualRetriever, asserts ReferenceEquals(sut, IoC.Resolve<IDefaultIndividualRetriever>()), and the file ends with a newline. (It was being discovered as a nested fixture and did run, but the structure was wrong regardless.)

5. Suggestion taken. Both "was blocked" observations now carry a comment noting the mutation verification — removing the gate (or making the creation eager) makes the observation fail — so that evidence isn't lost.

6. Deliberate, now documented. A LoadResults fast path would need HasResults/Results to get the same publication treatment IsLoaded received (the reference is assigned inside the gate), and its callers run per chart or analysis rather than on hot paths — the uncontended gate is cheap there. A comment on LoadResults states this.

7. Correct — the description is rewritten: 13.0.165 with both Core PRs, the current per-object-gate design, volatile publication, Lazy eviction, single-gate LoadResults, the current spec list with the mutation-verification note, and the #3718/#3720 follow-ups.

8. Added — the UI-thread comment now names the loadSimulations → SimulationChartsLoader/ChartTask edge in PKSim.Presentation.

9. Correct — fixed. All Cleanup() overrides call base.Cleanup() first, matching the existing convention (TemplateDatabaseCreatorSpecs).

10. ReferenceEquals — done. The failure spec keeps its single observation deliberately: the retry assertion is only meaningful after the throwing call in the same sequence, and splitting them into order-dependent sibling observations would be fragile — the two assertions verify one behavior, "the failure is not cached". The duplicate file and concern names follow the established project convention of same-named spec files in Core/ and IntegrationTests/ (DimensionRepositorySpecs, IndividualMapperSpecs, EventBuildingBlockCreatorSpecs, ImportPopulationFactorySpecs, …), so they stay.

Yuri05:
@rwmcintosh Hopefully the last one :) (it's only about the comments)

<details><summary><b>Review No. 4 by Perplexity</b></summary>

Checked `cf67cb1` against the author's reply. Points 2-9 and the `ReferenceEquals` part of 10 are correctly fixed (I verified the reordering in both setters, every `Because()` handshake now asserts, `base.Cleanup()` everywhere, and the integration fixture is now a proper sibling deriving from `concern_for_DefaultIndividualRetriever` with a trailing newline). The declined items (10's single failure observation, the same-named spec files) are fine as argued. Two problems remain, both in the reply to point 1 and in the comment it produced. [[github](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3716#issuecomment-5442244378)](https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3716#issuecomment-5442244378)

## The safety argument for the cascade states a premise that is false

The new comment in `LazyLoadTask` says the flagged children are "private copies deserialized as part of their owner's content: they are not reachable until the owner's load completes and **no other load path targets the same instances**". The last clause does not hold — `Load` is called on simulation-owned building blocks in several places:

- `SimulationConfigurationTask.CreateFor` does `var individual = simulation.Individual;` (`Simulation.cs:790`, `public virtual Individual Individual => BuildingBlock<Individual>()` — the simulation's own copy) and passes it to `_individualBuildingBlockMapper.MapFrom(individual)`, whose base implementation starts with `_lazyLoadTask.Load(pkSimBuildingBlock)` (`PathAndValueBuildingBlockMapper.cs:169`).
- `SimulationCompoundCollectorPresenterBase.cs:73` fans `EditSimulation(simulation, compound)` out over the simulation's compounds (`Simulation.cs:684`, `Compounds => AllBuildingBlocks<Compound>()`), and `SimulationCompoundParameterAlternativesSelectionPresenter.cs:50` / `SimulationCompoundOverwriteParameterSetSelectionPresenter.cs:37` then call `_lazyLoadTask.Load(_compound)`. `SimulationCompoundProtocolPresenter.cs:79` does the same with `SelectedProtocol`.

The conclusion — no reachable race — still stands, but for the *other* reason given in the same sentence: a child is not reachable before its owner's content has been deserialized, and every one of those call sites has already loaded the owning simulation, so the child's flag is set and the child load fast-paths out. That is the argument worth writing down. As it stands, the paragraph makes a claim a maintainer can disprove with one grep, which costs the rest of the paragraph its credibility — and this comment is the only thing enforcing the invariant until #3720.

## The "deserialized as part of their owner's content" premise has a second exception

`UsedBuildingBlock.BuildingBlock` is not only assigned by `UsedBuildingBlockXmlSerializer`. `BuildingBlockInSimulationSynchronizer.UpdateUsedBuildingBlockBasedOnTemplateIn` swaps a fresh clone in at runtime:

```csharp
usedBuildingBlock.BuildingBlock = _cloner.Clone(templateBuildingBlock);
```

That clone takes its flag from the template (`PKSimBuildingBlock.cs:83`, `IsLoaded = buildingBlock.IsLoaded`), and the synchronizer does not load the template first. So a simulation whose own flag is already `true` can end up holding a child whose `IsLoaded` reads `false` — precisely the "a reader that sees the parent loaded also sees its children loaded" property that the reordering in `cf67cb1` was added to establish. It is not a new race (this runs as part of interactive configuration), but it means the property is not an invariant, and the comment should not present it as one. Either load the template before cloning, or state the guarantee as "a *load* flags children before the parent" rather than as a property of the object graph.

## Smaller

- Nothing pins the ordering the reordering commit introduced. A one-line spec (`simulation` with a used building block → set `IsLoaded = true` → assert the building block is flagged; and the same for `Population`/`FirstIndividual`) would stop a future tidy-up from silently restoring `base.IsLoaded = value` to the front, which is invisible in review and untestable after the fact.
- The `LoadResults` comment says its callers "run per chart or analysis rather than on hot paths"; `PKCalculationOptionsFactory.cs:33` sits in `CreateFor`, which runs per simulation and compound during PK-parameter calculation. The uncontended monitor is cheap enough that the decision is still right — just don't justify it with a claim about the call sites that isn't quite true.
</details>

rwmcintosh:
> @rwmcintosh Hopefully the last one :) (it's only about the comments)

All items fixed.

Yuri05 (approved):

PR #3729, opened 2026-09-01
Fixes #3728 Apply the representation info when creating a module from a snapshot

### LINKED ISSUE #3728: No organ icons are shown in simulation tree view on recreated PKSim module
Image
In v12, Organism, organs, and sub-compartments had icons shown.

rwmcintosh:
Fixes #3728

### Cause

Container icons and descriptions are not intrinsic to a PK-Sim spatial structure — no organ container carries them until they are stamped on. That stamping happens in `MoBiExportTask`, which visits the simulation and copies `IconName` and `Description` from the representation info repository:

```csharp
public void Visit(IObjectBase objToVisit)
{
   var repInfo = _representationInfoRepository.InfoFor(objToVisit);
   objToVisit.Icon = repInfo.IconName;
   ...
}
```

That pass only ever ran on the *Export simulation to MoBi* path. Recreating a module in MoBi from the PK-Sim snapshot embedded in it goes through `SnapshotExchange.CreateModule` → `IProjectSnapshotToModuleMapper` → serialize, which never touches it. So a recreated module came back with no icons and no container descriptions, and MoBi's simulation tree — which renders `container.Icon` with no name-based fallback — showed organs blank.

This is why the same model showed icons before: that module had been produced by a MoBi export rather than recreated from its snapshot.

### Change

- Lift the visitor out of `MoBiExportTask` into a new `RepresentationInfoUpdater` in `PKSim.Core.Services`, so export and snapshot-recreate share one implementation.
- Apply it in `PKSim.Starter.SnapshotExchange.CreateModule` and `CreateModuleAndExportInputs`, so a recreated module matches an exported one.
- Apply it in `PKSim.R.Exchange.SnapshotExchange.CreateModule` as well — the headless counterpart that MoBi.R loads had the identical gap.
- `MoBiExportTask` now consumes `IRepresentationInfoUpdater` instead of `IRepresentationInfoRepository`.

`ContainsInfoFor` was the only accessor on `RepresentationInfoRepository` that did not start it: it reads the lookup cache directly, and that cache is only filled in `PerformPostStartProcessing`. Warm in the PK-Sim UI, empty in the CLI/R container the snapshot exchanges run in, where the guard would return false for every object and skip the whole pass silently. It now starts like every sibling accessor.

### Verification

Round-tripped a module to PK-Sim from MoBi and it came back with organ icons.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency of object icons and descriptions during module creation and export.
  * Representation details are now refreshed before modules are serialized or exported.
  * Existing descriptions are preserved when already available.

### DIFF
--- a/src/PKSim.Core/Services/RepresentationInfoUpdater.cs
+++ b/src/PKSim.Core/Services/RepresentationInfoUpdater.cs
@@ -1,0 +1,39 @@
+﻿using OSPSuite.Core.Domain;
+using OSPSuite.Utility.Visitor;
+using PKSim.Core.Repositories;
+
+namespace PKSim.Core.Services
+{
+   public interface IRepresentationInfoUpdater
+   {
+      void UpdateRepresentationInfoIn(IVisitable<IVisitor> visitable);
+   }
+
+   public class RepresentationInfoUpdater : IRepresentationInfoUpdater, IVisitor<IObjectBase>
+   {
+      private readonly IRepresentationInfoRepository _representationInfoRepository;
+
+      public RepresentationInfoUpdater(IRepresentationInfoRepository representationInfoRepository)
+      {
+         _representationInfoRepository = representationInfoRepository;
+      }
+
+      public void UpdateRepresentationInfoIn(IVisitable<IVisitor> visitable)
+      {
+         _representationInfoRepository.Start();
+         visitable.AcceptVisitor(this);
+      }
+
+      public void Visit(IObjectBase objToVisit)
+      {
+         if (!_representationInfoRepository.ContainsInfoFor(objToVisit))
+            return;
+
+         var repInfo = _representationInfoRepository.InfoFor(objToVisit);
+         objToVisit.Icon = repInfo.IconName;
+
+         if (string.IsNullOrEmpty(objToVisit.Description))
+            objToVisit.Description = repInfo.Description;
+      }
+   }
+}
--- a/src/PKSim.Core/Services/RepresentationInfoUpdater.cs
+++ b/src/PKSim.Core/Services/RepresentationInfoUpdater.cs
@@ -18,11 +18,7 @@ namespace PKSim.Core.Services
          _representationInfoRepository = representationInfoRepository;
       }
 
-      public void UpdateRepresentationInfoIn(IVisitable<IVisitor> visitable)
-      {
-         _representationInfoRepository.Start();
-         visitable.AcceptVisitor(this);
-      }
+      public void UpdateRepresentationInfoIn(IVisitable<IVisitor> visitable) => visitable.AcceptVisitor(this);
 
       public void Visit(IObjectBase objToVisit)
       {
--- a/src/PKSim.Starter/SnapshotExchange.cs
+++ b/src/PKSim.Starter/SnapshotExchange.cs
@@ -34,6 +34,7 @@ public static class SnapshotExchange
       var projectSnapshotToModuleMapper = container.Resolve<IProjectSnapshotToModuleMapper>();
 
       var module = projectSnapshotToModuleMapper.MapFrom(projectSnapshot).module;
+      container.Resolve<IRepresentationInfoUpdater>().UpdateRepresentationInfoIn(module);
       var objectIdResetter = container.Resolve<IObjectIdResetter>();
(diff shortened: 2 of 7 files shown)


rwmcintosh:
@coderabbitai review please

src/PKSim.Starter/SnapshotExchange.cs
```diff
@@ -34,6 +34,7 @@ public static object CreateModule(string projectSnapshot)
       var projectSnapshotToModuleMapper = container.Resolve<IProjectSnapshotToModuleMapper>();
 
       var module = projectSnapshotToModuleMapper.MapFrom(projectSnapshot).module;
+      container.Resolve<IRepresentationInfoUpdater>().UpdateRepresentationInfoIn(module);
```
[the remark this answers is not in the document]
rwmcintosh: Good catch — fixed in 8ff72661.

`PKSim.R.Exchange.SnapshotExchange` is the headless counterpart that MoBi.R dynamically loads (`PKSIM_R_SNAPSHOT_EXCHANGE`), so its `CreateModule` had exactly the same gap: the module was mapped and serialized without ever being visited. Anyone recreating a module outside the desktop app would still have got no icons and no container descriptions.

`CreateModule` now resolves `IRepresentationInfoUpdater` through the 4-arg `Api.ResolveTasks` overload and applies it after the mapping, before `ResetIdFor`. `PKSim.R.Bootstrap.ApplicationStartup` registers `CoreRegister`, so the updater resolves in that container too.

On the test: I did not add one against the static `CreateModule` entry points — they need a fully initialized container plus PKSimDB, and there is no existing harness for them. Instead I added `RepresentationInfoUpdaterSpecs` covering the unit that actually carries the behaviour: nested container traversal, icon assignment, that an existing description is preserved, that unknown objects are left alone, and that the repository is started before visiting. That last one matters — `ContainsInfoFor` reads the cache directly, so without `Start()` the guard returns false for every object in the CLI/R container and the whole pass is skipped silently. 8/8 passing.

src/PKSim.Core/Services/RepresentationInfoUpdater.cs
```diff
@@ -0,0 +1,39 @@
...
+      private readonly IRepresentationInfoRepository _representationInfoRepository;
+
+      public RepresentationInfoUpdater(IRepresentationInfoRepository representationInfoRepository)
+      {
+         _representationInfoRepository = representationInfoRepository;
+      }
+
+      public void UpdateRepresentationInfoIn(IVisitable<IVisitor> visitable)
+      {
+         _representationInfoRepository.Start();
```
Yuri05: why do we need an explicit call to `Repository.Start()` here?
Usually it's something the repo must handle internally.

src/PKSim.Core/Services/RepresentationInfoUpdater.cs
```diff
@@ -0,0 +1,39 @@
...
+      public void Visit(IObjectBase objToVisit)
+      {
+         if (!_representationInfoRepository.ContainsInfoFor(objToVisit))
+            return;
+
+         var repInfo = _representationInfoRepository.InfoFor(objToVisit);
+         objToVisit.Icon = repInfo.IconName;
+
+         if (string.IsNullOrEmpty(objToVisit.Description))
+            objToVisit.Description = repInfo.Description;
```
Yuri05, line 32: why are we checking if the description was set before overwriting, but do not check the same for the icon?

rwmcintosh: IDK that. I didn't verify the functionality of this code. It's copy/pasted into this service just for reusability. It's always been like this. I guess that's if the description was set explicitly it shouldn't be overwritten.

Yuri05: OK. Not a big deal

src/PKSim.Core/Services/RepresentationInfoUpdater.cs
rwmcintosh: yes, updated. Start was not called before during `ContainsInfoFor`. Now it does

Yuri05 (approved):

PR #3763, opened 2026-09-16
Fixes #3762 Export to snapshot: "Add an event for every administration" is not exported

### LINKED ISSUE #3762: Export to snapshot: "Add an event for every administration" is not exported
[Itraconazole-Model_events.json](https://github.com/user-attachments/files/32295730/Itraconazole-Model_events.json)
[Itraconazole-Model_events.zip](https://github.com/user-attachments/files/32295731/Itraconazole-Model_events.zip)

In the attached project, I enabled the "Add an event for every administration" for the application protocol and used it the in the simulation

Image

Exported to snapshot, this property is lost. After loading from snapshot, the checkbox is disabled, and the simulation is missing the evetns.

rwmcintosh:
Fixes #3762

## Issue
In the attached project, "Add an event for every administration" was enabled for the application protocol and used in the simulation. Exported to snapshot, this property is lost. After loading from snapshot, the checkbox is disabled and the simulation is missing the events.

## Root cause
The checkbox is bound to `SimpleProtocol.HasEvent`, which is derived from `SimpleProtocol.EventKey`. The event protocol feature (#3617) added `EventKey` to the domain model, `UpdatePropertiesFrom` and the XML serializer, but not to the snapshot `Protocol` class or `ProtocolMapper`. Project save/load therefore worked while snapshot round-trips silently dropped the key. Advanced protocols were unaffected because `SchemaItemMapper` already maps `EventKey` per schema item.

The simulation half of the report follows from the same gap: the event placeholder mappings are exported and restored by `CompoundPropertiesMapper`, but the reloaded protocol declares no used event keys, so there is nothing for them to attach to.

The event offset parameter did not need changes. It is created hidden, but the snapshot export rule ignores visibility and exports any non-default parameter, so a changed offset already round-tripped.

## Fix
- Add `EventKey` to the snapshot `Protocol` class.
- Export it in `ProtocolMapper.createSnapshotFromSimpleProtocol` via `SnapshotValueFor` (null when empty, so the field is omitted for protocols without an event) and restore it in `createSimpleProtocolFrom`.
- Extend `ProtocolMapperSpecs` in both directions. The new observations fail without the mapper change and pass with it.

No snapshot version bump is needed. The field is optional and existing snapshots load exactly as before.

## Summary by CodeRabbit

* **New Features**
  * Protocol event keys are now preserved when protocols are saved to and restored from snapshots.
  * Restored protocols retain their associated event information, improving continuity across snapshot-based workflows.

* **Tests**
  * Added coverage to verify event keys are correctly saved and restored.


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

PR #3767, opened 2026-09-17
Fixes #3765 Add the applied overwrite parameter set to a new set created from a commit

### LINKED ISSUE #3765: When creating a new OPS from a commit, add all the currently applied parameters to the new changes
Basically, take the union of the two sets (the newly generated set, plus the currently applied set from the simulation building block).

Then it should also be possible to accommodate remove/reset when creating a new OPS, and we don't have to prevent it when the user has selected remove items.

rwmcintosh:
Stacked on #3764 — review that one first, this PR targets its branch.

Committing to a **new** set built the set from the tracked changes alone. A simulation running with a set applied therefore produced a new set that dropped every value the simulation was actually using, keeping only what the user had changed since. A new set is now the union of the two, and the dialog shows that union rather than the changes alone.

### What the dialog lists

The grid follows the selected commit mode.

| Mode | Rows |
| --- | --- |
| Create new | the changed parameters, plus the entries of the set applied to the simulation that the user did not touch, marked `Unchanged`. A path the user reset is not listed. |
| Update existing | the changes and the removals, exactly as before. |

Unticking a row when creating a new set leaves its path out of the set. A reset path gets no row because it is not going into the new set and there is nothing to decide about it — showing it would give it a checkbox that means the same thing in both positions. Removing a path from a set therefore stays an update-existing concept, where the path is already in the target and has to be deleted out of it.

Update existing is deliberately left alone. Unticking a change there means "do not write this entry", and making it mean "delete this entry" would turn a deselect into a destructive act on a set the user did not set out to edit.

### Where the union is built

In the mapper that fills the dialog, not in the commit task. Once the dialog lists every path, the presenter passes them all and `createParameterValuesFor` resolves each one from the simulation: an `Unchanged` row is by definition an untracked path, so the simulation still holds the value and the value origin the set applied to it. `CommitSimulationParametersTask` is unchanged by this PR and writes what it is given.

`CompoundCommitDTO.VisibleParameters` does the filtering, so the validation rules, the OK button and the presenter all see the rows of the current mode. That also makes `HasSelectedRemovals` false while creating a new set, which is why the rule about removals targeting the applied set needs no change.

### Tests

- `SimulationToCompoundCommitDTOMapperSpecs` — the untouched entries of the applied set are listed as `Unchanged`, they are shown when creating a new set and hidden when updating an existing one, and a changed path that is also in the applied set is listed once as a change.
- `CompoundCommitDTOSpecs` — `VisibleParameters` per mode, `HasSelectedRemovals` only while updating an existing set, and creating a new set with a selected removal is valid.
- `ParameterToParameterCommitDTOMapperSpecs` — the new row kind.

Each new test was confirmed to fail with its source change reverted.


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

Issue #3461, opened 2026-03-25
Event placeholder mapping in simulation configuration


msevestre:
## Parent PRD

#3424

## What to build

Add event placeholder mapping to the simulation configuration workflow. When a user creates or configures a simulation with a protocol containing event placeholders, the Protocol configuration step must allow mapping each event placeholder to a PKSimEvent building block — alongside the existing formulation mapping.

Specifically:
- Protocol configuration step scans the selected protocol for event placeholders
- UI presents event placeholder mapping alongside formulation placeholder mapping
- User selects a PKSimEvent building block for each event placeholder
- Validation: error if any event placeholder is left unmapped
- The same event can be mapped to multiple placeholders, or different events to different placeholders

See PRD sections: "Event Placeholder Mapping" and "Events Tab Behavior" for full specification.

## Acceptance criteria

- [ ] Protocol configuration step identifies all event placeholders from protocol schemas
- [ ] Mapping UI shows event placeholders alongside formulation placeholders
- [ ] User can map each event placeholder to a PKSimEvent building block
- [ ] Same PKSimEvent can be mapped to multiple placeholders
- [ ] Validation error displayed when an event placeholder is not mapped
- [ ] Simulation cannot be created with unmapped event placeholders
- [ ] Presenter tests cover mapping logic and validation rules

## Blocked by

- Blocked by #3460 (Advanced Protocol UI for event entries)

## User stories addressed

- User story 2: Define event placeholder and map to different events in different simulations
- User story 5: Reuse same protocol with different event mappings
- User story 8: Map event placeholders in the Protocol step of simulation configuration
- User story 9: Clear validation errors when event placeholder is not mapped
- User story 18: Formulation and event placeholder mapping combined in Protocol step

msevestre:
already implemented

PR #3768, opened 2026-09-18
Fixes #3766 Allow an event placeholder to stay unmapped in a simulation

### LINKED ISSUE #3766: Allow an event placeholder to stay unmapped in a simulation
## Summary

Allow an event placeholder of an administration protocol to be left unmapped in a simulation, for example by offering an "Empty" or "None" entry in the mapping dropdown. Today the Protocol configuration step refuses to create the simulation until every event placeholder points at an event building block, as specified in #3461 ("Validation error displayed when an event placeholder is not mapped", "Simulation cannot be created with unmapped event placeholders").

## Why

The feature introduced with #3424 removes a lot of repetition: a meal that accompanies every administration is defined once on the protocol instead of once per occurrence in the simulation. In the OSP model library, however, the same protocol is very often used by both a fed and a fasted simulation, and only the fed one wants the meal.

Going through the published models, 12 protocols are shared this way. Examples:

| Model | Protocol | Simulations with a meal | Simulations without |
| --- | --- | ---: | ---: |
| Itraconazole | `PO SD 200 mg` | 3 | 1 |
| Itraconazole | `PO SD 100 mg` | 2 | 2 |
| Tizanidine | `PO_8mg_SD` | 1 | 3 |
| Raltegravir | `Iwamoto 2008 400mg PO (Figure 1) omeprazole study` | 1 | 4 |
| Atazanavir | `400mg_SD` | 3 | 1 |
| Dapagliflozin | `PO SD 250 mg` | 1 | 1 |

With the current validation, each of these protocols has to be duplicated into a version carrying the placeholder and a version without it. That defeats the purpose of the feature: the modeler trades a repeated event list for a duplicated protocol, and the two copies then have to be kept in step by hand. The fed and fasted arms of a food effect study are exactly the case where one protocol with one dosing regimen is the honest representation.

## Suggested behavior

Treat an unmapped event placeholder as "no event is applied at this placeholder" and create the simulation without it. If a silent omission is considered too easy to overlook, an explicit "Empty" entry in the dropdown, or a warning rather than an error, would both be enough.

The formulation mapping is not a precedent here, because a dosing that has no formulation cannot be simulated, while a protocol without a meal is a perfectly valid simulation.

## Context

Found while converting the OSP model library snapshots to the Version 13 features. 26 simulations across 8 models currently repeat a meal event once per administration and would benefit from the new protocol events.

Related: #3424, #3461.

rwmcintosh:
Closes #3766

## Why

An administration protocol that declares an event placeholder (#3424) is very often shared between a fed and a fasted simulation, where only the fed one wants the meal. Because #3461 requires every placeholder to be mapped before the simulation can be created, such a protocol has to be duplicated into a version carrying the placeholder and a version without it — the two then have to be kept in step by hand. That is the repetition protocol events were meant to remove. Pavel found 12 protocols in the OSP model library shared this way.

## What

An unmapped event placeholder now means *no event is applied at that placeholder*, and the simulation is created without it.

- `EventBuildingBlockCreator` skips a placeholder it cannot resolve instead of throwing. Protocol events only contribute start times, so there is nothing else to leave dangling.
- The `eventNotNull` business rule on `EventPlaceholderMappingDTO` is gone — that rule was what blocked the wizard.
- The mapping dropdown offers an explicit `<None>` entry, so an empty placeholder reads as a deliberate state rather than an unfinished one. `PKSimConstants.UI.NoEvent` already existed and was unused.
- `NoEventFoundException(Protocol, eventKey)` and `Error.NoEventFoundForPlaceholder` became unreachable and are removed.

The chart in the wizard is built from the protocol alone, so it kept plotting the event even when the placeholder was left on `<None>`. The mapping step now passes the unmapped placeholder keys to `IProtocolChartPresenter`, which skips those schema items. The new argument is optional, so the two protocol editors are unaffected — there is no simulation there and the placeholder really is part of the protocol, so it should stay visible.

Nothing about the stored format changes: `EventPlaceholderMappings` is already a list, so an unmapped placeholder is simply an absent entry, in both the project XML and the snapshot.


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

Issue #3232, opened 2025-08-29
Snapshot based runners seem to only use a single core


rwmcintosh:
When fixing [this issue](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3229) for population simulations. It seemed like running individual simulations was always using a single core and never truly parallelizing async/await.

For that issue, the population runner does parallelize all the individual simulations, but that doesn't mean that all separate individual simulations are also parallelized. We have to make another adjustment somewhere.

PR #3770, opened 2026-09-18
Fixes #3769 Update the observed data in simulation charts once the analyses are mapped

### LINKED ISSUE #3769: Observed data not shown in chart editor after snapshot import
E.g. importing the Atazanavir snapshot.

In v12:

Image

In v13 the group is missing:

Image

rwmcintosh:
Fixes #3769

## Symptom

After importing a snapshot, the chart editor's data browser no longer shows the **Observation** category — only **Simulation** — although the observed data curves are still drawn in the chart.

## Root cause

The **Category** group in the data browser comes from `DataColumn.DataInfo.Origin` (`DataColumnDTO.Category` in Core). Nothing is wrong with the observed-data column mapping: the observed data repositories never reach the data browser at all.

`ChartPresenter.InitializeAnalysis` feeds the editor from `chart.AllObservedData()`. On a snapshot load, the only thing that fills a chart's observed-data collection is `ChartTask.UpdateObservedDataInChartsFor`, which walks `simulation.ChartWithObservedData`.

That call used to sit in `SimulationMapper.MapToModel` immediately after the analyses were added to the simulation. #3232 moved analysis mapping out of `SimulationMapper` into `ProjectMapper.addAnalysesToSimulations`, which runs after the entire simulation-mapping phase, but left the `UpdateObservedDataInChartsFor` call behind. It now runs against a simulation with no analyses, so it is a no-op and every chart is left without its observed data.

The curves still render because `CurveMapper` resolves them from `SimulationAnalysisContext(project.AllObservedData, …)`, independently of the chart's own observed-data collection — hence the reported symptom of a visible curve with no Observation group.

## Change

- `ProjectMapper.addAnalysesToSimulations` calls `UpdateObservedDataInChartsFor` right after both `AddAnalyses` calls, so the charts exist when the observed data is attached.
- `SimulationMapper` drops the `IChartTask` dependency it no longer uses.

## Tests

- New `When_converting_a_project_snapshot_with_a_simulation_using_observed_data` in `ProjectMapperSpecs`. It uses a real `ChartTask` so it asserts the outcome (`chart.AllObservedData()` contains the repository) rather than the interaction; verified to fail without the fix.
- Removed the `SimulationMapperSpecs` observation that pinned the call to the location it has now left.
- Full `PKSim.Tests` suite green: 5909 passed, 0 failed, 6 skipped.

## Not addressed

`SimulationComparisonMapper` never populated its comparison charts' observed data either, but that predates this regression and is outside the scope of #3769.


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):
100% sure we had this bug not so long time ago. 
They come again and again 😆

Issue #3776, opened 2026-09-22
"Overwrite parameter sets" commit behavior


PavelBal:
Point 9 is definitely a BUG, and could be solved by implementing 7, I think.

[Midazolam-Model-overwrite-parameter-sets.json](https://github.com/user-attachments/files/32506245/Midazolam-Model-overwrite-parameter-sets.json)

PK-Sim 13 Build 169

1. Load the attached snapshot
2. The Compound "Midazolam" has one OPS "Midazolam - Human", which is selected in the configuration of the simulation "po 0.01 mg (solution)"
3. Open the simulation -> Parameters -> Permeability -> P (interstitial -> intracellular) and change the value of the first entry (Bone):

<img width="895" height="399" alt="Image" src="https://github.com/user-attachments/assets/c2f5ac90-69e0-472d-8f17-8d3899999645" />

4. In the parameters view, select "Mucosa" and reset the value of the first parameter, and change the value of the second:

<img width="1076" height="404" alt="Image" src="https://github.com/user-attachments/assets/db03406e-3421-4743-8d71-a10cd00fb54f" />

5. In the simulation, select "Commit Simulation Parameters to Compound...".
5.a. For "Update Existing Parameter Set", the list of parameters to be updated is correct:

<img width="1312" height="787" alt="Image" src="https://github.com/user-attachments/assets/f9b67b1f-0ce6-4f52-892f-feca7b2e2872" />

5.b. For "Create New Parameter Set", the list of parameters that will be written into the new set is correct:

<img width="1345" height="1007" alt="Image" src="https://github.com/user-attachments/assets/64a85a96-396e-44cc-a2fb-e7d454e41018" />

But the "Unchanged" entry is confusing. However, this is not so critical.

6. Click "OK". Select "Commit to building block"

<img width="716" height="378" alt="Image" src="https://github.com/user-attachments/assets/c0930f70-a7a5-4b27-8520-de1a6872aef8" />

7. Now the **old** OPS is still selected ("Midazolam - Human", the original one), the parameters are still changed, and the orange indicator is active.

<img width="949" height="593" alt="Image" src="https://github.com/user-attachments/assets/3384b0db-1543-47bc-ae89-d30c4e03300c" />

[...]

PR #3779, opened 2026-09-23
#3776 Fix tracking and reset detection of overwritten parameters

rwmcintosh:
Part of #3776. First of four PRs for the overwrite parameter set commit workflow. The first three are independent of each other; the fourth follows once the third is merged.

Two small fixes to how changes of overwritten parameters are tracked and classified. Both are reachable on develop today, and the later PRs of #3776 rely on them.

### Edits of a parameter whose set is no longer selected are tracked

A parameter applied from an overwrite parameter set is flagged as a compound parameter and keeps an empty `Origin.BuilingBlockId`. Once its path is no longer in the selected set, for example after its reset was committed as a removal from the set, `EditParameterCommand` skipped it: it is neither a simulation parameter nor an overwritten path. Editing or resetting it again left no uncommitted change behind, although the value no longer matched the set.

The tracking predicate now also accepts a compound parameter without a building block origin, which is exactly the parameter an applied set leaves behind. Compound parameters that come from the compound building block carry that origin and are unaffected.

### A value origin edit is no longer mistaken for a reset

The commit dialog classified a tracked path of the selected set as a removal when the parameter's default flag was set. Applying a set does not clear that flag, and a value origin edit tracks the path without touching it, so the dialog offered to remove a value the simulation still uses. The reset state is now detected with `ValueDiffersFromDefault`, the check PK-Sim already uses for it: the fixed-value flag for explicit formulas, value against default value for constants.

### Tests

- `SetParameterValueCommandSpecs` and `ResetParameterCommandSpecs`: editing and resetting a parameter applied from a set that is no longer selected.
- `SimulationToCompoundCommitDTOMapperSpecs`: the reset fixtures now reset the value instead of flipping the default flag, and a value origin edit of an applied parameter is listed as an update.

The new observations fail with the source changes reverted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected how parameter changes are tracked when they come from an overwrite set that is no longer selected. Resetting such a parameter now removes its tracked change, while changing its value continues to track it.
  * Improved how parameter updates are reflected when committing changes to a compound: a change in value origin is treated as an update rather than a removal, and removal detection now accounts for whether the current value differs from its default.


rwmcintosh:
@coderabbitai review please

PR #3780, opened 2026-09-23
#3776 Commit simulation parameters only to the set selected in the simulation

rwmcintosh:
Part of #3776. Second of four PRs for the overwrite parameter set commit workflow. Independent of the first and the third.

### What changes for the user

The commit dialog offered **Update Existing Parameter Set** with a drop-down of every set of the compound. It now offers **Update Parameter Set '&lt;selected set&gt;'**, the set selected for the compound in the simulation, next to **Create New Parameter Set**. When no set is selected, the update option is disabled with a tooltip and only a new set can be created. Selecting which rows to commit works as before.

### Why

Committing to a set that the simulation does not use cleared the uncommitted changes, although the simulation still differed from its own set. That is the same mismatch as steps 7 to 9 of #3776, reached through the update option instead of a new set. Restricting the update to the selected set means a commit always writes to the set the simulation applies. To store values in another set, the user selects it when configuring the simulation and commits there.

With a single possible target, the rule that reset parameters can only be removed from the selected set, and the combo lock that enforced it, are no longer needed.

### Code

- `CompoundCommitDTO`: `SelectedExistingSet`, `HasSelectedRemovals` and the two rules built on them are removed. One rule requires a selected set for an update. `CanCreateNewSet` disables the create option when the only rows are resets, since resets alone cannot make up a new set.
- `CommitSimulationParametersView`: the set combo and its layout item are removed, and the update option shows the name of the selected set.
- `SimulationToCompoundCommitDTOMapper` and `CommitSimulationParametersPresenter` use `SetSelectedInSimulation` as the update target.
- `PKSimConstants`: the strings of the removed combo and rule are replaced by the new option and tooltip texts.

### Tests

`CompoundCommitDTOSpecs`, `CommitSimulationParametersPresenterSpecs` and `SimulationToCompoundCommitDTOMapperSpecs` are adapted to the selected set as the only update target.

## Summary by CodeRabbit

* **Updates**
  * The simulation parameter commit dialog now lets you choose between creating a new parameter set and updating the set selected in the simulation.
  * The option to choose a different existing parameter set has been removed. Available commit choices reflect whether creating a new set or updating the selected set is possible.


rwmcintosh:
@coderabbitai review please

